PR #4, opened 2020-09-14
Fixed player movement axis bug

DapperDino:
Bug is fixed.

Issue: https://github.com/UnityTechnologies/UOP1/issues/2
Card: https://unity-open-projects.codecks.io/decks/13/card/12w-movement-axis-is-not-recalculated-if-camera-rotates-while-holding-keyboard-keys


dimmduh:
many changes because of tab changed to 4 spaces, someone should create and check code style before merge

PR #5, opened 2020-09-27
Did some code refactoring inside the Character Controller

Quickz:
Added some spaces. Separated certain lines of code - made sure comments and attributes are in separate lines.

### DIFF
--- a/UOP1_Project/Assets/Scripts/Characters/Character.cs
+++ b/UOP1_Project/Assets/Scripts/Characters/Character.cs
@@ -7,20 +7,45 @@ public class Character : MonoBehaviour
 {
 	private CharacterController characterController;
 
-	[Tooltip("Horizontal XZ plane speed multiplier")] public float speed = 8f;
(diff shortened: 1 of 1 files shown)


Westerveld:
Hey, I don't think this is really a necessary pull request. There isn't any additions to the project other than a couple of spaces here and there. Its not really a refactor but more of a reformat, so may not be accepted. Please bare in mind that every Pull Request brings additional work of needing to be reviewed and tested.

MileyHollenberg:
I think it's best to first have the discussion inside this thread, https://forum.unity.com/threads/deciding-coding-standards-and-guidelines.980106/, regarding the code-style before making any changes to the formatting for now. Once we're aligned we can start working on getting a code-style in place but until then it's best not to make any changes yet

ciro-unity:
I agree with @MileyHollenberg and @Westerveld on this. Small reformats are more work to parse review for us than the value they provide. I'm working on a style guide which I will share soon.

Thanks anyway, @Quickz.

PR #6, opened 2020-09-29
Initial addition of adding a Finite State Manager.

Dustin00:
AppManager takes over program flow control, all MonoBehaviors with an Update will need to "register" with it.

Sets up for Title, Game Menu, Settings, Exit, etc.

MonoBehaviors can supply an Update for when in game play and when the game is paused (looking at Settings, etc).

DefineStateEngine() shows the States and transitions you can do.

Lots of logging for state changes so you can see the flow a bit in the console log

### DIFF
--- a/UOP1_Project/Assets/Scripts/FSMAction.cs
+++ b/UOP1_Project/Assets/Scripts/FSMAction.cs
@@ -1,0 +1,45 @@
+﻿namespace UOP1.FSM35
+{
+	public delegate void StateAction();
+
+	class ActionKey
+	{
+		public int SourceStateID;
+		public int SourceTransitionID;
+	}
+
+	public class FSMAction<TState, TTransition>
+	{
+		private FSMState<TState> initialState;
+		private FSMTransition<TTransition> transition;
+		private FSMState<TState> finalState;
+		private StateTransitionType transitionType;
+
+		public FSMAction(FSMState<TState> initialState, 
+			FSMTransition<TTransition> fsmTransition, 
+			FSMState<TState> finalState, 
+			StateTransitionType transitionType)
+		{
+			this.initialState = initialState;
+			transition = fsmTransition;
+			this.finalState = finalState;
+			this.transitionType = transitionType;
+		}
+
+		public override string ToString()
+		{
+			return $"{initialState} to {finalState}";
+		}
+
+		public StateTransitionType StateTransitionType()
+		{
+			return transitionType;
+		}
+
+		public FSMState<TState> InitialState => initialState;
+
+		public FSMTransition<TTransition> Transition => transition;
+
+		public FSMState<TState> FinalState => finalState;
+	}
+}
--- a/UOP1_Project/Assets/AppManager.cs
+++ b/UOP1_Project/Assets/AppManager.cs
@@ -1,0 +1,249 @@
+﻿using System;
+using System.Collections;
+using System.Collections.Generic;
+using UnityEditor;
+using UnityEditor.VersionControl;
+using UnityEngine;
+
+public class AppManager
+{
+  #region Singleton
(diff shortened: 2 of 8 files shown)


tomate-salat:
I'd prefer to see that bolt becomes added to the project.

dimmduh:
too much over complicated
I would recommend to use classic MonsterLove FSM
https://github.com/thefuntastic/Unity3d-Finite-State-Machine/tree/master/example_project/Assets/MonsterLove/StateMachine

TJHeuvel:
I think [the card](https://open.codecks.io/unity-open-project-1/decks/15/card/157-state-machine) is a bit misleading, it asks for a solution but doesnt state the problem.

Judging by the states given there, what they're after is a state machine to keep track of what the character is doing.  This pull request seems to be more about the state of the entire game, e.g. are we paused or not.

Dustin00:
I isolated the FSM logic. You define the States and Transitions between them, and those are the only "legal" changes you can make during execution. My design has a complete, easy-for-humans-to-read map in a single location of how the program flows and what happens any time you enter or leave a state.

`       .In(AppStates.Title)
          .EntryAction(ShowTitle)
          .Go(AppStates.GameMenu)
          .ExitAction(HideTitle)

        .In(AppStates.Play)
          .EntryAction(ResumePlay)
          .On(AppTransitions.UpdateSettings).GoSub(AppStates.Settings)
          .On(AppTransitions.MainMenu).Goto(AppStates.GameMenu)`

MonsterLove allows anybody, anywhere to fire off a state change to any new state. It also allows/requires, everybody to monitor the FSM in their Update statements:

`	void Update()
	{
		var state = fsm.State;
		if (state == States.One)
		{
			fsm.ChangeState(States.Two);
		}
	}`

With mine, if you try to fire the event ResumePlay and you're in the Title state, you get a log entry of "Transition "ResumePlay" not found in "Title" state".

You can use code that guards against bugs or not.

UOP1_Project/Assets/Scripts/FSMAction.cs
```diff
@@ -0,0 +1,45 @@
...
+	class ActionKey
+	{
+		public int SourceStateID;
+		public int SourceTransitionID;
+	}
+
+	public class FSMAction<TState, TTransition>
+	{
+		private FSMState<TState> initialState;
+		private FSMTransition<TTransition> transition;
```
atb-brown, line 14: I'm seeing tabs in some files and spaces in others. Is there a codestyle that's been decided on?

tacman:
I quite like the idea of a FSM here, and pretty much everywhere an object can exist in a state where transitions aren't trivial.  I think, though, that this would be better to discuss in an issue first, then once it's agreed upon that it's a good idea, code it up and do a PR.

UOP1_Project/Assets/Scripts/FSMAction.cs
MileyHollenberg: Not yet, the code-style is still in the works. It will be posted on the forum and somewhere in Github as well most likely once we have one :)

L4ZZA: @MileyHollenberg even if it was not established yet it should at least be the same throughout the files for now. Wouldn't you agree?

MileyHollenberg: Once we have the code-style in place we'll also make sure to add an automatic linter so that we don't have to worry too much about these specifics :) (will be tabs instead of spaces btw). For now I don't see this as a priority for this PR

ciro-unity:
I'll close this one since we all agreed it's a bit over scoped.

Plus @Dustin00 you'll have to sign the CLA if you want us to approve any PR (you only need to do it once).

Issue #8, opened 2020-09-30
Player can climb steep surfaces


ciro-unity:
**Short description**
The player can climb very steep surfaces if it spams the jump button.

**Expected behaviour**
One would expect the player to slide down, and to be prevented from jumping if not on a stable surface.

**To Reproduce**
1. Open gameplay scene, press Play.
2. Move up the mountain made with Terrain.
3. Keep moving, the player will stop on a steep slope.
4. Press the jump button. Now you can climb up by jumping repeatedly.

scascar:
I'd like to work on this issue.

ciro-unity:
Oh no, an overlap! Oscar, maybe you can jump in the PR #12 and check if the solution works / the code is good?

dhumildholiya-code:
Hello, I'd Like to work on this issue. Let me know if it's not assigned.

richard-fine:
Related to this: If the player is walking up a steep-but-not-too-steep surface, should they move more slowly? I think that if you walk up a slope which is gradually increasing in steepness, it'd be strange to move at full speed and then suddenly be stopped. 

Making the player walk more slowly depending on how much the ground is sloping upwards will better communicate to the player that they're approaching a limit on how steep the ground can be.

MEGATANK58:
@ciro-unity I actually tested the game today again and the bug is still there

PR #10, opened 2020-09-30
Fix the Player bug that allow him to climb steep slopes

lazarovttac:
I made a really quick fix to the bug @ciro-unity commented in this issue: https://github.com/UnityTechnologies/open-project-1/issues/8.

This is done by obtaining the normal of the surface the Player's character controller is hitting in every frame. If this angle is gratter than the specified in the Chareacter Controller Component, then a boolean paramenter called "canJump" is set to false. 
If "canJump" = false, the Player can´t jump.

### DIFF
--- a/UOP1_Project/Assets/Scripts/Characters/Character.cs
+++ b/UOP1_Project/Assets/Scripts/Characters/Character.cs
@@ -16,6 +16,11 @@ public class Character : MonoBehaviour
     [Tooltip("The maximum speed reached when falling (in units/frame)")] public float maxFallSpeed = 50f;
     [Tooltip("Each frame while jumping, gravity will be multiplied by this amount in an attempt to 'cancel it' (= jump higher)")] public float gravityDivider = .6f;
 
+    //Peventing Player from climbing up slopes
+    private float slopeLimit = 30f;
+    private Vector3 hitNormal;
+    private bool canJump;
+
     private float gravityContributionMultiplier = 0f; //The factor which determines how much gravity is affecting verticalMovement
     private bool isJumping = false; //If true, a jump is in effect and the player is holding the jump button
     private float jumpBeginTime = -Mathf.Infinity; //Time of the last jump
@@ -27,6 +32,7 @@ public class Character : MonoBehaviour
     private void Awake()
     {
         characterController = GetComponent<CharacterController>();
+        slopeLimit = characterController.slopeLimit;
     }
 
     private void Update()
@@ -98,6 +104,16 @@ public class Character : MonoBehaviour
                 ref turnSmoothSpeed,
                 turnSmoothTime);
         }
+
+        if (!(Vector3.Angle(Vector3.up, hitNormal) <= slopeLimit))
+        {
+            canJump = false;
+        }
+        else
+        {
+            canJump = true;
+
+        }
     }
 
     //---- COMMANDS ISSUED BY OTHER SCRIPTS ----
@@ -109,7 +125,7 @@ public class Character : MonoBehaviour
 
     public void Jump()
     {
-        if (characterController.isGrounded)
+        if (characterController.isGrounded && canJump)
         {
             isJumping = true;
             jumpBeginTime = Time.time;
@@ -121,5 +137,10 @@ public class Character : MonoBehaviour
     public void CancelJump()
     {
         isJumping = false; //This will stop the reduction to the gravity, which will then quickly pull down the character
+    }
+
+    private void OnControllerColliderHit(ControllerColliderHit hit)
+    {
+        hitNormal = hit.normal;
     }
 }
--- a/UOP1_Project/ProjectSettings/QualitySettings.asset
+++ b/UOP1_Project/ProjectSettings/QualitySettings.asset
@@ -21,7 +21,7 @@ QualitySettings:
     skinWeights: 2
     textureQuality: 0
     anisotropicTextures: 1
-    antiAliasing: 0
+    antiAliasing: 2
     softParticles: 0
     softVegetation: 1
     realtimeReflectionProbes: 1
--- a/UOP1_Project/Packages/packages-lock.json
+++ b/UOP1_Project/Packages/packages-lock.json
@@ -119,7 +119,7 @@
       "url": "https://packages.unity.com"
     },
     "com.unity.test-framework": {
-      "version": "1.1.14",
+      "version": "1.1.16",
       "depth": 1,
       "source": "registry",
       "dependencies": {
--- a/UOP1_Project/ProjectSettings/ProjectVersion.txt
+++ b/UOP1_Project/ProjectSettings/ProjectVersion.txt
@@ -1,2 +1,2 @@
-m_EditorVersion: 2019.4.7f1
-m_EditorVersionWithRevision: 2019.4.7f1 (e992b1a16e65)
+m_EditorVersion: 2019.4.8f1
+m_EditorVersionWithRevision: 2019.4.8f1 (60781d942082)


UOP1_Project/ProjectSettings/QualitySettings.asset
```diff
@@ -21,7 +21,7 @@ QualitySettings:
     skinWeights: 2
     textureQuality: 0
     anisotropicTextures: 1
-    antiAliasing: 0
```
lazarovttac, line 24: This may have changed because of the version change...

pgpais:
Hey, was browsing around and noticed that you have changed project files.

This will probably not be accepted because of this

lazarovttac:
> Hey, was browsing around and noticed that you have changed project files.
> 
> This will probably not be accepted because of this

lazarovttac:
https://github.com/UnityTechnologies/open-project-1/pull/10#issuecomment-701507064

I didn't change that on purpose, I think it's because I have a different Unity version...

DapperDino:
When you choose which files to commit you can ignore changes to those particular files so that you only commit your code.

lazarovttac:
https://github.com/UnityTechnologies/open-project-1/pull/10#issuecomment-701511451

Thank you, I didn't know...

UOP1_Project/Assets/Scripts/Characters/Character.cs
```diff
@@ -27,6 +32,7 @@ public class Character : MonoBehaviour
     private void Awake()
     {
         characterController = GetComponent<CharacterController>();
```
TJHeuvel, line 34: Preferably just serialized the CharacterController, GetComponent is more costly!

UOP1_Project/Assets/Scripts/Characters/Character.cs
```diff
@@ -27,6 +32,7 @@ public class Character : MonoBehaviour
...
+        slopeLimit = characterController.slopeLimit;
```
TJHeuvel, line 35: Why do you copy over fields? This just creates double bookkeeping, and makes it more error prone. Instead you can just make the slopeLimit variable return the limit from the controller. 

Currently if the limit on the controller is changed its not reflected here.

UOP1_Project/Assets/Scripts/Characters/Character.cs
```diff
@@ -98,6 +104,16 @@ private void Update()
                 ref turnSmoothSpeed,
                 turnSmoothTime);
         }
+
+        if (!(Vector3.Angle(Vector3.up, hitNormal) <= slopeLimit))
```
TJHeuvel, line 108: A few comments here:

1. If you're doing `if(a) b = false; else b = true;` you might as well write `b=a`. 
2. The negating is quite confusing, why say `not smaller than` instead of `bigger than`?

Instead of these 10 lines i'd write:

`canJump = Vector3.Angle(Vector3.up, hitNormal) <= slopeLimit;`

Functionally equivalent, concise and (in my opinion) more readable.

UOP1_Project/Assets/Scripts/Characters/Character.cs
lazarovttac: You are right, but I did not write that, it was already there...

UOP1_Project/Assets/Scripts/Characters/Character.cs
```diff
@@ -122,4 +138,9 @@ public void CancelJump()
     {
         isJumping = false; //This will stop the reduction to the gravity, which will then quickly pull down the character
     }
+
+    private void OnControllerColliderHit(ControllerColliderHit hit)
+    {
+        hitNormal = hit.normal;
```
TJHeuvel, line 144: Why dont you assign `canJump` here? Instead now the same limit is being calculated every frame, instead of every touch. The latter is more performant!

TJHeuvel:
Hope you dont mind that i added some comments, really good start!

lazarovttac:
https://github.com/UnityTechnologies/open-project-1/pull/10#issuecomment-701570661
Thank you!, It's okay, I appreciate your comments. I have a lot to learn about C# in unity...

TJHeuvel:
Please dont be discouraged, its really good. Its functional, thats what counts. Collaborating on a project like this is a great way to learn!

UOP1_Project/Assets/Scripts/Characters/Character.cs
SideSwipe9th: I disagree @TJHeuvel . Using GetComponent<T> in Awake or Start is largely accepted best practice when caching references for use later. The performance difference between setting the variable in the inspector, and using GetComponent<T> is negligible. It's only when you're using GetComponent<T> frequently, for example in the Update loop that it becomes a performance issue. 

For the character script, Awake should only be getting called maybe a couple of times over the life cycle of the scene; once on scene load, and any time the character gameobject is deactivated and later reactivated for example in a cutscene.

TJHeuvel:
I really dont think its generally accepted at all, and i cannot see any
benefit over using the serialization system.

Besides it being more code, it is also non-obvious that this particular
script needs a charactercontroller. If it is indeed serialized anyone can
immediately see from the editor it requires one. It can also be beneficial
to be able to drag it in from a parent or child, instead of being forced to
add the component directly. Using OnValidate callbacks you can even find
the link automatically.

Spawning gameobjects is already expensive, i definitely have seen these
getcomponents come up in the profiler and saved milliseconds delegating it
to the serialization system.

Is there any benefit at all to using GetComponent?

TJHeuvel:
> For the character script, Awake should only be getting called maybe a
couple of times over the life cycle of the scene;

You are right here, but what if a particular script is spawned more often?
Why not use the same approach everywhere?

SideSwipe9th:
> I really dont think its generally accepted at all, and i cannot see any benefit over using the serialization system.

Well setting aside that the GetComponent call you were commenting on was written by the developers at Unity, using GetComponent in this way is covered in the course content for the Unity Certified Professional Programmer certification, as well as the numerous authorised "Learn to program with Unity" courses you can find on Zeneva and Udemy. 

In addition many of the most popular and commonly used asset store systems utilise it in this way, for example A* Pathfinding Project, EasyRoads3D, Microsplat, Amplify, DoozyUI, and Pixel Crusher's Dialogue System. I could probably list more, however these are all assets I am familiar with and have source code for. 

The only place where it's absolutely considered bad practice to use GetComponent is if you're calling it frequently, for example in every frame within Update, or multiple times a frame within some other behaviour. In that circumstance it is considered best practice to either use GetComponent in Start/Awake to cache a reference like the Unity devs did with the Character script above, or to use a serialised/public field and set it directly in the inspector. 

> Spawning gameobjects is already expensive, i definitely have seen these getcomponents come up in the profiler and saved milliseconds delegating it to the serialization system.

How expensive a GetComponent call is depends on the structure of the GameObject, and how many components there are on it. The more objects and scripts, the more it has to iterate through to find a component. I have no doubt that you'll have saved time by swapping a GetComponent in a Start or Awake to a serialized variable, in the same way you can save time by caching it whenever you're using it freqently in Update. 

> Is there any benefit at all to using GetComponent?

Yes. Scope. GetComponent checks for the component type from the GameObject your script is attached to and through its children. The variants GetComponentInChildren and InParent likewise either search down or up a GameObject's hierarchy from your script's current location in the hierarchy. Using the script above as an example, if I have a GameObject with a Character script attached and I only ever want to get the CharacterController that is also attached to this GameObject, the safest way is to use GetComponent. Furthermore I can ensure that this GameObject will have a CharacterController if I were to use the RequireComponent attribute. 

> it is also non-obvious that this particular script needs a charactercontroller

Easily solved with the RequireComponent attribute. 

> If it is indeed serialized anyone can immediately see from the editor it requires one. It can also be beneficial to be able to drag it in from a parent or child, instead of being forced to add the component directly.

Yes this is true, but this is also potentially dangerous. By using a serialised field you lose the ability to ensure the scope of the component. For example, again using the Character/CharacterController from above. While I, you, or some other developer can as you say drag and drop the CharacterController component from this GameObject onto the field in the inspector, we can just as easily drag and drop the CharacterController from another GameObject that has it as a component. And there is, as far as I'm aware, no easy way to ensure you are correctly scoping the component when applying it in the inspector. You might be able to do it with a custom inspector, checking if the GameObject that the CharacterController you are trying to assign is the same as, a child of, or parent of the GameObject the Character script is assigned to. But you get that for free as part of a GetComponent call in Awake/Start. 

> You are right here, but what if a particular script is spawned more often? Why not use the same approach everywhere?

That is a very different question. Some of the asset store systems I listed before like A* use a GetComponent call in Awake and OnEnable without significant performance overhead. If this script were to be used more often in a scene the performance overhead will most likely be felt most on scene loading, or if you are disabling/enabling a mass of GameObjects with Character components on them. And depending on the impact at that time, it may be prudent to move to a serialised field while also being mindful of the scope protection that you will lose by doing so.

TJHeuvel:
Firstly thanks for your reply. Although we've far passed my initial intent of some quick feedback, i appreciate it. It has shown me that what is considered best practice can vary greatly.

> Well setting aside that the GetComponent call you were commenting on was written by the developers at Unity

Lets set that aside indeed, just because it was written here does not mean its good practice. In the current repo it seems like everything that should be serialized [is made public](https://github.com/UnityTechnologies/open-project-1/blob/master/UOP1_Project/Assets/Scripts/InputReader.cs#L9), instead of private with a SerializeField attribute. I hope you agree that is not best practice. 

> In addition many of the most popular and commonly used asset store systems utilise it in this way

I understand that they do, but that doesnt make it any good. Microsplat also makes a copy of their shader for every scene you use it for, that does not make it good. Just because its widespread, does not mean it is a best practice. Easyroads main script is called `runtimeScript`, half the file is commented out, it creates an entire primitive in the Awake method. Those are not good practices in any way. 

> Yes. Scope. GetComponent checks for the component type from the GameObject your script is attached to and through its children.

Personally i dont see the benefit of this, though if thats a hard requirement i'm sure some attribute could be contrived to make sure it exists on that particular object. If that is such a vital part of your setup, i would highly recommend that you validate this. Especially considering this is an open project, where the programmers will never even meet the designers that use this component, making contracts like these visible in the editor is incredibly important. 

> For example, again using the Character/CharacterController from above. While I, you, or some other developer can as you say drag and drop the CharacterController component from this GameObject onto the field in the inspector, we can just as easily drag and drop the CharacterController from another GameObject that has it as a component. 

Perfect! If that is what the designer intents to do, i dont see why thats an issue at all. Let say in one part of the level we want a different limit, that is completely possible. Again i personally have never seen this as an issue, especially when working in isolation on prefabs where you cannot even assign something outside the hierarchy. I see this as a huge benefit, being able to assign it to some script down the hierarchy or even somewhere completely different.

 > Easily solved with the RequireComponent attribute.
There are several issues with RequireComponent. First and foremost, again, performance. It affects both editor and runtime performance, in large projects quite severely. [This is from a profile i did](https://i.imgur.com/0hlLhnj.png), i have since reported this issue with Unity. This same cost is paid every time you add or remove a script in the Editor, attributes like these scale _very_ poorly with the Unity Editor. Just refreshing the assemblies in search for attributes is an incredibly slow process, and should be avoided. 

Secondly, lets compare our methods. Mine would be:

```
class SomeClass : MonoBehaviour
{
 [SerializeField] private CharacterController characterController;
}
```

Yours would be:

```
[RequireComponent(typeof(CharacterController))]
class SomeClass : MonoBehaviour
{
 private CharacterController characterController;
 void Awake() 
 {
  characterController = GetComponent<CharacterController>();
 }
}
```
I think serializing it is better because it is far less code to write, less error prone (try using characterController in the OnEnable!), more performant, scales better with project complexity, is more flexible and it automatically makes it obvious to anyone using the component what it dependencies it has. That we cannot easily limit the component to that exact same gameobject to me is not worth changing over.

PR #12, opened 2020-09-30
Player Climbing Slopes Bug Fixed

lazarovttac:
I made a really quick fix to the bug @ciro-unity commented in this issue: #8.

### DIFF
--- a/UOP1_Project/Assets/Scripts/Characters/Character.cs
+++ b/UOP1_Project/Assets/Scripts/Characters/Character.cs
@@ -17,7 +17,6 @@ public class Character : MonoBehaviour
     [Tooltip("Each frame while jumping, gravity will be multiplied by this amount in an attempt to 'cancel it' (= jump higher)")] public float gravityDivider = .6f;
 
     //Peventing Player from climbing up slopes
-    private float slopeLimit = 30f;
     private bool canJump;
 
     private float gravityContributionMultiplier = 0f; //The factor which determines how much gravity is affecting verticalMovement
@@ -31,7 +30,6 @@ public class Character : MonoBehaviour
     private void Awake()
     {
         characterController = GetComponent<CharacterController>();
-        slopeLimit = characterController.slopeLimit;
     }
 
     private void Update()
@@ -130,6 +128,6 @@ public class Character : MonoBehaviour
 
     private void OnControllerColliderHit(ControllerColliderHit hit)
     {
-        canJump = Vector3.Angle(Vector3.up, hit.normal) <= slopeLimit;
+        canJump = Vector3.Angle(Vector3.up, hit.normal) <= characterController.slopeLimit;
     }
 }
--- a/UOP1_Project/Assets/Scripts/Characters/Character.cs
+++ b/UOP1_Project/Assets/Scripts/Characters/Character.cs
@@ -18,7 +18,6 @@ public class Character : MonoBehaviour
 
     //Peventing Player from climbing up slopes
     private float slopeLimit = 30f;
-    private Vector3 hitNormal;
     private bool canJump;
 
     private float gravityContributionMultiplier = 0f; //The factor which determines how much gravity is affecting verticalMovement
@@ -131,7 +130,6 @@ public class Character : MonoBehaviour
 
     private void OnControllerColliderHit(ControllerColliderHit hit)
     {
-        hitNormal = hit.normal;
-        canJump = (Vector3.Angle(Vector3.up, hitNormal) <= slopeLimit);
+        canJump = Vector3.Angle(Vector3.up, hit.normal) <= slopeLimit;
     }
 }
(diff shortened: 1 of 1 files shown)


UOP1_Project/Assets/Scripts/Characters/Character.cs
```diff
@@ -122,4 +128,10 @@ public void CancelJump()
     {
         isJumping = false; //This will stop the reduction to the gravity, which will then quickly pull down the character
     }
+
+    private void OnControllerColliderHit(ControllerColliderHit hit)
+    {
+        hitNormal = hit.normal;
+        canJump = (!(Vector3.Angle(Vector3.up, hitNormal) <= slopeLimit));
```
MileyHollenberg, line 135: Nit: there's no need for the outer parentheses here

atb-brown:
This doesn't seem to work for me. :(

I checked out your branch and noticed the following behaviors:
 1. I was still able to climb up steep angles
 2. I was no longer able to jump on flat ground.

https://imgur.com/a/SpeOWRV

lazarovttac:
https://github.com/UnityTechnologies/open-project-1/pull/12#issuecomment-701651530

I think I can see the error without even opening Unity... Try removing the "!" here, in line 135.

` canJump = (!(Vector3.Angle(Vector3.up, hitNormal) <= slopeLimit));`

atb-brown:
@lazarovttac 
`Try removing the "!" here`

Worked like a charm!

lazarovttac:
https://github.com/UnityTechnologies/open-project-1/pull/12#issuecomment-701655206

Great!, sorry, that was my mistake because I changed a few lines, just before making the request, and didn't test it out...

atb-brown:
No worries! I'm just here to learn, haha.

lazarovttac:
> No worries! I'm just here to learn, haha.

Me too!

pgpais:
This doesn't seem to provide the full fix. The player is supposed to slide down at certain slopes (as if falling). This only disables jumping.

The player should slide down until the slope angle is "acceptable" to jump, if I understood correctly.

DaiganIT:
try adding a physics material to let the player slide

pgpais:
> 
> 
> try adding a physics material to let the player slide

@DaiganIT I don't think tou can do that with CharacterController

DaiganIT:
> > try adding a physics material to let the player slide
> 
> @DaiganIT I don't think tou can do that with CharacterController

ah yes, just opened the project now. I assumed it had a RigidBody

pgpais:
A nudge away from the slope could "fake" the sliding down behaviour. It doesn't feel correct, though

MileyHollenberg:
> A nudge away from the slope could "fake" the sliding down behaviour, as suggested in the forums. It doesn't feel correct, though

This could be achieved by performing a raycast downwards and moving the character across the normal of the slope, that would not make it have a stepping effect at least

UOP1_Project/Assets/Scripts/Characters/Character.cs
```diff
@@ -122,4 +128,10 @@ public void CancelJump()
     {
         isJumping = false; //This will stop the reduction to the gravity, which will then quickly pull down the character
     }
-}
\ No newline at end of file
+
+    private void OnControllerColliderHit(ControllerColliderHit hit)
+    {
+        hitNormal = hit.normal;
+        canJump = (Vector3.Angle(Vector3.up, hitNormal) <= slopeLimit);
```
tomate-salat, line 135: You could use characterController.slopeLimit here instead of an variable. So you still could use the editor to try out different values during runtime.

If you don't like that then adding a SerializedField-Attribut to the variable may would be a compromise.

pgpais:
> 
> 
> > A nudge away from the slope could "fake" the sliding down behaviour, as suggested in the forums. It doesn't feel correct, though
> 
> This could be achieved by performing a raycast downwards and moving the character across the normal of the slope, that would not make it have a stepping effect at least

I tried implementing this and sent a PR #22 . I basically move the player directly through the transform so it falls back onto the slope (which keeps happening if slope is too steep)

scascar:
Adding 
``` csharp     
     movementVector += Vector3.Cross(hitNormal,Vector3.Cross(hitNormal, Vector3.up));
```
In the Update method when the player is on a slope could make it slide along the hit normal tangent.

atb-brown:
@pgpais @lazarovttac 
> This doesn't seem to provide the full fix. The player is supposed to slide down at certain slopes (as if falling). This only disables jumping.
> 
> The player should slide down until the slope angle is "acceptable" to jump, if I understood correctly.

It looks like there's another PR which is attempting to do this: #23 

Maybe there needs to be some consolidation between these two PRs?

UOP1_Project/Assets/Scripts/Characters/Character.cs
MileyHollenberg, line 135: Could you remove the outer parenthesis still? They're not needed here

ciro-unity:
Currently this PR doesn't fix the issue, as noted by others.

ciro-unity:
Closing as it doesn't fix the issue, and locks the player from jumping while not moving it.

PR #13, opened 2020-09-30
Spawn System v1

Mufio:
Added a simple spawn point system for the player, based on 2 prefabs:
-**SpawnPoint** (labelled markers which can be placed anywhere in the scene)
-**PlayerLevelChanger** (the Singleton spawn point manager and interface to the spawning system)

**Usage:**
1) Simply place a _PlaterSpawnSystem>PlayerLevelChanger_ in the scene, along with n _SpawnPoints_.

2) Label the spawn points appropriately in their _PlayerSpawnPoint_ component.

3)
Call _PlayerLevelChange.main.SpawnPlayerAtSpawnPoint(currPlayer, "spawn pt label");_ from any script.
_OR_
Set the "Spawn Label" field on the _PlayerLevelChanger_'s _PlayerLevelChange_ component and hit the "Spawn" button in the game window.

Tried to make it as simple and modular as possible, hopefully this meets the requirements :)

### DIFF
--- a/UOP1_Project/Assets/PlayerSpawnSystem/PlayerLevelChange.cs
+++ b/UOP1_Project/Assets/PlayerSpawnSystem/PlayerLevelChange.cs
@@ -1,0 +1,94 @@
+﻿using System.Collections.Generic;
+using UnityEngine;
+using UnityEngine.SceneManagement;
+
+
+//The main "Spawn System" management class. Call the
+public class PlayerLevelChange : MonoBehaviour
+{
+    public static PlayerLevelChange main;
+    [SerializeField] List<PlayerSpawnPoint> spawnPoints = new List<PlayerSpawnPoint>(); //Contains all the spawn points in the current scene
+
+    void Awake()
+    {
+        //========== Singleton design stuff ==========
+        if (main == null)
+        {
+            DontDestroyOnLoad(gameObject);
+            main = this;
+        }
+        else if (main != this)
+        {
+            Destroy(gameObject);
+        }
+
+        //========== Initialization ==========
+        SceneManager.sceneLoaded += OnSceneLoaded;
+    }
+
+    void OnSceneLoaded(Scene scene, LoadSceneMode mode)
+    {
+        ClearSpawnPoints();
+    }
+
+    void ClearSpawnPoints()
+    {
+        spawnPoints.Clear();
+    }
+
+    public void AddSpawnPoint(PlayerSpawnPoint sp)
+    {
+        spawnPoints.Add(sp);
+    }
+
+    //Teleports/'Spawns' the player at the specified spawn point in the current scene
+    public void SpawnPlayerAtSpawnPoint(Protagonist player, string spawnPointLabel)
+    {
+        foreach(PlayerSpawnPoint sp in spawnPoints)
+        {
+            if(sp.label == spawnPointLabel)
+            {
+                Spawn(player, sp.transform.position, sp.transform.rotation);
+                return;
+            }
+        }
+    }
+
+    void Spawn(Protagonist player, Vector3 pos, Quaternion rot)
+    {
+        player.GetComponent<CharacterController>().enabled = false; //Disable the controller due to it interfering with changing the position directly
+        player.transform.position = pos;
+        player.GetComponent<CharacterController>().enabled = true;
+
+        player.transform.rotation = rot;
+    }
+    
+    public string[] GetSpawnPointLabels()
+    {
+        string[] res = new string[spawnPoints.Count];
+        int i = 0;
+        foreach(PlayerSpawnPoint sp in spawnPoints)
+        {
+            res[i] = sp.label;
+            i++;
+        }
+        return res;
+    }
+
+    //For testing purposes: Type the spawn point label into the "Spawn Label" field in the inspector, and hit the "Spawn" button in the game window
+    //"Labels" prints all spawn points to the console
+    [SerializeField] string spawnLabel = "";
+    void OnGUI()
+    {
+        GUI.backgroundColor = Color.white;
+        if(GUI.Button(new Rect(10, 10, 50, 30), "Spawn"))
+        {
+            SpawnPlayerAtSpawnPoint(GameObject.Find("Pig").GetComponent<Protagonist>(),spawnLabel);
+        }
+        if(GUI.Button(new Rect(10, 70, 50, 30), "Labels"))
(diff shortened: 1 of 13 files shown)


UOP1_Project/Assets/PlayerSpawnSystem/PlayerLevelChange.cs
```diff
@@ -0,0 +1,94 @@
...
+    public void SpawnPlayerAtSpawnPoint(Protagonist player, string spawnPointLabel)
+    {
+        foreach(PlayerSpawnPoint sp in spawnPoints)
+        {
+            if(sp.label == spawnPointLabel)
+            {
+                Spawn(player, sp.transform.position, sp.transform.rotation);
+                return;
+            }
+        }
```
MileyHollenberg, line 54: This can be reduced to a single LINQ call, here's a suggestion

```suggestion
        PlayerSpawnPoint sp = spawnPoints.First(point => point.label == spawnPointLabel);
        Spawn(player, sp.transform.position, sp.transform.rotation);
```

Since the code is setup to only allow for a single spawn point per label you might instead want to use a `Dictionary<string,PlayerSpawnPoint>` object to store all spawnpoints which takes it down from a linear search to instant access.

If there can be multiple spawn points with the same name then this code wouldn't work but at the same time that doesn't sound like a good option so keeping it focused on just 1 name seems fine to me

UOP1_Project/ProjectSettings/ProjectVersion.txt
```diff
@@ -1,2 +1,2 @@
-m_EditorVersion: 2019.4.7f1
-m_EditorVersionWithRevision: 2019.4.7f1 (e992b1a16e65)
+m_EditorVersion: 2019.4.9f1
```
MileyHollenberg, line 1: Regarding changing the Unity editor version, I've created a thread on the forum to discuss this https://forum.unity.com/threads/how-to-handle-different-unity-versions.980235/

UOP1_Project/Assets/PlayerSpawnSystem/PlayerLevelChange.cs
```diff
@@ -0,0 +1,94 @@
...
+    }
+
+    void Spawn(Protagonist player, Vector3 pos, Quaternion rot)
+    {
+        player.GetComponent<CharacterController>().enabled = false; //Disable the controller due to it interfering with changing the position directly
+        player.transform.position = pos;
+        player.GetComponent<CharacterController>().enabled = true;
```
MileyHollenberg, line 61: Please try to avoid calling `GetComponent` multiple times, it would be better to cache the `CharacterController` object inside the `Protagonist` class itself on Awake. Using `GetComponent` is quite inefficient in general and should be avoided outside of `Awake` or `Start` if at all possible, calling it twice in a row would already warrant storing a local variable for it at least as there's no use to do the same lookup twice

UOP1_Project/Assets/PlayerSpawnSystem/PlayerLevelChange.cs
Mufio: I've implemented your other suggestions, thank you for the feedback!

I figured that there would only be a couple of spawn points in the scene, that wouldn't be so large as to need a system that prevents duplicates, but perhaps I should add a warning in the console if duplicates are found, just to be safe?

UOP1_Project/ProjectSettings/ProjectVersion.txt
MileyHollenberg: Please remove these changes from the PR, they aren't required and the Unity team will update the project periodically :)

MileyHollenberg: Please remove these changes from the PR, they aren't required and the Unity team will update the project periodically :)

UOP1_Project/Assets/PlayerSpawnSystem/PlayerLevelChange.cs
MileyHollenberg: A debug warning could work, just be mindful of where you perform the check so you don't accidentally spam the console :)

SkyTech6: I think it's an interesting question in the appropriate times to use LINQ functions. With enough usage they create a noticeable overhead; they are also just implement loops with multiple iterators, dereferencing, delegate invokes, etc... in exchange for just somewhat cleaner code.

MileyHollenberg: Yeah they have to be used sparingly, with small parts like this it's mostly fine but once there are larger LINQ setups it might cause issues

UOP1_Project/ProjectSettings/ProjectVersion.txt
```diff
@@ -1,2 +1,2 @@
-m_EditorVersion: 2019.4.9f1
```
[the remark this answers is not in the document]
Mufio: My apologies. I forgot to deselect it in the original PR, so I was undoing that mistake here. °~°

UOP1_Project/Assets/PlayerSpawnSystem/PlayerLevelChange.cs
```diff
@@ -0,0 +1,94 @@
...
+    {
+        PlayerSpawnPoint sp = spawnPoints.First(point => point.label == spawnPointLabel);
+        Spawn(player, sp.transform.position, sp.transform.rotation);
+    }
+
+    void Spawn(Protagonist player, Vector3 pos, Quaternion rot)
+    {
+        player.characterController.enabled = false; //Disable the controller due to it interfering with changing the position directly
+        player.transform.position = pos;
+        player.characterController.enabled = true;
```
FabbyD, line 61: You should save the original state of the characterController and set it back to its orginal state rather than setting true directly. For all you know, this controller might have already been disabled and you just enabled it by accident.
```suggestion
        bool originalState = player.characterController.enabled;
        player.characterController.enabled = false; //Disable the controller due to it interfering with changing the position directly
        player.transform.position = pos;
        player.characterController.enabled = originalState;
```

ciro-unity:
Unfortunately this PR seems to be more of a teleport than a spawn system, as it assumes that the player is already in the scene.
Also not a super fan of the Singleton approach (although no mistake in itself!).

Great that you made it simple to start with, though!

PR #14, opened 2020-09-30
Spawn System

EmreBugday99:
In addition of spawning the character on scene changes + with a publicly callable SpawnCharacter method for runtime spawnings, I also implemented an event system which gets invoked after a character is spawned in scene. The event takes the GameObject that was spawned as a parameter for furthermore runtime checks. This is an utility that I use on my own personal projects to check from my game manager which player is spawned, does the spawned player has any specific components, tags etc than act accordingly.

Here is an example of usage:  [Screenshot](https://i.ibb.co/r25GWyG/image.png)

### DIFF
--- a/UOP1_Project/Assets/Scripts/SpawnManager.cs
+++ b/UOP1_Project/Assets/Scripts/SpawnManager.cs
@@ -1,0 +1,91 @@
+﻿using System;
+using UnityEngine;
+using UnityEngine.SceneManagement;
+
+[Serializable]
+public struct SpawnData
+{
+    //I couldn't decide on whether to use build index or scene name. Please let me know if any change is needed either from GitHub or from emreb25@outlook.com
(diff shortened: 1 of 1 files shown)


MileyHollenberg:
There's also another PR open that claims to implement this: #13

EmreB99:
> There's also another PR open that claims to implement this: #13 

Two implementation does completely different things. That implementation only teleports an already existing player to a different location. My implementation spawns a character as if it was not existing on SceneLoad. These two systems are completely different and not to mention that the PR you mentioned does not support Instantieting new players on scene changes, it is required to have the player on that scene before hand.

ciro-unity:
Not a super fan of the Singleton approach.
The manual coordinates could use a reference to a GameObject in the scene, much better for faster level design + you get to see the spawn point in the scene.

In the end I ended up merging #25, which was a bit more lightweight. Thanks anyway!

PR #17, opened 2020-09-30
Add Camera Sensitivity

atb-brown:
This commit adds a variable to CameraManager.cs that will allow the
sensitivity/speed of the player camera to be modified. By default, the
camera sensitivity is now quicker.

### Note: I'd love to learn how to write some unit tests for this; if someone knows how to do that, let me know.

### DIFF
--- a/UOP1_Project/Assets/Scripts/CameraManager.cs
+++ b/UOP1_Project/Assets/Scripts/CameraManager.cs
@@ -9,6 +9,15 @@ public class CameraManager : MonoBehaviour
 	public InputReader inputReader;
     public CinemachineFreeLook freeLookVCam;
 
+ 	[Tooltip("General multiplier for camera sensitivity/speed")]
+	public float cameraSensitivity = 7.0f;
+
+ 	private void Awake()
+    {
+		// Don't allow the camera to be slower than default and then also 20+ gets pretty crazy.
+        cameraSensitivity = Mathf.Clamp(cameraSensitivity, 1.0f, 20.0f);
+    }
+
 	private void OnEnable()
 	{
 		inputReader.cameraMoveEvent += OnCameraMove;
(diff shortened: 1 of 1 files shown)


UOP1_Project/Assets/Scripts/CameraManager.cs
```diff
@@ -9,6 +9,15 @@ public class CameraManager : MonoBehaviour
 	public InputReader inputReader;
     public CinemachineFreeLook freeLookVCam;
 
+ 	[Tooltip("General multiplier for camera sensitivity/speed")]
+	public float cameraSensitivity = 7.0f;
+
+ 	private void Awake()
```
atb-brown, line 15: I put this in `Awake` instead of `Update` because

1. I figured this shouldn't be changing.
2. I figured it'd save us from doing one more thing each frame.

UOP1_Project/Assets/Scripts/CameraManager.cs
```diff
@@ -9,6 +9,15 @@ public class CameraManager : MonoBehaviour
...
+    {
+		// Don't allow the camera to be slower than default and then also 20+ gets pretty crazy.
+        cameraSensitivity = Mathf.Clamp(cameraSensitivity, 1.0f, 20.0f);
```
guilhermeCoutinho, line 18: I dont think we need this sanity check. One should not commit/approve a change in the `cameraSensitivity` if it breaks the camera. 

You can still "tell" people that you want them to use a value between 1 and 20 by using a range attribute over the `cameraSensitivity` field. It will make people hesitant to use values outside the range and keep the code clean.

atb-brown: Cool. Good suggestion. I'd forgotten about the range option in Unity. I'll do this.

UOP1_Project/Assets/Scripts/CameraManager.cs
MileyHollenberg: It's fine to have it here indeed, here's a reference to the Docs about Awake: https://docs.unity3d.com/ScriptReference/MonoBehaviour.Awake.html

Specifically the sentence: "Use Awake to initialize variables or states before the application starts." Applies directly here :)

MileyHollenberg:
Looks good to me, @ciro-unity what's your thought on this?

ciro-unity:
Looks good. I have made the field private and added the `SerializeField` attribute, which is the practice we're going to adopt.
Thanks Austin, and Guilherme and Miley for reviewing.

Merging.

PR #19, opened 2020-10-01
Setup Pig's animator controller states

CharlExMachina:
Added animator controller states for the Pig's animator component, which includes "Idle", "Walking", "Jumping", "Attack", "Hit" and "Fainted".

States for walking, idle and jump are already being updated in Character component accordingly.

### DIFF
--- a/UOP1_Project/Assets/Scripts/Characters/Character.cs
+++ b/UOP1_Project/Assets/Scripts/Characters/Character.cs
@@ -28,6 +28,7 @@ public class Character : MonoBehaviour
     private void Awake()
     {
         characterController = GetComponent<CharacterController>();
+        characterAnimator = GetComponent<CharacterAnimator>();
     }
 
     private void Update()
--- a/UOP1_Project/Assets/Scripts/Characters/Character.cs
+++ b/UOP1_Project/Assets/Scripts/Characters/Character.cs
@@ -6,16 +6,7 @@ using UnityEngine;
 public class Character : MonoBehaviour
 {
     private CharacterController characterController;
-    private Animator animator;
-
-    #region Animator_Hashes
-
-    private int isWalking = Animator.StringToHash("IsWalking");
-    private int isGrounded = Animator.StringToHash("IsGrounded");
-    private int receivedHit = Animator.StringToHash("ReceivedHit");
-    private int isAttacking = Animator.StringToHash("IsAttacking");
-
-    #endregion
+    private CharacterAnimator characterAnimator;
 
     [Tooltip("Horizontal XZ plane speed multiplier")] public float speed = 8f;
     [Tooltip("Smoothing for rotating the character to their movement direction")] public float turnSmoothTime = 0.2f;
@@ -37,7 +28,6 @@ public class Character : MonoBehaviour
     private void Awake()
     {
         characterController = GetComponent<CharacterController>();
-        animator = GetComponent<Animator>();
     }
 
     private void Update()
@@ -46,12 +36,12 @@ public class Character : MonoBehaviour
         //This is 0f at the beginning of a jump and will raise to maximum 1f
         if (!characterController.isGrounded)
         {
-            animator.SetBool(isGrounded, false);
+            characterAnimator.SetIsIdle(false);
             gravityContributionMultiplier += Time.deltaTime * gravityComebackMultiplier;
         }
         else
         {
-            animator.SetBool(isGrounded, true);
+            characterAnimator.SetIsIdle(true);
         }
 
         //Reduce the influence of the gravity while holding the Jump button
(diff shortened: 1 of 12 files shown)


UOP1_Project/Assets/Scripts/Characters/Character.cs
```diff
@@ -72,6 +88,15 @@ private void Update()
             //Full speed ground movement
             movementVector = inputVector * speed;
 
+            if ((Math.Abs(movementVector.x) > 0.01f || Math.Abs(movementVector.z) > 0.01f) && characterController.isGrounded)
```
chriswingler, line 91: You can set these conditions as variables to reduce some duplication and make it more readable.  Also, what is the significance of 0.01f, rather than say 0.001f? ( New to Unity / Maybe that's a Unity thing? )

HyagoOliveira: This is probability to deal with analogical deadzones. Since they are using the new input system and there is a property for deadzones, this `0.01f` could be replaced for 0F

HyagoOliveira: Also, you could rewrite this part as follow:
```suggestion
            bool isMoving = Math.Abs(inputVector.sqrMagnitude) > 0F;
            if (isMoving && characterController.isGrounded)
```

MileyHollenberg: Writing it like that could introduce floating point error bugs. The reason for checking if the absolute value is greater than 0.01 is to avoid a floating point error from technically making the number greater than 0 even though in practical terms it is 0. It would be better to either define a threshold (which that 0.01f would be) or to use `float.Epsilon` which is a very small value that's specifically used for checks like this, here's a [StackOverflow post](https://stackoverflow.com/questions/30216575/why-float-epsilon-and-not-zero) that explains it better

UOP1_Project/Packages/manifest.json
```diff
@@ -1,11 +1,11 @@
 {
   "dependencies": {
-    "com.unity.addressables": "1.8.4",
+    "com.unity.addressables": "1.8.5",
```
MileyHollenberg, line 3: Please avoid updating packages in PRs

UOP1_Project/ProjectSettings/QualitySettings.asset
```diff
@@ -21,7 +21,7 @@ QualitySettings:
     skinWeights: 2
     textureQuality: 0
     anisotropicTextures: 1
-    antiAliasing: 0
+    antiAliasing: 2
```
MileyHollenberg, line 24: Please avoid changing quality settings when they are not related to the PR

UOP1_Project/Assets/Scripts/Characters/Character.cs
```diff
@@ -6,6 +6,16 @@
 public class Character : MonoBehaviour
 {
     private CharacterController characterController;
+    private Animator animator;
```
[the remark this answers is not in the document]
MileyHollenberg: Plus one to that idea, it would be better to separate the animations from the mechanics :)

UOP1_Project/Assets/Scripts/Characters/Character.cs
CharlExMachina: I used `float.Epsilon` instead of 0.01f to avoid using magic numbers, and created a separate variable to check if the player is moving or not, instead of having the whole comparison on the if statement

UOP1_Project/Packages/manifest.json
CharlExMachina: Reverted to original state

UOP1_Project/ProjectSettings/QualitySettings.asset
CharlExMachina: Reverted to original state

UOP1_Project/Assets/Scripts/Characters/Character.cs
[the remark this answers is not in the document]
CharlExMachina: Created a separate component to take care of animation state changes. Implementation is very basic for now

UOP1_Project/Assets/Scripts/Characters/Character.cs
```diff
@@ -37,7 +28,6 @@ public class Character : MonoBehaviour
     private void Awake()
     {
         characterController = GetComponent<CharacterController>();
```
HyagoOliveira, line 30: You need to cache `characterAnimator` here.
```suggestion
        characterController = GetComponent<CharacterController>();
        characterAnimator= GetComponent<CharacterAnimator >();
```

HyagoOliveira (changes requested):

UOP1_Project/Packages/packages-lock.json
```diff
@@ -150,7 +150,8 @@
       "depth": 0,
       "source": "builtin",
       "dependencies": {
-        "com.unity.modules.ui": "1.0.0"
+        "com.unity.modules.ui": "1.0.0",
+        "com.unity.modules.imgui": "1.0.0"
```
MileyHollenberg, line 154: These changes are not needed

ciro-unity:
Closing this since we are leaning towards more fully-featured state machine implementations (see other PRs).

PR #20, opened 2020-10-01
Inventory System Prototype

trevorjmcdonald:
### Overview

Characters have an **InventoryController** component that references a **CharacterInventory** ScriptableObject. The **CharacterInventory** maintains a list of **ItemStack**, which is a classic C# class that has a quantity, and references an **InventoryItem** scriptable object. An **InventoryItem** has a name, description, icon, one or more **ItemEffect** ScriptableObject references, and whether or not it is usable. I included a prefab on it but had no use for it at the moment.

For the demo, I created a scene (Scenes/InventoryDemo) which is a duplicate of the CharController demo, except that it has a (very preliminary) UI panel displaying inventory items. This panel is automatically updated by a UnityEvent invocation from the InventoryController (when it detects that it has picked up an item.) Currently, it is using a trigger collision event to cause this, but I set the methods to be public for when we wire up the pickups to work with them. I have some icons I made myself in Aseprite, they are terrible.

### Possible Risks or Issues with Design

There is no unit testing assembly as far as I am aware. Thus, I was unable to write unit tests for this. I tested as well as I could, given the circumstances.

The inventory uses a List. I ultimately felt that it was a slightly nicer management experience on the Editor side. That being said, we'd probably want to use a custom editor anyways.

I have zero experience working on a system like this, so I could have easily missed something in my implementation.

### Components

Of note: I created a namespace for this to keep it separated. I put it under Inventory (and also have Inventory.Effects.) They can be removed if preferred.

**CharacterInventory** - A ScriptableObject with a collection of inventory items. It has the code necessary to add and remove items (this code could be moved to the InventoryController - however, I assumed that the controller is more related to the player and game world interactions and not necessarily the responsibility of managing stacks and whatnot. Feel free to correct me on this.) I'm not thrilled with the name, feel free to suggest others.

**InventoryController** - A MonoBehaviour used to assign a CharacterInventory to a character. It has a public method for picking up a stack of items (which it communicated to the CharacterInventory SO.) It has a public method that allows it to attempt to use an item. It has the concept of an item target (currently just set to the game object itself but there is a public method for setting it - in case we want to allow for using items on other entities.)

**InventoryItem** - A ScriptableObject that represents a specific item instance. It includes MaxStackSize, one or more item effects (which are ScriptableObjects as well.)

**InventoryLayout** - For demo purposes. I probably should have put it in a separate folder. I wanted the ability to demonstrate the inventory (for viewers and for myself.)

**ItemPickup** - A stack of items on the ground. I have a couple of prefabs - carrots and health potions. Right now, it just uses a collider, but the methods are public for when we want to change it to use an interact button.

**ItemStack** - POCO for referencing an item and a quantity.

**ItemStackIcon** - MonoBehaviour for demo purposes, it just allows me to update the quantity on items in the demo panel.

**ItemEffect** - Abstract ScriptableObject for creating effects. It has an ApplyEffect public method with a GameObject target. We probably want to make this more specific.

**HealingEffect** - ScriptableObject implementing ItemEffect, it's what allows the healing potion to heal.

### Usage

Create items via the right-click menu - I put them under Inventory. Create a character inventory via the same menu. You can create healing effects and assign them to items. Add the ItemPickup MB to game objects in the world or prefabs, assign an item and a stack size. Add the InventoryController to the player character. Run over the pickups, watch the inventory go up, stacks will automatically split and be added.


trevorjmcdonald:
I'm going to close this for now so that we can further discuss the design of this system.

UOP1_Project/Assets/Scripts/Inventory/InventoryLayout.cs
```diff
@@ -0,0 +1,58 @@
...
+        {
+            _icons = GetComponentsInChildren<ItemStackIcon>().ToList();
+            if (_icons.Count < _inventory.Inventory.MaxInventorySize)
+            {
+                for (var i = _icons.Count; i < _inventory.Inventory.MaxInventorySize; i++)
+                {
+                    _icons.Add(Instantiate<ItemStackIcon>(_itemStackIconPrefab, transform));
+                }
+            }
+            else if (_icons.Count > _inventory.Inventory.MaxInventorySize)
```
Dustin00, line 46: Shouldn't this be 
`
else if (_icons.Count >= ...
`

PR #21, opened 2020-10-01
Spawn System

DaiganIT:
The player is spawned on Awake or by the use of a public function (could be triggered by a UnityEvent later on)

The only dependency required is the Player Prefab.

InputReader and Camera are searched in the scene.
The Spawn Point GO is used as Follow and LookAt by the cameras, and it's added to the player after the spawn.

It's very simple


MileyHollenberg:
There's also another PR open that claims to implement this: #13

UOP1_Project/Assets/Scripts/PlayerSpawnPoint.cs
```diff
@@ -0,0 +1,48 @@
+﻿using UnityEngine;
+
+/// <summary>
+/// Spawns the player at its own location
+/// </summary>
+public class PlayerSpawnPoint : MonoBehaviour
```
HyagoOliveira, line 6: This could be a generic spawner.

UOP1_Project/Assets/Scripts/PlayerSpawnPoint.cs
```diff
@@ -0,0 +1,48 @@
...
+
+    /// <summary>
+    /// Spawns the player and puts this GO as child of the newly spawned player.
+    /// </summary>
+    public void SpawnPlayer()
+    {
+        var player = Instantiate(playerPrefab, transform.position, transform.rotation);
+        transform.parent = player.transform;
+        var protagonistComponent = player.GetComponent<Protagonist>();
+        protagonistComponent.Initialise(inputReader, gameplayCamera);
```
HyagoOliveira, line 46: Initiate components inside the player should be a responsibility of the player itself.
Your spawn script should only do one thing: spawn

UOP1_Project/Assets/Scripts/PlayerSpawnPoint.cs
DaiganIT: Correct, but it's named Player only because it's doing some Player-Only relevant stuff.

UOP1_Project/Assets/Scripts/PlayerSpawnPoint.cs
DaiganIT: You are not wrong about this. The spawn script should only do the spawn.
But to achieve what you say, the Protagonist script should be changed upfront. Not sure it's in the scope of this task.

HyagoOliveira: Yes, that's why I think your script should only spawn a generic prefab in the scene.
The Protagonist script (or any other one inside the player GameObject) should initialize those components for the player. If so, your script will not need to be refactored in the future.

DaiganIT: let me try and we'll see what the others think

UOP1_Project/Assets/Scripts/PrefabSpawnPoint.cs
```diff
@@ -0,0 +1,45 @@
...
+    [SerializeField]
+    [Tooltip("The Prefab of the spawn")]
+    GameObject prefabToSpawn;
+
+    /// <summary>
+    /// Gather external components. Throws if it can't find them.
+    /// </summary>
+    void Awake()
+    {
+        if (prefabToSpawn == null)
```
FabbyD, line 23: No need to check it there too. You already do it in SpawnPrefab

UOP1_Project/Assets/Scripts/PrefabSpawnPoint.cs
```diff
@@ -0,0 +1,45 @@
...
+    /// </summary>
+    public void SpawnPrefab()
+    {
+        if (prefabToSpawn == null)
+        {
+            Debug.LogError("No prefab to spawn");
+            return;
+        }
+
+        var player = Instantiate(prefabToSpawn, transform.position, transform.rotation);
```
FabbyD, line 41: Instead of spawning and assigning the parent afterwards, you just use the Instantiate(prefab, transform) override which instantiates the prefab as a child of the transform directly.

ciro-unity:
Simple implementation, does what it needs to do.
In the end I merged #25, which had a nice Inspector trick... even though I stripped that one down considerably too.

Thanks anyway!

PR #23, opened 2020-10-01
Fix > Player now slide down steep slopes

Mdotta:
I've made a PR yesterday with this fix but I commited my scene with some objects that where not needed. This PR has only the changes to Character.cs and one object was added to Pig.prefab that is needed to calculate a falling direction to slide down

### DIFF
--- a/UOP1_Project/Assets/Scripts/Characters/Character.cs
+++ b/UOP1_Project/Assets/Scripts/Characters/Character.cs
@@ -6,7 +6,7 @@ using UnityEngine;
 public class Character : MonoBehaviour
 {
     private CharacterController characterController;
-
+    [Tooltip("Vertical falling direction reference object")] public Transform fallDirection;
     [Tooltip("Horizontal XZ plane speed multiplier")] public float speed = 8f;
     [Tooltip("Smoothing for rotating the character to their movement direction")] public float turnSmoothTime = 0.2f;
     [Tooltip("General multiplier for gravity (affects jump and freefall)")] public float gravityMultiplier = 5f;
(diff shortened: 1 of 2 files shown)


RunninglVlan:
I don't think Unity will be able to accept your PR without you signing the Contributor License Agreement.

Mdotta:
> I don't think Unity will be able to accept your PR without you signing the Contributor License Agreement.
> BTW, in previous PR you asked if it's possible to update its contents. As far as I know it should be easily possible, just add commits to your repository's branch.

I've already signed but it still shows as unsigned.
Thanks! It'd have been easier that way

UOP1_Project/Assets/Scripts/Characters/Character.cs
```diff
@@ -102,6 +125,9 @@ private void Update()
     
     private void OnControllerColliderHit(ControllerColliderHit hit)
     {
+        //updates where the collider is hiting;
+        collisionPoint = hit.point;
+        collisionPoint = (collisionPoint - transform.position);
```
atb-brown, line 130: Just asking so that I can learn, but is this figuring out the collisionPoint relative to the character? (As opposed to the global coordinates of the collision.)

atb-brown:
https://github.com/Mdotta/open-project-1/blob/master/UOP1_Project/Assets/Prefabs/Pig.prefab#L1003

I was playing around with your branch. This might be an appropriate time to increase the slope limit. 30 seems too limiting. When I was walking around, 50 seemed to feel a lot more natural. But I'm curious what others think.

Mdotta:
> https://github.com/Mdotta/open-project-1/blob/master/UOP1_Project/Assets/Prefabs/Pig.prefab#L1003
> 
> I was playing around with your branch. This might be an appropriate time to increase the slope limit. 30 seems too limiting. When I was walking around, 50 seemed to feel a lot more natural for a more "whimsical" experience. But I'm curious what others think.

While testing I was playing with slope limit and i got that feeling too, I think 50-55 is a great value

UOP1_Project/Assets/Scripts/Characters/Character.cs
CharlExMachina: Not OP but yeah, it's figuring out where it did collide, in this case, relative to the GO's position

ciro-unity:
CLA still not signed

ciro-unity:
Sorry Mdotta, closing this since I found a leaner fix, #26 
Thanks anyway!

PR #25, opened 2020-10-01
Simple Player Spawn System

Wodopo:
Card: https://open.codecks.io/unity-open-project-1/decks/15/card/15q-spawn-system
Forum: https://forum.unity.com/threads/spawn-system.980658/

Usage: 
- Drag and drop the prefab to each gameplay scene.
- As an extra: Right Click the Spawn System (Script) attached to the scene object and select Attempt Auto Fill to try and auto fill the Protagonist's dependencies.
- Set all desired spawn points as Transform references. It will align the Protagonist's position and rotation.
- Set the `Default Spawn Index` for use when entering Play Mode.
- Use the OnPlayerSpawnEvent to set any Protagonist's related dependency, like: CM FreeLook1's Follow and Look At transform references.
- Call the static method `SpawnSystem.SetSpawnIndex(index)` before triggering a scene load.

Example setup:
image


UOP1_Project/Assets/Scripts/SpawnSystem.cs
```diff
@@ -0,0 +1,104 @@
+﻿using System;
+using System.Linq;
+using UnityEditor.PackageManager;
+using UnityEngine;
+using UnityEngine.Events;
+
+[Serializable]
+public class PlayerSpawnEvent : UnityEvent<Transform> { }
```
chriswingler, line 8: New to C# and Unity, why is this empty class here? Is there a better way to inherit from UnityEvent<Transform> ?

Wodopo: It's done so it can be serialized and show up in the inspector.
Doing `public UnityEvent<Transform> onPlayerSpawnEvent;` would work the same way at C# level but can't be manipulated by the level designer in the editor.

richard-fine: FWIW, it's no longer necessary in 2020.1, but this project is using 2019 LTS so indeed it's still necessary for now.

UOP1_Project/Assets/Scripts/SpawnSystem.cs
```diff
@@ -0,0 +1,104 @@
...
+
+public class SpawnSystem : MonoBehaviour
+{
+    private static int requestedSpawnIndex = -1;
+
+    [Header("Settings")]
+    [SerializeField] private int defaultSpawnIndex = 0;
+
+    [Header("Project References")]
+#pragma warning disable 649
```
FabbyD, line 18: Is this to avoid the nasty warning where unity says this value is never initialized? Could we disable this warning by default for the project instead? This can be done a csc.rsp file with the following:

-nowarn:0649

UOP1_Project/Assets/Scripts/SpawnSystem.cs
```diff
@@ -0,0 +1,104 @@
...
+    [SerializeField] private InputReader inputReader;
+    [SerializeField] private Transform gameplayCamera;
+    [SerializeField] private Transform[] spawnLocations;
+
+    [Header("Events")]
+    public PlayerSpawnEvent onPlayerSpawnEvent;
+
+    void Awake()
+    {
+        try
```
FabbyD, line 32: I'ld rather see a full stacktrace in the logs rather than hiding an exception :/

Since this is run in Awake(), it won't prevent any other scripts from running.

FabbyD: After reading through the rest of the code, I see this is how you centralized your error management. I'm not exactly sure how I feel about it. While it is nice and tidy, I've never been fond of using exceptions to return "expected" errors. I would personally prefer that the Spawn method return a boolean for success since exceptions can be a bit more destructive.

UOP1_Project/Assets/Scripts/SpawnSystem.cs
```diff
@@ -0,0 +1,104 @@
...
+        {
+            Spawn();
+        }
+        catch (Exception e)
+        {
+            Debug.LogError($"[SpawnSystem] Failed to spawn player. {e.Message}");
+        }
+    }
+
+    void Reset()
```
FabbyD, line 42: What is this called from?

richard-fine: Reset() is one of Unity's magic methods. It's called automatically when the component is newly added to a GameObject.

FabbyD: Oh wow. How have I never seen it before?! Thanks, good to know! :)

UOP1_Project/Assets/Scripts/SpawnSystem.cs
ciro-unity: I also don't think this is necessary.

ciro-unity:
I like the AutoFill function on the right click! Nice idea.

In the end, I applied a few changes:
- There was an error in the Prefab (script was missing).
- Removed the pragma ignore, and implemented initialisation instead.
- Changed spaces into Tabs.
- Removed all static methods and variables as they were not needed at this point.

I also merged it with another PR #27 that was changing the Input Reader from a script in the scene to a ScriptableObject, which simplified your script too. You'll see it shortened quite a while, and I was able to remove the enable/disable trick which felt like a hack.

PR #27, opened 2020-10-01
InputReader made into ScriptableObject asset

andrew-raphael-lukasik:
What:
Turns scene singleton `InputManager` from `Monobehaviour` into `ScriptableObject`

Why:
Referencing **asset singletons** reduces project complexity while **scene singletons** increase it.


ciro-unity:
Love the idea! Great contribution.

PR #30, opened 2020-10-02
Settings system

Acimaz:
Added the following: 

- General - Language
- Graphics - Graphics Presets, Resolution (all supported resolutions on users device), fullscreen mode,
- Advanced Graphics - Anisotropic filtering, Anti-Aliasing, Shadow Distance, Shadow Quality
- Audio - Music Volume, Sfx Volume
- Simple UI for **testing** purposes

Unitys "QualitySetting.SetQualityLevel()" with true for the applyExpensiveChanges parameter does not change the anisotropic filtering, anti-aliasing, shadow distance, and shadow quality settings. 
It defaults them to some arbitrary value: anti-aliasing becomes 0, shadow distance becomes 40

This happens although I've set up 3 graphics presets (low, middle, high) with different settings, none of them have the values unity is defaulting to.

When changing the value by hand and hitting the save button the settings are updated correctly.

To fix the preset issue we might have to create preset values by hand (ScriptableObject) and setting each value by hand (like it's already happening in code for custom settings, instead of depending on the "SetQualityLevel()" method, which unfortunately does not what it is supposed to.


Acimaz:
Here are a few screenshots from within the editor. The labels and names are ofc all WIP and subject to change:

Screenshot 2020-10-02 at 02 11 05

Screenshot 2020-10-02 at 02 12 02

Screenshot 2020-10-02 at 02 12 08

MileyHollenberg:
Haven't looked at the actual content yet but I noticed that you added Text Mesh Pro to the project in this PR as well. Whilst I definitely support having TMPro in it I don't think it's in the scope of this PR. 

Could you instead create a separate PR with the sole purpose of adding TMPro to the project and then rebase this PR on top of master once it's merged in? That would help reduce the scope of the 25K lines changes that this PR would otherwise bring

Acimaz:
Hm, I'll try to create a separate branch where I will import TMPro and make a PR. Do you want to separate the import of plugins/packages from the actual features?

MileyHollenberg:
Yes, it's already one of the contribution rules to not add new packages to the project but TMPro is basically inevitable. A PR itself should only make the changes it sets out to do but not add any packages alongside it

Acimaz:
Alright, that's understandable. Though it would've been a lot of work to later remove legacy text components and replace them with the TMPro components. I Shouldve thought of making a separate PR 👍 

I just made a new PR with the TMPro import. And by the way, the package itself was in the project already. What I've done now is to import the TMPro essentials. More info here https://github.com/UnityTechnologies/open-project-1/pull/37

ciro-unity:
Thanks for the great work on this, @Acimaz.

From your first comment it seems like the menu doesn't work fully. Did you manage to resolve those issues in the end?

Also, I noticed the menu references an ActionMap which is within the folder of the InputSystem package. I think we should reference the ActionMap we are using for the game.

Looks good otherwise!

Acimaz:
Ah the action map thingy is due to the fact that the eventsystem spit out exceptions because the project seems to use another input system (the new one I guess?). I had to click a button which used the new input system to get the scene and testing going without throwing exceptions. Maybe someone else activated or imported the new input system in some way?
I followed the instructions the editor gave me in the eventsystem object to get rid of them (as I said, there appeared a button to fix it).

Acimaz:
And no, I couldn't fix the SetQualityLevel issue. The method simply doesn't work as intended.

I think we might have to make a workaround. I have an idea for that to create structs which hold the low mid and high settings and can be changed inside the inspector as a list. That way it should apply the changes correctly. Scriptable objects would be a solution too. Going to think about those two ideas a bit further and make an update to this PR after that.

Acimaz:
Also, I think I will do a bug report regarding this SetQualityLevel issue to unity.

Acimaz:
And if there is an action map used for the game I will set the correct reference as well 👍🏼

ciro-unity:
> And if there is an action map used for the game I will set the correct reference as well 👍🏼

There is one in the project but it doesn't contain any mapping for UI just yet. And sorry: I should have said "Input Action Assets".

Then, it requires a new Action Map for the UI. Check this page: https://docs.unity3d.com/Packages/com.unity.inputsystem@1.0/manual/UISupport.html

UOP1_Project/Assets/Scripts/Systems/SettingsSystem.cs
```diff
@@ -0,0 +1,278 @@
+﻿using System;
+using System.Collections;
+using System.Collections.Generic;
+using System.Linq;
+using TMPro;
+using UnityEngine;
+using UnityEngine.Rendering;
+using UnityEngine.UI;
+
+public class SettingsSystem : MonoBehaviour
```
richard-fine, line 10: I'm not sure that having a single class be responsible for all the different kinds of settings is a great idea... there's a lot of different things in here, and if we add more settings over time this class will become a bit of a dumping-ground.

Have you considered making separate components for each of the different kinds of control - e.g. a ResolutionsDropdownController component, which you attach to the resolutions dropdown GameObject, a LanguageDropdownController component that you attach to the languages dropdown GameObject, etc?

You could still have the SettingsSystem provide a facade for accessing all the settings if you want, but the logic around how to populate each setting and how to apply it would be isolated into separate classes that would be easier to maintain.

Acimaz: In my experience, separating a logical block as "Settings Systems" into too small pieces is making it too complex to maintain. Especially when adding more settings later on. Each having its own class with maybe 1-3 lines doesn't make it worth it for me to create dozens of classes, instead of letting one settings class do the job.

Also, components like "ResolutionsDropdownController" would only redirect a method call to the Main settings class, which you can skip and directly call in on said settings class.

But splitting it up into Sub classes like "GeneralSettingsComponent", "GraphicsSettingsComponent", "AudioSettingsComponent" etc would be an option. Currently these logical blocks are separated by a #region statement in the SettingsSystem class.

richard-fine: > Also, components like "ResolutionsDropdownController" would only redirect a method call to the Main settings class, which you can skip and directly call in on said settings class.

How so?

Acimaz: Assuming that the settings are not saved right away, but fist saved when hitting the save button (advanced graphics settings), you would have to temporarily save the settings in some sort of variable.

So of course there needs to be a central class holding this information and that would be the Main Settings class. That's why there has to be a redirection.

Without redirection, each component has to handle the save/cancel operation itself, which leads to the same code in dozens of small components. Also loading previously saved data would have had the same problem. All in all there are several more reasons which add up to a more maintainable "One Class"-Solution.

MEGATANK58: when you referred to storing data when the button isn't clicked maybe we can get the data straight from the object?

We will obviously be changing the properties of the object to give visual response but we can take the data straight from the Object it self that has been changed but getting the text or anything like that instead of making variables?

richard-fine: > Assuming that the settings are not saved right away, but fist saved when hitting the save button (advanced graphics settings), you would have to temporarily save the settings in some sort of variable.

Yes...

> So of course there needs to be a central class holding this information

...no.

Each separate controller would hold that state itself. Then you'd have e.g. an Apply() method on each controller, that the overall controller for the page would call when you hit the 'Apply' button. It's not the same code in each controller because the way of applying each setting is different - one controller is calling SetQualityLevel, another is doing `vsyncCount = 1;`, etc.

Acimaz: But is it worth creating a whole class just for the sake of calling QualitySetting.antiAliasing = x? In my opinion, it's not, but this is a taste issue rather than an implementation one. Both would work equally.

Also, this "Apply" button would have to call the "apply" in a center class, which then calls it on all sub-components or you would have to add each apply call directly to the button click event itself (which honestly I'm not a big fan of because the button events are much more prone to errors when developing).

richard-fine: > But is it worth creating a whole class just for the sake of calling QualitySetting.antiAliasing = x?

Antialiasing might be that simple, but I think that in some other cases - such as resolutions - it's a bit more involved: the class would not just apply the resolution, it would also be responsible for populating the dropdown with the list of supported resolutions, identifying which resolution is the current one, converting each Resolution object to a good-looking string, etc. In the case of antialiasing, I agree there's very little code to write if we are just setting QualitySettings, but it might be more involved if we later want to use the antialiasing features in the postprocessing stack, instead of the 'built-in' antialiasing.

Note that each individual setting control would not necessarily need to have exactly one control associated with it - if we wanted to offer a dropdown for 'which AA technique' as well as a slider for 'what level' then it would probably make sense for those to be driven by the same controller component, especially if e.g. the slider is enabled/disabled based on the dropdown. For something like the input bindings it's plausible that all of them would be associated with the same controller.

> Also, this "Apply" button would have to call the "apply" in a center class, which then calls it on all sub-components or you would have to add each apply call directly to the button click event itself (which honestly I'm not a big fan of because the button events are much more prone to errors when developing).

Yes, what I imagine is that you'd have an abstract base class `SettingController` that offers the Apply() method, and each individual controller class would derive from it, implementing that method. Then the overall controller for the page could just do `GetComponentsInChildren<SettingController>()` to get them all, then loop through and call `Apply()` on each one.

Acimaz: What i see here is what our company did and learned from in the past. Overly abstracting things in the very beginning and adding more work for a feature that otherwise would have a very simple way of implementation. I don't say that it would come in handy in certain situations, but at this point we cant say for what this is going to be used. And so we should stick to "keep it simple" and save us unnececary work.

Also calling any "GetComponent" code, even if its just UI is not a good idea in general. I always try to avoid using this when I can. For example it wont work ob objects which are deactivated (which happens quite often in UI).

Acimaz:
Applying presets is working now. Using a scriptable object which holds the graphics preset data.
The class could be refactored and split up into smaller logical pieces (general, sound, graphics, input (when input system is done)).

UOP1_Project/Assets/Scripts/Systems/SettingsSystem.cs
richard-fine: >  I don't say that it would come in handy in certain situations, but at this point, we can't say for what exactly this is going to be used.

21 years of experience tells me that we're going to need it - but sure, it would be good if we had something to go on more than gut feelings. Maybe we should have done some actual design on this first 😂 

Anyway, it's your PR, I'm just here to try and help everyone learn and explore possibilities. If it's not a direction you want to go in, so be it. :)

> GetComponentInChildren doesn't work on inactive objects

It does if you pass `true` for the `includeInactive` parameter.

MileyHollenberg:
Could you pull the latest master and rebase to that? I see that the TMPro essentials are still set to be added through this PR even though that should already be in master now

Acimaz:
Hm, I've rebased it already and looking at the commit history my previous tmpro import happens after the one we've merged into master. Rebasing the branch won't make the commit go away as it's just moving all commits a few commits up and glues it to the latest master commit.The issue is that I previously committed the tmpro update and the settings implementation in one commit and now it's not that easy to split them apart.What I could do though is to completely start a new branch and put only the settings system changes on top of the master branch. Going to cancel this PR then and open another one once it's done. I should've split up the commits of tmpro and system settings in the first place! On Oct 4, 2020 12:59, Miley Hollenberg <notifications@github.com> wrote:
Could you pull the latest master and rebase to that? I see that the TMPro essentials are still set to be added through this PR even though that should already be in master now

Acimaz:
image

MileyHollenberg:
Aha, in that case I believe you can run the command `git rebase -i master` and then drop the commit that would have added TMPro, there might be some merge conflicts you need to resolve though so you may want to create a quick local backup just in case

Acimaz:
> Aha, in that case I believe you can run the command `git rebase -i master` and then drop the commit that would have added TMPro, there might be some merge conflicts you need to resolve though so you may want to create a quick local backup just in case

I just created a new clean PR which does not contain any TMPro related commits. I was already working on it when you wrote your message, so I didn't see it in time. Next time I'll remember that command, could come in handy. I think we can close this PR.

UOP1_Project/Assets/Scripts/Systems/SettingsSystem.cs
Acimaz: This is indeed a great parameter. This should be possible in Transform.Find() as well! (unfortunately it's not 😂 )

I've just created a new PR (https://github.com/UnityTechnologies/open-project-1/pull/67) which in some way utilizes what you have said already. I split up the logic in sub-components (General, Graphics, Audio, Input) which do the job of populating the dropdowns, etc.
It's not exactly how you described it, but I felt like this is more like I would implement it. Maybe we can talk about the pros and cons of this. Btw I didn't want to be rude in any way. I appreciate your contribution and feedback to my code :)

MileyHollenberg:
@Acimaz please close this PR if we're going to continue with the new one

PR #33, opened 2020-10-02
Adding simple state machine + implementing existing states

Proryanator:
Based on this discussion: https://forum.unity.com/threads/character-state-machine.979371/

What this PR brings:
- State/transition changes occur within code, not prefabs, thus tracking changes to states is simple
- Moved some chunks of code in Update() into methods to make it clearer what they accomplish
- A simple state implementation for existing states, with stubs for expandable states
- No bloat is added other than introducing a code based state machine

Recent Additions:
- after merging in recent code updates, moved 'UpdateSlide' into rightful place


UOP1_Project/ProjectSettings/EditorBuildSettings.asset
```diff
@@ -5,7 +5,7 @@ EditorBuildSettings:
   m_ObjectHideFlags: 0
   serializedVersion: 2
   m_Scenes:
-  - enabled: 1
-    path: Assets/Scenes/SampleScene.unity
-    guid: d1c3109bdb54ad54c8a2b2838528e640
+  - enabled: 0
```
ErikBehar, line 8: should probably not touch this file

UOP1_Project/Assets/Scripts/DesignPatterns/StateMachine/StateMachine.cs
```diff
@@ -0,0 +1,87 @@
...
+	{
+		if (state == _currentState)
+		{
+			return;
+		}
+		
+		_currentState?.OnExit();
+		_currentState = state;
+  
+		_allTransitionsForType.TryGetValue(_currentState.GetType(), out _transitionsForCurrentState);
```
ErikBehar, line 37: Could save transitions as part of the states and just call _currentState.transitions() whenever you need them instead.

ErikBehar:
> Should I create a MonoBehaviour script to hold all the state setup code?

Maybe a scriptable object instead.

UOP1_Project/ProjectSettings/EditorBuildSettings.asset
Proryanator: Ah thought I reverted that one sec. Surprised it's not in the git ignore good catch!

UOP1_Project/Assets/Scripts/DesignPatterns/StateMachine/StateMachine.cs
Proryanator: Done and done, quite a good suggestion.

UOP1_Project/Assets/Scripts/Characters/Character.cs
```diff
@@ -1,6 +1,7 @@
 ﻿using System;
 using System.Collections;
 using System.Collections.Generic;
+using UnityEditorInternal;
```
ErikBehar, line 4: Why did this get added ?

ErikBehar:
To expand on this, one issue I see is that currently states themselves can't be configured by a user.

One idea would be that you instantiate an instance of a specific state (SO), so the user can change the fields and drag and drop it into the state machine ... you'd need to have an id to refer to them to add the transitions though. 

Another idea, you could add Monobehaviours for states and collect them on Awake and use the runtime type to add transitions.

You could do similar with Transitions, maybe have a custom drop down for the "To" State

UOP1_Project/Assets/Scripts/Characters/Character.cs
Proryanator: What a good question 😆  I'll remove it.

Proryanator:
> To expand on this, one issue I see is that currently states themselves can't be configured by a user.
> 
> 
> 
> One idea would be that you instantiate an instance of a specific state (SO), so the user can change the fields and drag and drop it into the state machine ... you'd need to have an id to refer to them to add the transitions though. 
> 
> 
> 
> Another idea, you could add Monobehaviours for states and collect them on Awake and use the runtime type to add transitions.
> 
> 
> 
> You could do similar with Transitions, maybe have a custom drop down for the "To" State

I posted about this discussion in the main thread! Might need to re-think how the data is stored for states at some point 🤔

UOP1_Project/Assets/Scripts/DesignPatterns/StateMachine/StateMachine.cs
```diff
@@ -0,0 +1,64 @@
+﻿using System;
+using System.Collections.Generic;
+using UnityEngine;
+
+public class StateMachine
```
richard-fine, line 5: It would be nice if this could be visualized in the Inspector somehow. Seeing what state a statemachine is currently in is very often a useful thing to be able to see.

UOP1_Project/Assets/Scripts/DesignPatterns/StateMachine/StateMachine.cs
```diff
@@ -0,0 +1,64 @@
...
+
+		if (transitionIfAny == null || transitionIfAny.To == currentState)
+		{
+			transitionIfAny = GetNextTransitionFromList(this.currentState.Transitions);
+		}
+
+		return transitionIfAny;
+	}
+
+	private Transition GetNextTransitionFromList(List<Transition> transitions)
```
richard-fine, line 52: Looks like this could be `static`?

Proryanator: You are quite right! I'll modify that next.

UOP1_Project/Assets/Scripts/DesignPatterns/StateMachine/StateMachine.cs
Proryanator: Yeah I agree with you on that, since right now it's not really possible to be visualized at all.

ErikBehar: if you make a public getter for currentState you can then attach some visualizer easily enough

UOP1_Project/Assets/Scripts/Characters/Character.cs
Proryanator: Removed 🗡️

UOP1_Project/Assets/Scripts/Characters/Character.cs
```diff
@@ -29,69 +29,112 @@ public class Character : MonoBehaviour
...
+		IdleState idleState = new IdleState(this);
+		WalkingState walkingState = new WalkingState(this);
+		JumpingState jumpingState = new JumpingState(this);
+		FallingState fallingState = new FallingState(this);
+
+		// setup all explicit transitions
+		idleState.AddTransition(jumpingState, PerformedJumpAction());
+		idleState.AddTransition(walkingState, IsMovingAndGrounded());
+		walkingState.AddTransition(jumpingState, PerformedJumpAction());
+		fallingState.AddTransition(idleState, IsCharacterNotMovingAndGrounded());
```
Eyap53, line 62: I believe there is a redundacy with l.66 
(stateMachine.AddAnyTransition(idleState, IsCharacterNotMovingAndGrounded());)

UOP1_Project/Assets/Scripts/DesignPatterns/StateMachine/StateMachine.cs
```diff
@@ -0,0 +1,64 @@
...
+		return transitionIfAny;
+	}
+
+	private static Transition GetNextTransitionFromList(List<Transition> transitions)
+	{
+		foreach (Transition transition in transitions)
+		{
+			if (transition.Condition())
+			{
+				return transition;
```
Eyap53, line 58: Problem with two (or more) transitionFromAnyState : If the first is Ok but transitionIfAny.To = currentState, it is returned and the other not even tested (or used if they are valid).

ciro-unity:
Thanks @Proryanator, this was a good implementation but in the end I wanted to go for an hybrid with code + SOs.
Thanks for keeping the PR light though!! 🙏🏻 

Also are you on Discord? If so, tell me your complete nickname and I can give you the role of "Open Project Contributor". Same for @ErikBehar @Eyap53 (and @richard-fine I guess :D)

ErikBehar:
> Same for @ErikBehar 

discord:  invadererik#1799

Eyap53:
> Same for @ErikBehar @Eyap53

Discord: Eyap#1181
Thanks :)

Proryanator:
@ciro-unity

I'd love the role in Discord!

My Discord Username: Proryanator

My full name is Ryan Laseter too!

Thanks a bunch! 😃

ciro-unity:
@Proryanator Role granted!

Proryanator:
> @Proryanator Role granted!

Thank you so much! It's an honor!

PR #36, opened 2020-10-02
Added walking particle to the pig

printto:


UOP1_Project/Assets/Scripts/Characters/WalkingParticle.cs
```diff
@@ -0,0 +1,37 @@
...
+    [Tooltip("Particle prefab to be attached to the charactor in the scene.")] public ParticleSystem walkingParticlePrefab;
+
+    private Vector3 lastPosition;
+    private Vector3 currentPosition;
+    private ParticleSystem particleObject;
+
+    private void Update()
+    {
+        if (lastPosition == null) lastPosition = transform.position;
+        if (currentPosition == null) currentPosition = transform.position;
```
ManuelRauber, line 17: If you check a struct against `null` it will always be false.
Either make it nullable or check it against `default(Vector3)` (that is `Vector3.zero` in this case). The later could introduce a bug, where the position really is `default`.

UOP1_Project/Assets/Scripts/Characters/WalkingParticle.cs
```diff
@@ -0,0 +1,37 @@
...
+        if (particleObject == null && walkingParticlePrefab != null)
```
ManuelRauber, line 18: You simply can use `if (!particleObject)` when checking against Unity objects.

UOP1_Project/Assets/Scripts/Characters/WalkingParticle.cs
```diff
@@ -0,0 +1,37 @@
...
+        {
+            particleObject = Instantiate(walkingParticlePrefab, transform);
+            particleObject.Stop();
+        }
+        if(currentPosition != null && lastPosition != null && particleObject != null)
```
ManuelRauber, line 23: The first two equations will always be true, since a struct is never not null.

UOP1_Project/Assets/Scripts/Characters/WalkingParticle.cs
```diff
@@ -0,0 +1,37 @@
...
+        {
+            currentPosition = transform.position;
+            if(currentPosition.x * 1000 != lastPosition.x * 1000 || currentPosition.z * 1000 != lastPosition.z * 1000)
```
ManuelRauber, line 26: better define a treshold to check against instead of the multiplication, e.g.

`if (Mathf.Abs(currentPosition.x) < 0.001)` etc. Define treshold as a const somewhere in the project.

UOP1_Project/Assets/Scripts/Characters/WalkingParticle.cs
richard-fine: More importantly - what is the check actually for?

There's a bunch of code here which looks like it is trying to do "set up on the first run" things - assigning the positions, instantiating the prefab, etc. It would be much better, and much simpler, to do that in an OnEnable() method, instead of doing it in Update().

UOP1_Project/Assets/Scripts/Characters/WalkingParticle.cs
richard-fine: Even simpler: use [Vector2.Distance()](https://docs.unity3d.com/ScriptReference/Vector2.Distance.html) to calculate the distance between lastPosition and currentPosition. Then you don't need to do separate checks for x/z.

If the X/Z position is all that ever matters here - never the Y - then currentPosition/lastPosition could be changed to be Vector2 fields as well, and then you'd assign them like `currentPosition = new Vector2(transform.position.x, transform.position.z)`.

UOP1_Project/Assets/Scripts/Characters/WalkingParticle.cs
```diff
@@ -0,0 +1,41 @@
...
+
+    private void OnEnable()
+    {
+        lastPosition = new Vector2(transform.position.x, transform.position.z);
+        currentPosition = new Vector2(transform.position.x, transform.position.z);
+    }
+
+    private void Update()
+    {
+        if (!particleObject && walkingParticlePrefab)
```
richard-fine, line 22: Why not do this in OnEnable too? You could also destroy the instantiated system in OnDisable, so there's a straightforward symmetry.

richard-fine: (If you do it in OnEnable then I think you will be able to just assume that it is not null in Update, so you won't need the if-check for it at all any more).

UOP1_Project/Assets/Scripts/Characters/WalkingParticle.cs
```diff
@@ -0,0 +1,41 @@
...
+        {
+            particleObject = Instantiate(walkingParticlePrefab, transform);
+            particleObject.Stop();
+        }
+        if(particleObject)
+        {
+            currentPosition = new Vector2(transform.position.x, transform.position.z);
+            if(Vector2.Distance(currentPosition, lastPosition) > 0)
```
richard-fine, line 30: I suggest making that 0 a value you can configure in the Inspector. It might be that you want it to be a small but non-zero value, instead of zero, but it's something that an artist should be able to tune.

ciro-unity:
Small issue: the particles are emitted even when you jump. The script has to speak with the CharacterController to know if it's in the air. How would you solve it?

UOP1_Project/Assets/Scripts/Characters/WalkingParticle.cs
```diff
@@ -0,0 +1,42 @@
...
+
+    private Vector2 lastPosition;
+    private Vector2 currentPosition;
+    private ParticleSystem particleObject;
+
+    private void OnEnable()
+    {
+        lastPosition = new Vector2(transform.position.x, transform.position.z);
+        currentPosition = new Vector2(transform.position.x, transform.position.z);
+        if (!particleObject && walkingParticlePrefab)
```
richard-fine, line 19: Alright, getting better :) What I would suggest now, is: there's no point having this behaviour be enabled if there's no particle prefab set up, right? It won't be able to do anything, it will just sit there wasting CPU cycles.

So, instead of saying "if there is a walkingParticlePrefab, then instantiate it" - the code will all be simpler if you say "if there is NOT a walkingParticlePrefab, then log an error and set `enabled = false;`. Setting enabled to false means that Update() will not run.

Once you have that in place, you can simplify Update() and remove the `if(particleObject)` check - you will be able to assume that there is always a particleObject, because you will have made it impossible to be running Update() without OnEnable() setting it up for you.

ciro-unity:
Just to quickly comment on this PR, I didn't pull it in because I didn't want the default Unity particle system show up in every livestream we make, since we don't have particles for the walk yet :)
(in fact I need to change the label, as it's not art-related)

Will take a look at it later!

ciro-unity:
Sorry @printto, I'm closing this one because we got some shiny particles implemented, in PR #192. Being particles... the eye wants its part!
Also they still need to be enabled by the State Machine, which is only now starting to come together.
Thanks anyway though, you've reserved yourself a little mention in the credits :)

PR #39, opened 2020-10-02
Added map boundaries for CharController scene.

GintasS:
This is a small PR where I added 4 boundaries with box colliders so a pig would not fall out from the map when trying to go over the edge of the map.


ciro-unity:
It's just a test scene, refinements are not needed at the moment. Thanks.

PR #43, opened 2020-10-02
New Prototyping Environment

NPBirader:
Created a new prototyping Environment with Pro Builder which can be expanded and added more.


UOP1_Project/ProjectSettings/ProjectVersion.txt
```diff
@@ -1,2 +1,2 @@
-m_EditorVersion: 2019.4.11f1
-m_EditorVersionWithRevision: 2019.4.11f1 (2d9804dddde7)
+m_EditorVersion: 2019.4.8f1
```
HyagoOliveira, line 1: We cannot change the project version.
You should not submitted this file. Please remove from your PR.

NPBirader: My mistake. How can i remove it? I dont know

ciro-unity:
In any case, this is not needed on master at the moment.

PR #46, opened 2020-10-03
added small cave for prototype environment 

notwill111:
I've added a small cave in the corner on the map to simulate what going inside a cave would be like.


ciro-unity:
You actually didn't push the scene file.
But in general, it's just a test scene, refinements are not needed at the moment. Thanks.

notwill111:
> You actually didn't push the scene file.
> But in general, it's just a test scene, refinements are not needed at the moment. Thanks.

Oh ok, thanks, yeah sorry my bad

Issue #48, opened 2020-10-03
Player gets stuck in mid-air when colliding with an obstacle


MEGATANK58:
**Short description**
Players can do some movement tricks to align themselves so their hands get stuck in Obstacle,not sure if that's because of the way obstacles are made in the project but its not intended to happen

**Expected behaviour**
Player should not be able to float in air and instead fall to ground

**To Reproduce**
1. Click Play
2. Go to the obstacles in top left side of map
3. Align your hands with the wall (or side) as much as you can
4. Click Jump and you can float (until you click jump again or change your position)

**Notes**
I have tried this in master branch and since snappy movement is patched this can be quite hard to replicate but is still possible (according to my findings)
Here are some in-game screenshots:
![Bug1](https://user-images.githubusercontent.com/51410502/94984304-49555980-0568-11eb-8d49-fc8a79fa5035.png)
![Bug2](https://user-images.githubusercontent.com/51410502/94984306-4c504a00-0568-11eb-95e0-79580f5bbcc8.png)
![Bug3](https://user-images.githubusercontent.com/51410502/94984307-4d817700-0568-11eb-97dd-5249216eaa25.png)

MPjeremy:
Hi, could you provide more details? I can't seem to reproduce the problem

Hemanth759:
Hey I was able to reproduce.
![image](https://user-images.githubusercontent.com/33123731/94985614-c9ce8700-0575-11eb-8dfb-e3ded8cbc7f0.png)

I think the problem is with the character collider.
![image](https://user-images.githubusercontent.com/33123731/94985625-dd79ed80-0575-11eb-9286-248e9e0c1d42.png)

Hemanth759:
So I moved the box collider of the building with some offset in x-axis and found that the player collider getting struck with one side of the box collider. (only one side though remaining side of the box collider works fine).
![image](https://user-images.githubusercontent.com/33123731/94985706-7577d700-0576-11eb-9d65-c04ba4d76a34.png)

Hemanth759:
I created simple script which detects the collision on character and draws a line in the editor with one unit long. 
Interestingly I found that the character is colliding with the boxcollider in sideways and that is preventing from falling down.
![image](https://user-images.githubusercontent.com/33123731/94985836-6e9d9400-0577-11eb-9691-1eaf8426cb4b.png)

[...]

PR #26, opened 2020-10-01
Fixed steep surfaces bug and implemented sliding

mwert09:
Fixed player climbing steep surfaces bug and implemented sliding.

Issue: https://github.com/UnityTechnologies/open-project-1/issues/8
Card: https://open.codecks.io/unity-open-project-1/decks/13/card/16f-player-can-climb-steep-surfaces

PR #49, opened 2020-10-03
Fixed the ability to climb steep surfaces & getting stuck mid-air

dyoxyne:
# Description

By introducing `isUnderneathStable` property, we will know if the character is under a stable or instable surface.
A surface is considered stable if the slope angle is lower or equal to the character slope angle limit.

I used a downward `SphereCastAll` that will calculate the total unstable slopes direction and slide the character accordingly. The steeper it is, the faster the character will fall.
It will correctly handle multiple unstable slopes (shown in the gif). It won't be interfering with pushing an object or collision with the head since it is not using `OnControllerColliderHit`.

PS: I saw there was already other PRs that were also trying to fix it but I wanted to share also my solution. Don't hesitate to close it if you find it redundant/irrelevant.

Fixes Issues #8 & #48

### DIFF
--- a/UOP1_Project/Assets/Scripts/Characters/Character.cs
+++ b/UOP1_Project/Assets/Scripts/Characters/Character.cs
@@ -88,23 +88,8 @@ public class Character : MonoBehaviour
         movementVector.y = verticalMovement;
         characterController.Move(movementVector * Time.deltaTime);
 
-        if (characterController.isGrounded) {
-            //Check if underneath the character is stable
-            isUnderneathStable = false; //By default, underneath is not stable
-            RaycastHit hit;
-            Ray ray = new Ray(transform.position + characterController.center, Vector3.down);
-            if (Physics.Raycast(ray, out hit, Mathf.Infinity)) {
-                float underneathSlopeAngle = Vector3.Angle(transform.up, hit.normal);
-                if (underneathSlopeAngle <= characterController.slopeLimit) {
-                    //We found a stable underneath
-                    isUnderneathStable = true;
-                } else {
-                    //Add a slide movement along the underneath slope direction
-                    Vector3 underneathSlopeDirection = Vector3.Cross(hit.normal, Vector3.Cross(hit.normal, Vector3.up));
-                    characterController.Move(underneathSlopeDirection * gravityMultiplier * Time.deltaTime);
-                }
-            }
-        }
+        //Check if underneath is stable and slide accordingly
+        CheckUnderneathStabilityAndSlide();
 
         //Rotate to the movement direction
         movementVector.y = 0f;
@@ -139,6 +124,28 @@ public class Character : MonoBehaviour
                 gravityContributionMultiplier = 1f;
 
                 verticalMovement = 0f;
+            }
+        }
+    }
+
+    private void CheckUnderneathStabilityAndSlide()
+    {
+        isUnderneathStable = false; //By default, underneath is not stable
+
+        if (characterController.isGrounded) {
+            //Perform a downward raycast to check underneath the character
+            RaycastHit hit;
+            Ray ray = new Ray(transform.position + characterController.center, Vector3.down);
+            if (Physics.Raycast(ray, out hit, Mathf.Infinity)) {
+                float underneathSlopeAngle = Vector3.Angle(transform.up, hit.normal);
+                if (underneathSlopeAngle <= characterController.slopeLimit) {
+                    //We found a stable underneath
+                    isUnderneathStable = true;
+                } else {
+                    //Add a slide movement along the underneath slope direction
+                    Vector3 underneathSlopeDirection = Vector3.Cross(hit.normal, Vector3.Cross(hit.normal, Vector3.up)).normalized;
+                    characterController.Move(underneathSlopeDirection * gravityMultiplier * Time.deltaTime);
+                }
             }
         }
     }
--- a/UOP1_Project/Assets/Scripts/Characters/Character.cs
+++ b/UOP1_Project/Assets/Scripts/Characters/Character.cs
@@ -23,6 +23,7 @@ public class Character : MonoBehaviour
     private float verticalMovement = 0f; //Represents how much a player will move vertically in a frame. Affected by gravity * gravityContributionMultiplier
     private Vector3 inputVector; //Initial input horizontal movement (y == 0f)
     private Vector3 movementVector; //Final movement vector
+    private bool isUnderneathStable = false; //If true, it is stable enough to be able to jump and to not slide if there is no input
 
     private void Awake()
     {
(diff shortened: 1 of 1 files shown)


UOP1_Project/Assets/Scripts/Characters/Character.cs
```diff
@@ -87,6 +88,24 @@ private void Update()
         movementVector.y = verticalMovement;
         characterController.Move(movementVector * Time.deltaTime);
 
+        if (characterController.isGrounded) {
```
HyagoOliveira, line 91: This method is getting pretty big.
Suggestion: put your code inside a function with a good name. This increase legibility among other developers.

UOP1_Project/Assets/Scripts/Characters/Character.cs
```diff
@@ -87,6 +88,24 @@ private void Update()
...
+            RaycastHit hit;
+            Ray ray = new Ray(transform.position + characterController.center, Vector3.down);
+            if (Physics.Raycast(ray, out hit, Mathf.Infinity)) {
+                float underneathSlopeAngle = Vector3.Angle(transform.up, hit.normal);
+                if (underneathSlopeAngle <= characterController.slopeLimit) {
+                    //We found a stable underneath
+                    isUnderneathStable = true;
+                } else {
+                    //Add a slide movement along the underneath slope direction
+                    Vector3 underneathSlopeDirection = Vector3.Cross(hit.normal, Vector3.Cross(hit.normal, Vector3.up));
```
HyagoOliveira, line 103: Why two `Cross` operations here? Can't be done with just one?

UOP1_Project/Assets/Scripts/Characters/Character.cs
```diff
@@ -87,6 +88,24 @@ private void Update()
...
+                    characterController.Move(underneathSlopeDirection * gravityMultiplier * Time.deltaTime);
```
HyagoOliveira, line 104: Would be better to have just one call to `characterController.Move()`.
You could add `underneathSlopeDirection * gravityMultiplier` on `movementVector` before line 89.

UOP1_Project/Assets/Scripts/Characters/Character.cs
```diff
@@ -87,6 +88,24 @@ private void Update()
         movementVector.y = verticalMovement;
         characterController.Move(movementVector * Time.deltaTime);
 
+        if (characterController.isGrounded) {
+            //Check if underneath the character is stable
+            isUnderneathStable = false; //By default, underneath is not stable
+            RaycastHit hit;
+            Ray ray = new Ray(transform.position + characterController.center, Vector3.down);
```
Hemanth759, line 95: Instead of calculating the hit position with raycast you can get the same in 
`private void OnControllerColliderHit(ControllerColliderHit hit)`

hit varaible contains the `hit.normal` using which you can calculate the contact angle which would be more efficient

dyoxyne: I am not using OnControllerColliderHit because this will be triggered each time a collider is hit.
That could be an object we want to push, a wall in front of us, an obstacle that hits the head, etc...

UOP1_Project/Assets/Scripts/Characters/Character.cs
dyoxyne: I decided to not add in the `movementVector` because it will affect the rotation of the character when the character slides down. I felt it was more natural (looking) to slide down by keeping the same rotation.
At the end, this should depend on the game design.

UOP1_Project/Assets/Scripts/Characters/Character.cs
dyoxyne: Doing one cross product in the current formula will only give "side" direction of the slope, but we want the "down" direction.
Maybe it could be achieved with only one cross product but I can't think about one. Do you have the formula in mind?

UOP1_Project/Assets/Scripts/Characters/Character.cs
dyoxyne: Sure. I will put it in another function.

UOP1_Project/Assets/Scripts/Characters/Character.cs
Hemanth759: ohhh that makes sense.
But what about the situations like below case
image

UOP1_Project/Assets/Scripts/Characters/Character.cs
```diff
@@ -124,6 +128,28 @@ private void OnControllerColliderHit(ControllerColliderHit hit)
         }
     }
 
+    private void CheckUnderneathStabilityAndSlide()
```
Hemanth759, line 131: I think your method will fail when the character stands at the end of an edge like this
image

dyoxyne: Does the character controller isGrounded returns true in your screenshot?
For me, the case you described, the character should be falling from the normal gravity and not from the slide.

Hemanth759: yes it does return true.
Yes the character should be falling but the capsule collider of the player is touching the wall at the edge and the charactercontroller thinks that player is on the ground as a result. 
I think the code you have written checks the angle with the actual ground rather then the wall here as you are casting a ray directly to the downwards from the center of the player.

dyoxyne: Thanks for the feedback. I was not aware of the exact behavior of isGrounded.
I changed my raycast into a SphereCastAll. It should catch the situation you described. I also saw #58 and tested it. It should fix it. Perhaps you could test it and confirm it?

Hemanth759: Yes this did fix the problem.😊

Hemanth759 (approved):
This pull request fixed the issues #8 and #48 by using the sphere cast at the bottom of player and finding the slope angle and also implemented sliding feature when the angle is steep.

UOP1_Project/Assets/Scripts/Characters/Character.cs
dyoxyne: Thanks for confirming it and the feedback!

ciro-unity:
Argh, I reviewed fixes to issue #8 yesterday and I didn't see this PR because the issue wasn't linked! (there were another 5 fixes!)
Your solution does seem solid, @dyoxyne.
If anyone has an opinion on the fix I merged vs. this one, let me know.

MEGATANK58:
I think this is a pretty good solution but your branch current has conflicts I'll suggest you to Push origin again from the GitHub Dekstop once then make a commit to this PR

Hemanth759:
> I think this is a pretty good solution but your branch current has conflicts I'll suggest you to Push origin again from the GitHub Dekstop once then make a commit to this PR

U mean pull origin right? @MEGATANK58

MEGATANK58:
> > I think this is a pretty good solution but your branch current has conflicts I'll suggest you to Push origin again from the GitHub Dekstop once then make a commit to this PR
> 
> U mean pull origin right? @MEGATANK58

Oh yes lol

dyoxyne:
Thanks for the feedback. I will solve the conflicts later and push again on my branch.

ciro-unity:
> I think this is a pretty good solution

Is there a reason why this is a better solution that the one already in?

Hemanth759:
> > I think this is a pretty good solution
> 
> Is there a reason why this is a better solution that the one already in?

I think this is better then the #26 solution as that uses OnControllerColliderHit to calculate the hit position where as this uses a sphere cast at the bottom of the player.

As @dyoxyne told in his comment https://github.com/UnityTechnologies/open-project-1/pull/49#discussion_r499194166 OnControllerColliderHit approach detects hit position of hands, head which aren't useful and moving according to that points isn't correct.

ciro-unity:
Sorry, this PR slipped my mind. Since we have implemented a state machine for the character, the code in this PR would have to go in a completely different location now. If you can refactor this PR to work with that, I'd be happy to implement the fix... as I can see where you're coming from (with the multiple slopes).
If it's too complex don't worry: the game is not a platformer so we might never have to handle that situation!

ciro-unity:
Closing because the character controller code changed too much. Thanks anyway!

PR #51, opened 2020-10-03
Scriptable State Machine first iteration

neon-age:
Forum thread: [Character State Machine](https://forum.unity.com/threads/character-state-machine.979371/page-2)

Based on Pluggable AI. Here's extended concept: 
We use ScriptableObjects to make reusable logic and share data across all instances by reference.
It allows designers at create new logic just by changing states in the inspector.
And for programmers it means no boilerplate code, as it's all modular.
It might be used not only for NPCs, but for any events or logic.

image
image

At runtime we create per-instance data and attach it to machine:
image
Then states are working upon that data:
image
image

What classes are we dealing with?
- Machine: behaviour that handles data relation and states execution. Can be derived from.
- IStateData: any non-shared data that needs to be attached to machine instance.
- Action: event or logic.
- Decision: condition for actions trigger.
- StateNode: state logic that machine can execute.

What we already have?
- StateMachine base.
- Basic triggers.
- Compact and reorderable state node GUI.
- Super simple chasing AI example.
- Simple sound and particle event example.

What we need to add?
- Timers. We need some nice way to "yield" our logic, and somehow implement it nicely in GUI.
- Multiple states execution at once.
- Runtime data displaying in inspector.
- Transition priority from top to bottom.
- Custom GUI field for each ScriptableState type inside transition. For now it's just "on true" and object field.

What we may want to add?
- Extended triggers.
- Decision operator enum. (Or, When, etc..)
- StateFlow SO that stores all relations and has a nice GraphView editor.

Possible problems?
- As data is stored inside dictionaries, it might not perform well on huge number of instances (1000 AI's?), as we need to unbox data to generic on every update. Need to make some tests. If it would be bottleneck, I might try to workaround it with fancy reflections (is there any other way?) or just introduce logic execution rate based on distance.

There is many things to discuss, so please play around with it and ask any questions!


Hansen-L:
Very cool work, I'll probably play with this in my own projects too! 

As for this PR, I wonder if it might be better to exclude the states, and to keep this simpler by only having the FSM framework. By making some assumptions on the states that the project will need, you might be:
1. Making this PR heavier than it needs to be
2. Prematurely setting the direction on game design elements that have not been discussed yet (ChaseAction for example)

Nice job!

neon-age:
Thanks, I want to discuss these design elements as well.
They are directly related to FSM and SO workflows, so I think it would be easier to work on them in a single separate branch

UOP1_Project/Assets/Scripts/StateMachine/Base/StateNode.cs
```diff
@@ -0,0 +1,120 @@
...
+        {
+            // Prioritised transition index
+            // [0] transitions have highest priority
+            // Example:
+            // We see player and start chase, but if we hungry and see food - we chase food instead
+            // If watching for player is higher on the list, we'll not care about food while chasing player
+            // TODO: Make priority matter
+            var priority = int.MaxValue;
+            
+            for (var i = 0; i < transitions.Length; i++)
```
Proryanator, line 51: I might suggest splitting chunks of this method into smaller private methods for readability, it's a bit large. It'd be really easy at that point to see that this chunk achieves 'X' while the other chunk achieves 'Y' without spending too much time scanning over the whole method 😄

UOP1_Project/Assets/Scripts/StateMachine/Base/StateMachine.cs
```diff
@@ -0,0 +1,208 @@
...
+            foreach (var action in state.actions)
+            {
+                if (action.trigger != trigger || !action.logic)
+                    continue;
+                action.logic.BeginUpdate(this);
+            }
+        }
+
+        
+        // TODO: Getting data introduces unboxing to generic on every update. How can we avoid this?
```
Proryanator, line 84: I noticed this in the HungerDecision.OnDecide() implementation, it does makes the 'GetData' call every frame.

Yeah this one's a hard one to figure out. If it's possible to almost add in a 'Init()' method that you'd call once before using the state, that'd be ideal. But since we're using SO's there's no 'Awake()' or 'Start()' we can take advantage of. Quite a pickle!

neon-age: For components, it's easy to add some kind of ComponentPerInstance<T> inside state, which would be generic dictionary that gets data only once. 
But it makes code look harder and more boilerplate:
image
It needs to be handles internally, I'll see if I can came up with something better

neon-age:
Refactoring to fix generic type casting

ErikBehar:
Was trying to make the pizza critters, chase player if they are not hungry anymore, but could not achieve it, would need the true / false for each decision... otherwise would need to create a "isNotHungry" decision ...

neon-age:
Eating logic is not implemented in example :)
There's OnTrue and OnFalse enum on every "Decision Action", is it not enough for this?
Another thing I need to implement is "Any State" node, so we don't have to add same transitions in every state

ErikBehar:
> There's OnTrue and OnFalse enum on every "Decision Action", is it not enough for this?

true / false is available in the ui for actions, but not for decisions...
I wanted to have this logic:

in "Chase Food" state:
   if( SeesPlayer && !isHunger) 
       transition to "Chase Player" state

neon-age:
Yeah, the problem is, I don't know how to put it nicely in the UI.
There is yet no decision "tree" or something like that

ErikBehar:
> Yeah, the problem is, I don't know how to put it nicely in the UI.
> There is yet no decision "tree" or something like that.
> I'm working on StateFlowGraph right now, so then we'll see how it can be changed

I mean for a quick fix just put it the same as the action ...

neon-age:
Oh my lord, I actually forgot to implement "True/False" on actions, maybe that's why you wasn't able to achieve what you wanted :P

That's how it looks now with "decision conditions?":
image

MileyHollenberg:
FYI, the `#if UNITY_EDITOR` checks are not needed if the code itself sits in a subfolder called `Editor` at any point in the path. Every script in a `Editor` folder (or deeper folder within that) gets excluded from the build automatically so the `#if UNITY_EDITOR` checks aren't needed for the entire file there, saw this used in a couple editor specific pieces of code from this PR

ErikBehar:
So still trying to make the critter do more stuff. How do you change NPCData ? 

```
var state = GetData<NpcState>();
state.energy -= 1;
```

doesn't seem to change the value at all.

Also I needed to add : 
```
if (chase.target == null)
    return;
```
to the chase , otherwise it would crash occasionally.

Also ran into cases where data is saved from previous choices when adding and moving stuff around (expect that adding always adds a default thing (action/decision/etc. at the bottom of the list )

neon-age:
NpcState is a struct, so you need to call SetData after change, as ref version is missing right now.
Anyway, I'm going to make it Jobs compatible, so there would be huge changes in second iteration.

ErikBehar:
> NpcState is a struct, so you need to call SetData after change, as ref version is missing right now.
ok SetData worked, but not being able to visualize the NPCData instance is a bit of a downer 

> Anyway, I'm going to make it Jobs compatible, so there would be huge changes in second iteration.
Why ? Have fun

> Also about last one, it seems like a bug with data not being re-initialized after moving nodes in playmode, am I got it right?
When you add new things they are copies of previous.

neon-age:
Data visualization and logic debugging is another big thing that needs to be made.

> Why ? Have fun
In flavour of ideal architecture with parallelism that runs in DOTS environment. 
It's not actually a high-priority for this project, but we would need to do this at some point (hint: sooner is better).

> When you add new things they are copies of previous.
You mean just in the inspector? That's the default behaviour of editor insert :)

ciro-unity:
Thanks @neon-age, but in the end we went with another implementation. Yours was great, but in the end the coding felt a bit convoluted and we need to factor code readability given that many people will work on this project.
I hope we can use some of your editor coding skills on the other StateMachine though.

neon-age:
It's okay, I still need to make FSM for my own projects.
I'm looking into chosen implementation right now, and I'm not really happy about it.
I'll write on forums my criticism and how I would improve it. Also, yeah, it totally needs nice GraphView.
My nickname on discord is Neonage#1349.

PR #53, opened 2020-10-03
dialogue system first iteration completed

dkaloger:
all features apart from timeline support and multiple choice work but some ui reorganization and cleaning up the code may be needed note(i used the tree as an example of how to set up dialogue ,just walk near it and dialogue will start playing


qdkalger3:
also all changes were made in the cutscene scene so they need to be imported  later on

dkaloger:
i will close this pull request  and make a new one . it is in the wrong unity version and the changes are mostly scripts which can be imported very easily

Caiuse:
I missed your PR as it was closed, we've introduced a similar concept with the custom playable track. I can see you've got a track to flag that the dialog is playing, I've instead used custom clips to define segments of dialog, I feel we could combine both concepts and have clips that start a set of dialog.

PR #54, opened 2020-10-03
Audio Manager

fqureshi:
**Description**

- Added a new `Audio Manager`. It can be used to play background music.
- The game object does not get destroyed when loading a new scene. The song will continue to play across levels.
- There are `Play`, `PlayOneShot` and `Stop` functions.
- There is an ability to play additional sound effects on top of the background music that `AudioManager` is playing.
- Basic support for `AudioMixerGroup`
- I did not add any actual music, as I don't want to risk license violations. Maybe some talented artists in our community can create a cool background loop for us? :)

### DIFF
--- a/UOP1_Project/Assets/Scenes/CharController.unity
+++ b/UOP1_Project/Assets/Scenes/CharController.unity
@@ -213,7 +213,7 @@ Transform:
   m_LocalScale: {x: 5.7737727, y: 1.8967, z: 4.9019}
   m_Children: []
   m_Father: {fileID: 0}
-  m_RootOrder: 11
+  m_RootOrder: 12
   m_LocalEulerAnglesHint: {x: -3.5180001, y: 26.323002, z: -19.474}
 --- !u!1 &353533559
 GameObject:
@@ -306,7 +306,7 @@ Transform:
   m_LocalScale: {x: 4.8320527, y: 1, z: 4.9019}
   m_Children: []
   m_Father: {fileID: 0}
-  m_RootOrder: 9
+  m_RootOrder: 10
   m_LocalEulerAnglesHint: {x: 0, y: 0, z: 0}
 --- !u!1 &358843509
 GameObject:
@@ -399,7 +399,7 @@ Transform:
   m_LocalScale: {x: 7.8816566, y: 6.5043144, z: 2.0265193}
   m_Children: []
   m_Father: {fileID: 0}
-  m_RootOrder: 22
+  m_RootOrder: 23
   m_LocalEulerAnglesHint: {x: 0, y: 90.00001, z: 0}
 --- !u!1001 &393067907
 PrefabInstance:
@@ -431,7 +431,7 @@ PrefabInstance:
     - target: {fileID: 7160299325435046207, guid: fe3004506b4c6cd478eb2cca639b3713,
         type: 3}
       propertyPath: m_RootOrder
-      value: 19
+      value: 20
       objectReference: {fileID: 0}
     - target: {fileID: 7160299325435046207, guid: fe3004506b4c6cd478eb2cca639b3713,
         type: 3}
@@ -567,7 +567,7 @@ Transform:
   m_LocalScale: {x: 1, y: 1, z: 1}
   m_Children: []
   m_Father: {fileID: 0}
-  m_RootOrder: 6
+  m_RootOrder: 7
   m_LocalEulerAnglesHint: {x: 0, y: 0, z: 0}
 --- !u!114 &526256410 stripped
 MonoBehaviour:
@@ -685,7 +685,7 @@ Transform:
   m_LocalScale: {x: 1, y: 1, z: 1}
   m_Children: []
   m_Father: {fileID: 0}
-  m_RootOrder: 7
+  m_RootOrder: 8
   m_LocalEulerAnglesHint: {x: 0, y: 0, z: 0}
 --- !u!1001 &802198071
 PrefabInstance:
@@ -747,7 +747,7 @@ PrefabInstance:
     - target: {fileID: 384383464925863578, guid: 1ebb75c12ef91014e94524019e01192d,
         type: 3}
       propertyPath: m_RootOrder
-      value: 14
+      value: 15
       objectReference: {fileID: 0}
     - target: {fileID: 384383464925863578, guid: 1ebb75c12ef91014e94524019e01192d,
         type: 3}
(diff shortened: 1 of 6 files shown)


UOP1_Project/Assets/Scenes/CharController.unity
```diff
@@ -188,7 +188,7 @@ Transform:
   m_Children:
   - {fileID: 2020721999}
   m_Father: {fileID: 0}
-  m_RootOrder: 5
+  m_RootOrder: 6
```
fqureshi, line 191: I adjusted the ordering so that `AudioManager `is grouped with the other managers

image

UOP1_Project/Assets/Scripts/Audio/AudioManager.cs
```diff
@@ -0,0 +1,37 @@
...
+    {
+        audioSource.Stop();
+    }
+
+    private void PreventMultipleAudioManagers()
+    {
+        UnityEngine.Object[] otherAudioManagers = FindObjectsOfType(GetType());
+        if (otherAudioManagers.Length > 1)
+        {
+            gameObject.SetActive(false);
```
m4rtint, line 29: why do you need to setActive as false if you are going to destroy it anyway?

UOP1_Project/Assets/Scripts/Audio/AudioManager.cs
```diff
@@ -0,0 +1,37 @@
...
+    }
+
+    public void Stop()
+    {
+        audioSource.Stop();
+    }
+
+    private void PreventMultipleAudioManagers()
+    {
+        UnityEngine.Object[] otherAudioManagers = FindObjectsOfType(GetType());
```
m4rtint, line 26: Shouldn't this be AudioManager[], since casting it as Object would be boxing, and that's bad for performance. 

And it's better for readability to do Audiomanager[].

And I think you're basically trying to create a singleton? you can just do the normal singleton initialization if you want, (have a static instance., check if instance is null)...ect

UOP1_Project/Assets/Scripts/Audio/AudioManager.cs
```diff
@@ -0,0 +1,37 @@
+﻿using UnityEngine;
+
+public class AudioManager : MonoBehaviour
```
m4rtint, line 3: Maybe do a [RequireComponent(AudioSource)] 

Or check if audioSource != null, and then AddComponent<AudioSource>. 

Since it's not guaranteed if someone else uses this script, they'll add an AudioSource.

UOP1_Project/Assets/Scripts/Audio/AudioManager.cs
fqureshi: Good question :) `Destroy` does not actually destroy the object immediately. Therefore, inactivating the object first is a bit safer.

STARasGAMES:
This is singleton but doesn't provide static access. 
It has poor functionality and supports playing only one sound at a time.
Also, it doesn't use AudioMixer which is mandatory for such games.

fqureshi:
Thanks for your feedback. I modified the `AudioManager` to give static access. I also added in the ability to play additional sound effects on top any background music that the `AudioManager` is playing.

I'm a bit new with Unity, so I'll have to do some additional research into `AudioMixer`. I'll try and add that in a future PR when I've done my due diligence and learned how to use it properly.

This is something we can iterate on 🙂,

davejrodriguez (changes requested):
I think this singleton could use some work.

In general, I do not think a singleton audio manager is the best approach to a flexible audio system. Maybe a factory that creates audio sources on demand?

UOP1_Project/Assets/Scripts/Audio/AudioManager.cs
```diff
@@ -0,0 +1,47 @@
...
+
+	public void Stop()
+	{
+		_audioSource.Stop();
+	}
+
+	private void PreventMultipleAudioManagers()
+	{
+		AudioManager[] otherAudioManagers = FindObjectsOfType<AudioManager>();
+		if (otherAudioManagers.Length > 1)
```
TheJames12112, line 36: Hey, another alternative is to check if the '_instance' is null. If the instance is null we know this is the first AudioManager, if it is not null we know there must alareay be an AudioManager in the scene. FindObjectsOfType can be quite an expensive operation where as null check is much quicker. 
```
if(_instance != null)
{
    gameObject.SetActive(false);
    Destroy(gameObject);
}
```

fqureshi:
Thanks for the suggestions @davejrodriguez! I agree with them.

They should be implemented now.

davejrodriguez:
> In regards to a "factory creating Audio Sources on demand" - that is interesting! Will that be overkill though? Have you had success with this approach in the past?

Well, overkill is a question for the audio designers. I do envision we'll need to be able to play more than one sound at once.

The idea isn't too difficult. You have a script much like yours, but instead Play(AudioClip clip) playing the clip right on the adjacent AudioSource component, you instantiate (a more performant option would be to enable from an object pool) an AudioSource prefab and play it on that. Allows for as many sounds as needed. And the methods can be extended to take other parameters such as position, various audio source settings, mixer group etc.

UOP1_Project/Assets/Scripts/Audio/AudioManager.cs
ciro-unity: I think SetActive would act from the next frame anyway, so it's not "safer".
If you're not sure about wheter another object is going to execute code in another Update before this one gets destroyed, I think it's best to eliminate the ambiguity than try to patch it with things like this. Safer in the long term!

fqureshi: Fair enough - I removed the `SetActive`!

ciro-unity:

fqureshi:
> I'm doing a big refactoring and getting ready to pull code from all 3 PRs regarding the AudioSystem, so please don't push new code or close it :)

Gotcha - thanks for the heads up :)

ciro-unity:
Hey @fqureshi, I'm closing the PR since I've decided to merge PR #90 to main. For the rest, I see people left a lot of comments here so I guess there's stuff to absorb!

In any case, I liked the contribution, in fact you'll see you're listed as a contributor on the repo already.
And if you're on the Unity Discord, write here your username and I'll grant you the Open Projects Contributor role!

fqureshi:
Awesome @ciro-unity, thanks for going through all 3 PR's. I know it must not have been an easy task.

My username on discord is Fuzzy#9546 😄 thank you!

Issue #40, opened 2020-10-02
Avoid using magic numbers


MileyHollenberg:
**Short description**
There seem to be some classes which use magic numbers, magic numbers are defined by simply being put into the code as a direct value instead of through a parameter. When using magic numbers it can be difficult to know what they actually mean and if you use the same number multiple times it will be difficult to refactor this if you want to change the value later on.

**Expected behavior**
All numbers and other values should be stored in either private or public variables and no magic numbers should be allowed in the codebase. Logical numbers like -1 for error states, 0 and 1 for simple resets and 2 for dividing a value in half are numbers that I would not consider magic numbers, -5 and 0.02 (that I currently found in the code) I would consider magic numbers.

**Current findings**
https://github.com/UnityTechnologies/open-project-1/blob/master/UOP1_Project/Assets/Scripts/Characters/Character.cs#L81
https://github.com/UnityTechnologies/open-project-1/blob/master/UOP1_Project/Assets/Scripts/Characters/Character.cs#L92

**Notes**
This should become part of the style guide

andrewreal:
I'd like to tackle this please.

andrewreal:
For https://github.com/UnityTechnologies/open-project-1/blob/master/UOP1_Project/Assets/Scripts/Characters/Character.cs#L92 I'm not totally clear on what that value represents?

It seems to be a rotationSensitivty but can you suggest a better, more descriptive variable name for this?

MileyHollenberg:
> For https://github.com/UnityTechnologies/open-project-1/blob/master/UOP1_Project/Assets/Scripts/Characters/Character.cs#L92 I'm not totally clear on what that value represents?
> 
> It seems to be a rotationSensitivty but can you suggest a better, more descriptive variable name for this?

Since it's referring to a movement vector I'd say it's a movement threshold of some sorts. @Quickz any idea what this value represents?

TheBossBaby:
Can I have the chance to tackle this issue?
 From the comments: https://github.com/UnityTechnologies/open-project-1/blob/master/UOP1_Project/Assets/Scripts/Characters/Character.cs#L81

This value seems `verticalMovementReset` 

And for  https://github.com/UnityTechnologies/open-project-1/blob/master/UOP1_Project/Assets/Scripts/Characters/Character.cs#L92

I agree with @MileyHollenberg .

casefc3s:
[...]

PR #55, opened 2020-10-04
Remove magic numbers

andrewreal:
Replaced magic numbers with variables.

Issue: 40


ritik210200:
Blank lines

MileyHollenberg (approved):
LGTM, no need to test locally since the values remain the same but are now just stored in a property

UOP1_Project/Assets/Scripts/Characters/Character.cs
```diff
@@ -23,6 +24,7 @@ public class Character : MonoBehaviour
     private float verticalMovement = 0f; //Represents how much a player will move vertically in a frame. Affected by gravity * gravityContributionMultiplier
     private Vector3 inputVector; //Initial input horizontal movement (y == 0f)
     private Vector3 movementVector; //Final movement vector
+    private float rotationThreshold = .02f; // Used to prevent NaN result causing rotation in a non direction
```
m4rtint, line 27: if it's not being changed on inspector, you can probably set it as a 
`private const float RotationThreshold`

ciro-unity: Good point!

ciro-unity:
Implemented @m4rtint's suggestion

ciro-unity:
Close #40

PR #57, opened 2020-10-04
Designer friendly State Machine

OTGLLC:
This implementation is an expansion on the Pluggable AI.

Core Components are as follows:
CombatStatemachineController - Stores the Current Combat State and Handlers - Unique to each combatant
image

Combat State - Stores the current CombatAnimation to be processed, combat actions and Transitions - Unique to each combatant
image

Handlers - Main data repository of the state machine. Currently there is a MoveHandler and InputHandler which holds relevant data and references required to achieve its specific goal. For example the Movement Handler accepts the character controller and transform from the parent object.
image
image

Combat Animation - Stores the Animation Clip to be played. In the future this can facilitate custom per animation data such as hit box location, damage etc) - Unique to each combatant
image

Combat Action - Main workhorse of the State Machine system. Actions modify, mutate and communicate directly to handlers and between handlers. Gameplay is built from Combat Actions - Shared between combatants
image

Combat Transitions - Data access only and will never mutate data.  Each combat Transition holds a list of Combat Transitions which holds a list of Transition Decisions and the Next Combat State. The transition of states is based on a short circuit approach. If at any point a decision fails, the transition will escape and the state will start evaluating the next set of decisions
image
image
image

This pull request contains the required states and combat animations to mimic movement in already in the base project. All Data assets can be found in the Settings Folder 
image

Future work:

How to handle animations - I have migrated to driving animations via the Playable API which gives a lot of functionality like dynamic animation assignment, pausing and reversing. Utilizing a animator controller seems a bit overkill as we are driving our own states. Its up to the team as to which approach as the base system can accommodate both. 

Amount of Actions, States and Combat Animation Assets can balloon making it hard to get organized. - I have noticed this issue when making a complex State Machine
https://www.youtube.com/watch?v=XlXO-iN-IFA
Perhaps a custom node editor would help....


dkaloger:
are you aware of https://forum.unity.com/threads/combat.980931/

Proryanator:
fancy

Very clean! 👍

ciro-unity:
Thanks @OTGLLC, but in the end we went with another implementation. I felt like yours was very skewed towards combat, incorporating Animator by default, etc. so I went with a more generic one in the end.
Good work though!

PR #65, opened 2020-10-04
Changed Camera Move Event binding to Mouse delta X and scroll

FarhanGul:
The current implementation uses the unusual TFGH keys to orbit camera. With this fix you can orbit with mouse delta X and zoom with mouse scroll. This overrides the current implementation and works for keyboard mouse setup only. Further works will be required to support multiple devices such as game pads.

### DIFF
--- a/UOP1_Project/Assets/Scenes/CharController.unity
+++ b/UOP1_Project/Assets/Scenes/CharController.unity
@@ -286,7 +286,7 @@ Transform:
   m_PrefabInstance: {fileID: 0}
   m_PrefabAsset: {fileID: 0}
   m_GameObject: {fileID: 233317031}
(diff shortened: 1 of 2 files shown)


ciro-unity:
image

Can you use the InputSystem's Input Actions to do this? You should be able to just add Mouse (Delta) like in the image above, but you also should implement separate sensitivity for vertical and horizontal axes.
In the future we can wire them to some Settings Menu.

Thanks!

I'll close the PR since there are merging conflicts. If you pull a fresh copy from master it should be easier to merge once you're done.

FarhanGul:
Thanks @ciro-unity . I could not find a way to do this purely using Input actions since I wanted input combos. Mouse Right Button + Mouse Delta x. I'll have to look into it. It is strange you are getting a merge conflict, am I missing something? 
image

PR #70, opened 2020-10-05
Added fire effect (2D flip-book style)

jthom330:
* Fire prefab is in the prefabs folder
*  Fire animation uses a script to billboard the effect
* Contains particle systems for sparks and smoke.  

Item link: https://open.codecks.io/unity-open-project-1/decks/24/card/158-fire


treivize:
Hi @jthom330,

Too sad, this one is opened for more than a month and there is not any comment or updates here for a white and there is not any forum thread link too. 
There are lots of files committed here. So could you open a forum thread to explain your approach with screenshot of the shader graph and the end result from the showcase scene, please?

ciro-unity:
Sorry @jthom330, in the end we went with a different solution for the fire. So I'll go ahead and close this PR.

Thanks anyway! And if you're on the Unity Discord, let me know your nick and I'll give you the Open Projects Contributor role!

PR #71, opened 2020-10-05
Introduce StyleCop code quality analyzer.

StanislavChankov:
As we can see there are many developers wants to improve their skill by contribution to the project. We have threads in the forum, where many people are talking about code quality by accepting some conventions, but even if we do that it will be hand to manage it on the pull-request. In such cases static Roslyn analyzers comes in hands. `StyleCop` is one of the widely used analyzer, along with `Microsoft` and `SonarLint`. I suggest everybody to install the Microsoft Visual Studio extension `SonarLint`.
I see the code style differentiates very much in the different PRs, but we want to keep the same as one developer worked on it and no later refactoring of the code style will be required.
There are so many PRs from different developers wanted to involve their skills, I think this will be very **useful** as this was the one of the tools that helped me maintain high quality code.

- [x] * Configured StyleCop warnings/errors/suggestions;
- [x] * Fixed errors/warnings/suggestions from StyleCop;
- [x] * Supress errors and warnings into GlobalSuppressions.cs that are generated from auto-generated files.
- [x] * Added editorconfig configuring the global analyzers of SonarLint, StyleCop and others.

Other tools/stuff that **will** increase the code quality. Unit-tests, code-coverage, SonarLint, Dependency Injection

PR #73, opened 2020-10-05
Setup automatic linting as a Github Action

MileyHollenberg:
**What?**
Adding an automatic linter to the project through Github Actions

**Why?**
In order to enforce IDE agnostic style-guide enforcement it's better to run this when you push code to a branch rather than relying on IDE specific implementations. Note that having IDE level support would be valuable still but we need to find a method that works for all major IDEs and isn't tied to a specific one.

**How?**
Through adding a Github Action I implemented the [dotnet-format](https://github.com/dotnet/format) linter in order to automatically format the code. After changes have been made it will add and commit those changes to the current branch. All the user has to do is pull these latest changes in order to be up to date again. 

**Why not do this inside an IDE?**
Eventually we will want that as well, the main concern is that we need to find a way to support the broadest amount of IDEs as some may not have a specific linter whilst others do. We can't assume nor enforce that everyone will be using the same IDE so for now it's better to ensure linting is done remotely whilst we look at a way to implement this locally.

**Where is the style-guide?**
This is currently still being discussed internally but it should be ready soon, keep an eye out on the forum :)


MileyHollenberg:
Just noticed that the Github Actions bot does not have the ability to sign the CLA, I'll setup a separate account for this tonight and use that instead

StanislavChankovSellerCloud:
@MileyHollenberg good one, but it duplicates in some way my PR 
https://github.com/UnityTechnologies/open-project-1/pull/71

StanislavChankov:
@MileyHollenberg , this seems to be super cool idea. I am not sure whether it work, but we definetely wants this as well.
Q: How we will clear the commits from the Github Actions, since they will be meaningless for the code/logic?

MileyHollenberg:
The Github Actions bot will be replaced with a different account, they will indeed not have any meaningful effect on the code itself but from what I can see so far in the commit history is that all PRs have been squashed meaning the bot probably won't even show up after a PR has been merged.

Also I will be updating this PR tomorrow, upgrading my PC took a bit more time than expected

StanislavChankov:
@MileyHollenberg , you could read and share your thoughts about DI - [](https://forum.unity.com/threads/dependency-injection.983484/)

ciro-unity:
Great contribution, Miley!

PR #74, opened 2020-10-06
Use null propagation operator

gayanW:
Since C# 6.0, we could use null propagation operator for null checks.

### DIFF
--- a/UOP1_Project/Assets/Scripts/InputReader.cs
+++ b/UOP1_Project/Assets/Scripts/InputReader.cs
@@ -1,5 +1,4 @@
-﻿using System.Collections.Generic;
-using UnityEngine;
+﻿using UnityEngine;
 using UnityEngine.InputSystem;
 using UnityEngine.Events;
 
@@ -66,10 +65,7 @@ public class InputReader : ScriptableObject, GameInput.IGameplayActions
 
 	public void OnMove(InputAction.CallbackContext context)
 	{
-		if (moveEvent != null)
-		{
-			moveEvent.Invoke(context.ReadValue<Vector2>());
-		}
+		moveEvent?.Invoke(context.ReadValue<Vector2>());
 	}
 
 	public void OnPause(InputAction.CallbackContext context)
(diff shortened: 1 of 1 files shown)


StanislavChankov:
It is a good refactoring.
@ciro-unity , let's have this merged.

Twilice:
Even though UnityAction itself supports null propogation. I do not think it is a good code practise to use null propogation inside ScriptableObject or MonoBehaviour which do not support it.

gayanW:
UnityAction is a simple delegate. So is there's any specific reason not to use the null propagation operator?

[Thread-safe delegate invocation](https://docs.microsoft.com/en-us/dotnet/csharp/language-reference/operators/member-access-operators#thread-safe-delegate-invocation)

Twilice:
UnityAction is indeed a simple delegate, hence it supports it.

But ScriptableObject and MonoBehaviour does not support it. None of the UnityEngine.Object does. In generall I think it's a slippery slope to use it in a Unity project, inside a class which itself does not support it.

I think there is a risk that someone who does not know the difference, in Unity, using a normal if null check compared to null propogation will see this code and think it can safely be used anywhere.

Now I do not agree that Units decision to overload the null operator for Unity.Object was a good decision. I would have much rather Unity did something similar to UnityEngine.Object.ReferenceNull( gameobject ) and we would not have had this discussion.

ciro-unity:
Unfortunately for what I know, @Twilice is correct.

Since Unity overloads the `==` operator to run some extra checks, the `?.` operator cannot be used so safely on objects deriving from `UnityEngine.Object` (such as `MonoBehaviour`).

However, since it can be used here since it's a `UnityAction`, why not accept it? If somebody uses it somewhere else incorrectly and it's wrong, we can catch it (?). There's plenty of people reviewing code on the project.
Or is it a risk?

gayanW:
It's about knowing the supported types for a given operator though. Anyway, I'd be closing this PR. Thank you for the comments.

PR #76, opened 2020-10-07
A State Machine implementation using SO (v1)

kcastagnini:
Thread link: [https://forum.unity.com/threads/character-state-machine.979371/](https://forum.unity.com/threads/character-state-machine.979371/)

**Overview**
This system implements a FSM (finite state machine) using SOs (ScriptableObject) and it's inspired by [the Pluggable AI live training session](https://www.youtube.com/watch?v=cHUXh5biQMg&list=PLX2vGYjWbI0ROSj_B0_eir_VkHrEkd4pi).
This type of approach will give game designers the possibility to edit the behaviour of a FSM without writing code.
You can find the system code at _Assets/Scripts/StateMachine/_ and an implementation example at _Assets/CharacterControllerExample/_.

The system is composed by these parts:
- **Action**: a task to be performed (e.g.: make a character jump, fire a weapon, play an audio clip, etc.).
- **State**: a collection of actions.
- **Condition**: a condition that needs to be verified (e.g.: is the character grounded?, is a button pressed?, etc.).
- **Transition**: a change from one state to another that needs to happen when a certain condition is verified.
- **TransitionTable**: a collection of transitions.
- **StateMachine**: this will hold a reference to a transition table for lookups in order to correctly update its internal state. It will also mantain a reference to a user defined class that will be passed as a parameter to actions and decisions.

In order for actions to perform tasks and for conditions to evaluate a boolean expression, they need to work on some data.
A data model class is a user defined class (can be a plain C# class, a MonoBehaviour, a ScriptableObject etc) referenced by
a StateMachine that will be automatically passed as an argument to actions and conditions in order to function properly.

To better understand how everything works let's take a look at the character controller example that you can find at _Assets/CharacterControllerExample/Scripts/_.
In this example we will implement the FSM shown in this diagram:
FsmExample

The boxes names Standing, Jumping, etc are the states of our FSM.
The arrows represent transitions and the lines like "Press 'F'" are the conditions that need to be true in order to perform the transition.

**Workflow**
First we will create our data model, the **Character** class on which actions and decisions will operate. It's a dummy implementation just for example purpose, you don't need to understand it, just take a look at the methods to get a general idea on how it works.

Now let's implements actions and decisions, you can find them under _Assets/CharacterControllerExample/Scripts/CharacterController/_.
As you can see all the actions derives from **Action<_Character_>** and all the conditions derives from **Condition<_Character_>**.
This is the type of our data model that we defined earlier and it's also the type of the parameter of our Act and IsMet methods.
Inside _JumpAction_ for example we ask the character to jump by calling the character.Jump() method. Inside _IsGroundedCondition_ we return true if character.IsGrounded, false otherwise.

After we created all the necessary actions and conditions, let's create the **CharacterControllerState** class by extending **State<_Character_>**. Leave the class empty. In order to associate actions to states we need to use the Inspector. Select and instance of a CharacterControllerState to add/remove/edit actions.
You can find all the SO instances under _Assets/CharacterControllerExample/SO/_ (to create an instance of a SO, go with the mouse over the Project window, right click with the mouse and select Create/PluggableFSM/[...]).
_Note: currently the Asset Picker doesn't work because it can't search for generic classes, you will need to drag and drop your SO instances from the Project window to the inspector._

On a State you will find 3 lists of actions: OnEnter, OnExit and OnUpdate. OnEnter actions will be executed once every time the FSM transits INTO this state, OnExit when the FSM transits OUT of this state. OnUpdate will be executed at every Update if the FSM is in this state.

Now in order to define the transition logic of our FSM we need to declare a transition table by extending **TransitionTable<_Character_>**. This is a class that needs to be populated in the Inspector too.
transitionTable

In a transition table we first define the Initial State which is the state the FSM will start with.
Then inside the "State to state transitions" we can add and remove transitions. You can read an entry of this list as:
"_When you are in [fromState], go to [toState] if [condition] is True/False_".

The "Any state to state transitions" on the other hand tells the FSM to transit to a specific state if a certain condition is verified (and if the starting state is different from the destination state, this to avoid self-loops on the same state). 
You can read an entry as:
"_Go from any state to [toState] if [condition] is True/False_".
This is useful for use cases like the one in this example. As you can see, when we press the Fire button we want to go to the Firing state (if we are not already in the Firing state). It would be time consuming to add a transition for every possibile state we are in, but with this method it's just a single entry.

Finally we can create our CharacterController by extending **StateMachine<_Character_>** and referencing an instance of a **Character** and an instance of a **CharacterTransitionTable**. At every OnUpdate this class will update the current state and test the current conditions. It will transit to a new state accordingly to the referenced transition table when necessary.
You find this class at _Assets/CharacterControllerExample/Scripts/CharacterController/CharacterControllerExample_.

You can try the example at _Assets/CharacterControllerExample/Scenes/CharacterControllerExampleScene_.


neon-age:
These scripts are super boilerplate.
image
Why not make generic action that has exposed InputActionReference field?

kcastagnini:
> These scripts are super boilerplate.
> image
> Why not make generic action that has exposed InputActionReference field?

Yes my bad, we actually need only one State subclass, I really don't know why I created so many states subclasses, I will correct it.
We still need subclasses for conditions and actions though.

UOP1_Project/Assets/Scripts/StateMachine/Initializable.cs
```diff
@@ -0,0 +1,17 @@
+﻿using UnityEditor;
+using UnityEngine;
+
+namespace KarimCastagnini.PluggableFSM
+{
+	//A class to perform initialization only when in PlayMode or when in Built, need to find a more elegant way
+	public abstract class Initializable : ScriptableObject
```
SamMacedo, line 7: One alternative could be to use an interface and an extension method:

```csharp
public interface Initializable
{
    void Initialize();
}

public static class InitializableExtensions
{
    public static void HandleInitialization(this Initializable initializable)
    {
        if (EditorApplication.isPlayingOrWillChangePlaymode || Application.isPlaying)
            initializable.Initialize();
    }
}
```

And then on each class that needs initialization implement Initializable interface:

```csharp
public class StateBase : ScriptableObject, Initializable
{
    public void Initialize()
    {
        // Do stuff...
    }

    private void OnEnable()
    {
        HandleInitialization();
    }
}
```

DeivSky:
I was thinking that too, we should make it an interface `IInitializable`, it's a rather common one so we could all use it. Plus I think it's a little awkward to have a shared base class for `StateBase` and `TransitionTableBase`, they're called 'Base' for a reason. I get that the implementation is the same but that doesn't mean they should be coupled together.

And then just use directives. 
``` csharp
private void OnEnable()
{
#if UNITY_EDITOR
    if (UnityEditor.EditorApplication.isPlayingOrWillChangePlaymode)
#endif 
        Initialize();
}
```
This will remove unnecessary checks in build (Application.isPlaying is always true), and **calling UnityEditor scripts is not allowed in build**.
image

SamMacedo:
> This will remove unnecessary checks in build (Application.isPlaying is always true)
You are right, no need to check it when running from a build.

> What do you mean with exposed InputActionReference field?
If I'm not mistaken I think he means allowing the button to be selected from the inspector instead of having a class for each button press (fire, jump, etc.)

kcastagnini:
> > What do you mean with exposed InputActionReference field?
> 
> If I'm not mistaken I think he means allowing the button to be selected from the inspector instead of having a class for each button press (fire, jump, etc.)

Oh ok but this just an example to show how to create different conditions, in a real world scenario I would have handled things differently.

And yes you are right, it is better to have a IInitializable interface. As @DeivSky pointed out it's a little weird to have StateBase and TransitionTableBase share the same base class.

ErikBehar:
not sure how much you care about the example but :
1) holding cntrl + pressing spacebar repeatedly  makes the char wiggle between standup and crouch ... 
2) jump + fire in mid air allows you to jump again as many times as you like

kcastagnini:
> not sure how much you care about the example but :
> 
> 1. holding cntrl + pressing spacebar repeatedly  makes the char wiggle between standing and ducking : expect no wiggling
> 2. jump + fire in mid air allows you to jump again as many times as you like : expect only allow jump if is grounded regardless of fire
> 
> probably due to transition back to standing from fire ... which leads me to point out that we need multiple conditions per transition ... which even if we had, you would still have another issue with the jump on enter action, because it would jump again regardless, which leads me to say that you also need conditions on actions.
> 
> Oh look we are doing visual programming now = }

Hello @ErikBehar :) I wanted to create a really simple FSM for the example just to show the workflow.
The behaviour of this CharacterController is not what you would expect from a classic controller, but the FSM still executes correctly accordingly to the diagram.

To fix the issue number 2 for example it would be awesome to be able to compose conditions in the editor using atomic propositions and logical connectives, for example: 
"Go from **Standing** to **Jumping** if **JumpButtonDown _&&_ IsGrounded** is True".

But as you said we are doing visual programming now!
It's about finding the right balance I guess.

fmeneuhe:
The first links above seems to be broken.

kcastagnini:
> The first links above seems to be broken.

For some reason it only works when I copy and paste the link in a new tab!

GFX5000i:
### Some wild suggestions (that will probably make this a little more complex):
1) Instead of a current state, keep a list of active states. Allowing multiple, independent states to work simultaneously.
2) Remove the "from" in the transition list and make the conditions a list for each state. Allowing multiple conditions, checking active states being one of them.
3) Add to the conditions an option to evaluate only to enter, to exit or both. Allowing different conditions to be evaluated for entering and exiting a state.

#### Dealing with multiple conditions:
They should either be independent, with complex ones being evaluated internally by a single custom condition.
OR
The evaluation should be in a simple cascading style, with each one indicating how to deal with the next one on the list (AND, OR, XOR, etc). This could be simpler than a full implementation of visual programming.

ciro-unity:
Thanks @kcastagnini, in the end we went with another implementation. I feel like yours was good too, maybe it fell a bit short in terms of possibilities (for instance I don't think it implements 2 conditions at the same time?) although it's nice that it has the "from any state".

Also are you on Discord? If so, tell me your complete nickname and I can give you the role of "Open Project Contributor". Same for @SamMacedo. This is different from the credits in the game, where we'll be using your real name taken from Github or the username if not possible.

kcastagnini:
> Thanks @kcastagnini, in the end we went with another implementation. I feel like yours was good too, maybe it fell a bit short in terms of possibilities (for instance I don't think it implements 2 conditions at the same time?) although it's nice that it has the "from any state".
> 
> Also are you on Discord? If so, tell me your complete nickname and I can give you the role of "Open Project Contributor". Same for @SamMacedo. This is different from the credits in the game, where we'll be using your real name taken from Github or the username if not possible.
> No worries if you don't care about Discord.

Ciao @ciro-unity , thanks for the update. Are we still gonna be able to contribute to the picked system by implementing features and improvements?
My Discord nickname is kcastagnini#8778 
Thanks!

PR #79, opened 2020-10-08
Remove newlines

RWOverdijk:
This is a combination between a question and maybe a fix.

I tried using the shader and it failed to run because of too few arguments to a macro. The only thing I could think of was that it doesn't like newlines for the arguments... And so I removed them. Now the shader works for me.

I'm not part of the game, so forgive me for not taking the official route. I hope this is useful. If not, close it and optionally tell me why this fix is wrong. I won't feel hurt either way 😄


RWOverdijk:
I'm not signing a CLA. If this PR ends up being a fix it's all yours. If not 🤷‍♂️I tried.

PR #80, opened 2020-10-08
Removal of unnecessary test files

MEGATANK58:
Removal of unnecessary Tests folder, InputTestScene and the content in the Scene. I have tested the game and it doesn't breaks or effects the game in any possible way.

Forum thread: [Lets stick to contribution guidelines](https://forum.unity.com/threads/lets-stick-to-the-contribution-guidelines.984449/)
Removal of unnecessary test files,
As per the community guidelines we are not supposed to make such testing scenes.
This has been fixed by deleting this, i have tested this and it doesn't introduce any errors for any major system of game.


ciro-unity:
Good point, thanks!

PR #85, opened 2020-10-14
Audio System

RKS13D:

**[for Generic PRs]**  
https://forum.unity.com/threads/audio-system.987549/
Audio System 
Audio System is a system controlling SFX and Theme songs in the scene.
Audio System is  Single class for managing audio us just need to add it to an empty game object and it will control add the require components you can also access functions like play and pause audio.


UOP1_Project/Assets/Scripts/AudioSystem.cs
```diff
@@ -0,0 +1,175 @@
...
+
+	public float delayInCrossfading = 0.3f;
+	public List<Theme> tracks = new List<Theme>();
+	public List<SFX> sounds = new List<SFX>();
+
+	private static int BoolToBinary(bool b) => b ? 1 : 0;
+	private SFX GetSoundByName(string sName) => sounds.Find(x => x.name == sName);
+
+	public float MusicVolume => !PlayerPrefs.HasKey("MusicVolume") ? 1f : PlayerPrefs.GetFloat("MusicVolume");
+	public float SfxVolume => !PlayerPrefs.HasKey("SFXVolume") ? 1f : PlayerPrefs.GetFloat("SFXVolume");
```
mdzahedhossain, line 27: Better to move these strings into constants. There are few inside the `Awake` method too.

UOP1_Project/Assets/Scripts/AudioSystem.cs
```diff
@@ -0,0 +1,175 @@
...
+		}
+		_theme.volume = MusicVolume;
+	}
+
+	// Sfx Button On/Off
+	public void SfxToggle()
+	{
+		_sfxMute = !_sfxMute;
+		_sfx.mute = _sfxMute;
+		PlayerPrefs.SetInt("sfxMute", BoolToBinary(_sfxMute));
```
mdzahedhossain, line 135: Constant for this key.

UOP1_Project/Assets/Scripts/AudioSystem.cs
```diff
@@ -0,0 +1,175 @@
...
+		PlayerPrefs.Save();
+	}
+
+	// Music Button On/Off
+	public void MusicToggle()
+	{
+		_musicMute = !_musicMute;
+		_theme.mute = _musicMute;
+		PlayerPrefs.SetInt("musicMute", BoolToBinary(_musicMute));
```
mdzahedhossain, line 144: Same here.

mdzahedhossain (commented):
Nitpicking.

davejrodriguez (changes requested):
Brief calls for centralized access, but the only way to use this class by instance. A single instance can only play two AudioSources at once. Yet each instance also controls global volume and mute.

Beyond that, no spatialization support. No looping or repeat support. No support for per-AudioSource settings changes (ie volume, pitch, etc).

Crossfade looks cool, but do we need to do this through code or can Timeline carry that load?

961974 (approved):

ciro-unity:

UOP1_Project/Assets/Scripts/AudioSystem.cs
```diff
@@ -0,0 +1,175 @@
...
+			_theme.Play();
+
+		while (delayInCrossfading < 1f)
+		{
+			_theme.volume = delayInCrossfading * MusicVolume;
+			delayInCrossfading += Time.unscaledDeltaTime;
+			yield return 0;
+		}
+		_theme.volume = MusicVolume;
+	}
```
ciro-unity, line 128: I didn't test the code, but did I understand it correctly that the first music will fade out before the second starts to fade in? Cross-fading usually means that the two songs fade in and out at the same time.

No big deal though and we'll see which method we need, just a matter of terminology :)

UOP1_Project/Assets/Scripts/AudioSystem.cs
```diff
@@ -0,0 +1,175 @@
...
+		{
+			time += Time.unscaledDeltaTime;
+			yield return 0;
+			if (time >= mixBufferClearDelay)
+			{
+				mixBuffer.Clear();
+				time = 0;
+			}
+		}
+	}
```
ciro-unity, line 78: Do I understand it correctly that this is for limiting the ability to play the same SFX twice in under 0.05?
I'm not sure we really need it, since we won't have that fast pace of random SFX, but at the same time we might have the same SFX play very close twice, and we want to hear both. Case in point: 2 same SFX happen but one is close and the other is far. I am supposed to hear both, or maybe just one, but for sure we want to hear the one close. With this method, the close one might be cutoff just because farther away, a similar SFX is playing.

ciro-unity: Unless I misunderstood the logic, in which case, apologies!

UOP1_Project/Assets/Scripts/AudioSystem.cs
```diff
@@ -0,0 +1,175 @@
...
+		}
+		delayInCrossfading = 0;
+
+		if (!_theme.isPlaying)
+			_theme.Play();
+
+		while (delayInCrossfading < 1f)
+		{
+			_theme.volume = delayInCrossfading * MusicVolume;
+			delayInCrossfading += Time.unscaledDeltaTime;
```
ciro-unity, line 124: Why here use `Time.unscaledDeltaTime` here? If the game has a slow-motion effect, then all these calculations won't work.

ciro-unity:
Hey @RKS13D, I'm closing the PR but I have taken some of your code and merged it into another PR #90, which I just merged to main. For the rest, I've left a few comments above.

In any case, I liked the contribution, and if you're on the Unity Discord, write here your username and I'll grant you the Open Projects Contributor role!

RKS13D:
That's great I am sorry for the late reply my discord username  is RKS

PR #89, opened 2020-10-15
Added a singleton to assist with project settings.

tacosontitan:
Created a singleton to manage settings for the project in a thread-safe way utilizing caching, and ideally would support loading settings from a file and a database.

**[for Generic PRs]**  
No.
This pull request will add a basic settings management utility for managing settings in a thread-safe way.
The project does not yet contain anything that I can find for managing user settings.
From scratch as a class.

**NOTE**: There is a minor bug in the check-in where I marked the class as internal; this needs to be adjusted to public.

### DIFF
--- a/UOP1_Project/Assets/Scripts/SettingsLibrary.cs
+++ b/UOP1_Project/Assets/Scripts/SettingsLibrary.cs
@@ -1,0 +1,72 @@
+using System;
+using System.Collections.Generic;
+using System.Reflection;
+
+/// <summary>
+/// A simple settings library with a built-in cache for global use.
+/// </summary>
+internal sealed class SettingsLibrary
+{
+
+    #region Singleton Setup
+
+    private SettingsLibrary() { }
+    private static readonly object instanceLock = new object();
+    private static SettingsLibrary instance;
+
+    /// <summary>
+    /// The globally available instance of the settings library.
+    /// </summary>
+    /// <remarks>Utilizes coupled null reference checks surrounding a lock for thread safety.</remarks>
+    public static SettingsLibrary Instance
+    {
+        get
+        {
+            if (instance == null)
+                lock (instanceLock)
+                    if (instance == null)
+                        instance = new SettingsLibrary();
+
+            return instance;
+        }
+    }
+
+    #endregion
+
+    #region Fields
+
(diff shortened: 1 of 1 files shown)


UOP1_Project/Assets/Scripts/SettingsLibrary.cs
```diff
@@ -0,0 +1,72 @@
+using System;
+using System.Collections.Generic;
+using System.Reflection;
+
+/// <summary>
+/// A simple settings library with a built-in cache for global use.
+/// </summary>
+internal sealed class SettingsLibrary
```
tacosontitan, line 8: This needs to be a public sealed class instead of internal.

ciro-unity:
Hey Taco, thanks for the contribution, but in general, we're trying to avoid the use of the singleton pattern. If anything, we do it through ScriptableObjects.
On the other hand, I'm not sure we need a generic class for this kind of stuff since many times the way these components cache information is quite custom. There's work going on with the Game Settings, and I think we already have a solution for this, so I'll probably close the PR soon.

I'd like to recognise your interest in the project anyway though, so I'll note this down as one of our "Alternate PRs". Also, if you're on Discord, would you like the Open Projects Contributor role? I'd need the username if so.

tacosontitan:
@ciro-unity I'd love to join the Discord! My username is tacosontitan#3060. For my future reference, is there a particular reason we're trying to avoid using the singelton pattern? I'm not upset with the decision at all, just want to make sure I understand the reasoning behind the decision so that I can make informed decisions when submitting work to the project.

ManuelRauber:
Singletons are a good, but often a really bad design pattern, chosen for the pure laziness of the developer. Most of the time, there is no good reason to favor a Singleton over a different design pattern. They often lead to bad design choices, are hardly interchangeable and testable. They couple your code together like glue.

ciro-unity:
Manuel explained some of the points. It's not like we don't want any Singleton at all in the project, but at least try not to do with them. This is the issue with some other open PRs actually...

For Discord, are you on the Unity Discord at all? https://discord.gg/EZBXA4V

ciro-unity:
Can't find you on Discord, @tacosontitan. Let me know if you solve it and I'll assign the role! Closing the PR for now.

PR #90, opened 2020-10-16
Audio System

elocutura:
# --- Usage ---

SoundManager component should be in the first scene, it has DontDestroyOnLoad, and it will instance the emitters specified in initial pool size, so there is no need to add emitter manually.

To play a sound, the user will only have to get the singleton of the SoundManager and call the function PlaySound with the audioClip for 2D sounds and the default settings, if the user wants to play a spatial sound will have to override spatialBlend in the settings and pass the position as a value. 
Volume, pitch and other audioSource settings can be modified overriding the default SoundEmitterSettings value like we did with spatialBlend in the example.

Examples:
```
// for 2D sounds
SoundManager.singleton.PlaySound(clip, new SoundEmitterSettings);

// for 3D sounds
SoundManager.singleton.PlaySound(clip, new SoundEmitterSettings {SpatialBlend = 1}, position);

// 3D looping sound specifying the volume
SoundManager.singleton.PlaySound(clip, new SoundEmitterSettings {SpatialBlend = 1, Volume = 0.5f, Loop = true}, position);
```
The system currently does not support moving sounds providing a transform, but its planned.

It is possible to set and get the normalized volume of a chosen audio mixer group by parameter.
These two static functions will return true if the volume was set/get correctly and false if there was an error.

Examples:
```
// Set
SoundManager.SetGroupVolume( audioMixerGroup, 0.7f);

// Get
SoundManager.GetGroupVolume( audioMixerGroup, out normalizedVolume);
```

# --- Details ---

This PR attempts to create an easy extendable base for a sound system

**- Created SoundEmitter with basic functionality of playing sounds for further extension**
Sound Emitters are super simple right now as they will function as the AudioSources so we have more control over features we might want to implement in the future such as: 
    - fade out/fade in volume
    - following of a transform for the duration of the sound
    - cross fades
    - play a random sound of a list provided in the function parameters
    - ...
Once a sound emitter has stopped playing a sound it will become available for the SoundManager again via the object pooler, if need be, SoundEmitter could be extended to yield control of availability to a script for as long as needed. (but that would over complicate the script for something that most likely will not be needed)

**- Created class SoundEmitterSettings**

Settings structure with default for 2D sounds.

**- Created singleton SoundManager with following functionality:**

    - Set and Get the volume of a selected audio mixer
These 2 static functions are made so its easy to access the volume of a desired audio mixer group
The functions assume the exposed volume variable in said mixer group is called "Volume" since mixer groups are not currently implemented in the project.
The 2 helper functions below it are made so its effortless to transition from the normalized values that sliders usually use to display volume levels, to applying those values to the -80 to 0 range in the mixer groups as volume.

    - Managing a pool of SoundEmitter objects with dynamic extension
It creates a pool of X SoundEmitter (quantity specified in the inspector) to use at any point, when the SoundManager is asked to play a sound it will retrieve the next free SoundEmitter and setup all the parameters for it to function, if no SoundEmitter is free the pooler will instantiate a new one for the sound unless there are enough emitters to cover for all virtual voices in the project.
The pooler also trims all extra audio sources past the initial pool that have not been in use recently (recently marked by _extraTimeBeforeTrim variable).

    - Playing 2D and Spatial sounds through SoundEmitter
PlaySound function will call the playSound function of a free SoundEmitter with the specified clip and settings (defaulted to 2D sounds), these function returns the SoundEmitter so scripts that manage loop sounds such and ambience or soundtrack can interact with the SoundEmitter directly and pause, stop or change the clips as the gameplay needs. As well as avoid unnecessary searches for emitters in the SoundManager.


Deragaun (changes requested):
- At least the unused code should be removed.

UOP1_Project/Assets/Scripts/Sound/SoundEmitter.cs
```diff
@@ -0,0 +1,58 @@
+﻿using System.Collections;
+using System.Collections.Generic;
+using UnityEngine;
+using UnityEngine.Audio;
+
+[RequireComponent (typeof(AudioSource))]
+public class SoundEmitter : MonoBehaviour
+{
+	private AudioSource thisAudioSource;
```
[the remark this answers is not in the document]
davejrodriguez: We do actually: [Open Projects - Conventions](https://docs.google.com/document/d/1-eUWZ0lWREFu5iH-ggofwnixDDQqalOoT4Yc0NpWR3k/edit)

Should be renamed to _audioSource.

UOP1_Project/Assets/Scripts/Sound/SoundEmitter.cs
```diff
@@ -0,0 +1,58 @@
...
+	{
+		if (position == Vector3.zero)
+		{
+			thisAudioSource.spatialBlend = 0;
+		}
+		else
+		{
+			thisAudioSource.spatialBlend = 1;
+			this.transform.position = position;
+		}
```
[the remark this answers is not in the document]
elocutura: Thanks for the input, you are right that this was unnecessary and removes current functionality. If I make this change I guess it would be better to just use one function to play all sounds in sound manager, and avoid having PlaySpatialSound and Play2DSound right? or I conserve those and default the value of spartialBlend to 1 and 0 respectively

UOP1_Project/Assets/Scripts/Sound/SoundManager.cs
```diff
@@ -0,0 +1,195 @@
...
+
+		// Cycles through the list of emitters until it finds one in use or has completed a full cycle
+		while (toReturn == null)
+		{
+			if (poolIndex >= soundEmitterPool.Count)
+				poolIndex = 0;
+
+			if (poolIndex == startIndex) // If none of the sound emitters are available, create a new sound emitter and assign the index to the newly created one
+			{
+				toReturn = CreateNewSoundEmitter();
```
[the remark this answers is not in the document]
elocutura: I am aware of that and its something mentioned in the initial comment, its something that I'm planning on adding soon.

> This functionality can be expanded to remove SoundEmitter instances past the initial pool once they are no longer needed.

UOP1_Project/Assets/Scripts/Sound/SoundManager.cs
```diff
@@ -0,0 +1,195 @@
...
+		}
+	}
+
+	void Start()
+    {
+		InitEmitterPool();
+
+		float tempVal = Mathf.Infinity;
+
+		Debug.Log(tempVal == Mathf.Infinity);
```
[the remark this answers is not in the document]
elocutura: Removed in latest commit. Thanks for pointing that out

UOP1_Project/Assets/Scripts/Sound/SoundEmitter.cs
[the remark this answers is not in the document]
elocutura: Sorry, since we did not have code guidelines I just defaulted to what I use on my personal projects, its fixed on the latest commit for audioSource. Thanks!

UOP1_Project/Assets/Scripts/Sound/SoundManager.cs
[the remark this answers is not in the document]
davejrodriguez: Agree that eventually the pool should trim audio sources that haven't been played in a while. Along these lines - should we cap pool at `AudioConfiguration.numVirtualVoices`?

UOP1_Project/Assets/Scripts/Sound/SoundManager.cs
```diff
@@ -0,0 +1,191 @@
...
+	private static float NormalizedToMixerValue(float normalizedValue)
+	{
+		return -80 + (normalizedValue * 80);
+	}
+	#endregion
+	#endregion
+
+	#region play sounds functions
+	// Vector3.zero is used as a marker of a 2D sound
+	public SoundEmitter Play2DSound(AudioClip clip, AudioMixerGroup mixer, bool loop = false, float volume = 1, float pitch = 1, float spartialBlend = 0)
```
Deragaun, line 86: spatialBlend does not need to be a parameter for 2D sounds. Like you did in your previous version this is always 0 if i am not mistaken.

(also you misspelled it :) )

UOP1_Project/Assets/Scripts/Sound/SoundEmitter.cs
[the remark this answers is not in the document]
elocutura: Thanks for the like davejrodriguez, I will read it and update all code that doesn't follow the convention as soon as possible.

UOP1_Project/Assets/Scripts/Sound/SoundManager.cs
elocutura: woops, trying to manage too many things at one atm, fixed it.

UOP1_Project/Assets/Scripts/Sound/SoundManager.cs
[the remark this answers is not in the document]
elocutura: I think using AudioConfiguration.numVirtualVoices as maximum would probably defeat the purpose of having a garbage collection since all sounds past this would not work at all.
My initial thought was to garbage collect all the extra instances created past the initial batch.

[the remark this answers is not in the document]
Deragaun: @davejrodriguez I do not think that would be smart. We can have more sounds queued up than would actually be playing. So we can queue up 100 sounds, but only 32 of those are playing at one time for example. If one sound currently playing is done, freeing one of the virtual voices, another sound could chirp in. I am not quite sure if this works like that though, I do not know much about the unity sound system.

If sounds would not start playing either way after being "queued up" like i explained, than your proposal would be the correct choice.

Deragaun (approved):

UOP1_Project/Assets/Scripts/Sound/SoundManager.cs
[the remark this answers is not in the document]
davejrodriguez: >We could use this as a maximum of created instances though, and if that limit is reached just not play this sound.
Not sure what you thought I meant, but maximum created instances is it. And I'm not saying do this instead of garbage collection. I'm saying do it in addition. No point in instantiating 1000 voices if your virtual max is 512.

@Deragaun There is no queuing happening in this code. If 1000 sound emitters start playing, (with current Unity defaults) 32 of them will be playing and you will hear them. 480 will be playing but you will not hear them. 512 will not be playing and you will not hear them. That last 512 is pointless to instantiate. If you're going to add queuing functionality, don't instantiate the sound emitters until you need them and can actually use them.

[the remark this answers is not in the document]
elocutura: Sorry I misunderstood you then, we agree on that way of doing it then.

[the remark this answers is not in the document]
davejrodriguez: As previously stated, I don't think it's absolutely crucial for a first PR, especially so early on in the project. Just some ideas to think about.

davejrodriguez (changes requested):
Really good start!

UOP1_Project/Assets/Scripts/Sound/SoundManager.cs
```diff
@@ -0,0 +1,190 @@
...
+	{
+		float tempVolume = PlayerPrefs.GetFloat(mixer.name, Mathf.Infinity);
+
+		if (tempVolume != Mathf.Infinity)
+		{
+			mixer.audioMixer.SetFloat("Volume", tempVolume);
+		}
+		else
+			Debug.Log("There is no saved volume preferences for mixer " + mixer.name + ", could not load volume.");
+	}
```
davejrodriguez, line 69: I think setting/getting player prefs might be out of scope. Brief just calls for being able to set the volume via a settings menu. Probably best to leave changing player prefs to a settings menu or the save/load system.
```suggestion
	public static bool SetGroupVolume(AudioMixerGroup group, float volume)
	{
		return group.audioMixer.SetFloat("Volume", NormalizedToMixerValue(volume));
	}
	
	public static bool GetGroupVolume(AudioMixerGroup group, out float volume)
	{
	        return group.audioMixer.GetFloat("Volume", out volume);
	}
```

UOP1_Project/Assets/Scripts/Sound/SoundManager.cs
```diff
@@ -0,0 +1,190 @@
...
+	{
+		if (!mixer)
+			mixer = masterMixer;
+
+		SoundEmitter foundSoundEmit = GetSoundEmitter();
+
+		foundSoundEmit.PlaySound(clip, mixer, position, loop, volume, pitch, spartialBlend);
+
+		return foundSoundEmit;
+	}
```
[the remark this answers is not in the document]
Deragaun: Well, i thought about doing it that way as well and that is a possibility. Nonetheless does it make sense to have two methods here to split into 2D and 3D for clarity of purpose of those methods. These methods are public API so to speak after all. Other people will be using them and the usage should be as streamlined as possible.

I would prefer leaving this as is.

elocutura:
This weekend I am very busy and will probably not be able to answer all issues and resolve everything. I will try my hardest to get everything done by monday night tho. Sorry for the delay

UOP1_Project/Assets/Scripts/Sound/SoundManager.cs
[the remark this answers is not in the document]
davejrodriguez: I see what you are saying, but consider this - we may eventually want to be able to specify the values of many more AudioSource settings. I'd say there are about 19 relevant settings that one could conceivably want to change. Instead of writing a method signature 21 params long, it would make more sense to wrap those in an AudioSourceSettings structure with Unity's own defaults, let the user override the ones they want, and the pass the whole structure to be applied.

The method signature would then look like this: 
`public void PlaySound(AudioClip clip, AudioSourceSettings settings, Vector3 position = default)`

Usage for 2D sounds:
```
_soundEmitter.PlaySound(clip,AudioSourceSettings.Default);
```
The user just passes in the clip and default settings (because the default is no spatialization). Position has the default of Vector.zero and does not need to be specified in this case.

Usage for 3D sounds:
```
_soundEmitter.PlaySound(clip,new AudioSourceSettings{SpatialBlend=1f},transform.position);
```
The user overrides the spatial blend and sets a position.

Possibly more importantly, this would match Unity's own API. They do not have separate Play methods for 2D vs 3D. The spatialization is done via fields on the AudioSource component.

davejrodriguez:
No worries. I realized I asked for a lot of changes. I have some time this weekend and I am starting to implement the changes I suggested in my local copy of your PR. I'll push them up and you can take them or discard them and implement what you've been working on.

UOP1_Project/Assets/Scripts/Sound/SoundManager.cs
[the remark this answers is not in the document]
elocutura: This was something that had to be done eventually for the reasons you stated. I thought of doing a struct as the settings as a start but decided against it to maintain a more simple PR to start.

Your implementation looks really good, if you are working on this on your repository I'll get to work tomorrow with fleshing out the pooler with what we discussed before.

UOP1_Project/Assets/Scripts/Sound/SoundManager.cs
```diff
@@ -0,0 +1,190 @@
...
+
+	// GetSoundEmitter with return the next SoundEmitter that is not playing any sounds, if all the sound emitters are busy it will extend the pool
+	private SoundEmitter GetSoundEmitter()
+	{
+		SoundEmitter toReturn = null;
+		int startIndex = poolIndex;
+		poolIndex++;
+
+		// Cycles through the list of emitters until it finds one in use or has completed a full cycle
+		while (toReturn == null)
```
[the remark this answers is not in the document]
elocutura: You are completely right on that one, I am super rusty in LINQ and coding late night didn't help on that front.

UOP1_Project/Assets/Scripts/Sound/SoundEmitter.cs
```diff
@@ -0,0 +1,51 @@
...
+	{
+		DontDestroyOnLoad(this);
+		_audioSource = this.GetComponent<AudioSource>();
+	}
+
+	// Start is called before the first frame update
+	void Start()
+    {
+		_audioSource.playOnAwake = false;
+    }
```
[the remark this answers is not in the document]
elocutura: For now its just a simple wrapper but to add functionality like moving sounds we need monobehaviour, since afaik there is no trigger for when a sound is no longer playing, so we would need to use a coroutine to trigger events when its finished playing with the clip.length such as unparenting from the moving transform that is emitting the sound.

UOP1_Project/Assets/Scripts/Sound/SoundManager.cs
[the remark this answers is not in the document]
elocutura: Afaik there is 2 ways of doing this next part of the pooler.
Either we make it so extra instances remove and destroy themselves from the pool after clip.length seconds have passed. or SoundManager removes periodically all instances not in use that that are tagged as instances.

I would probably go with the second one since I think SoundManager should handle everything related to the list to avoid issues, and its also more performant overall given you don't have all the instances updating timer or running coroutines.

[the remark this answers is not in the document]
davejrodriguez: I was thinking about this too and I agree. I do not think the SoundEmitters should be able to remove themselves from the list. Not their job.

> SoundManager removes periodically all instances not in use that that are tagged as instances.

So, you'll probably want to give it some lifetime where it could exist, inactive, ready to play another clip if one were requested. Otherwise, you lose the performance gains you get by not instantiating/destroying. Could SoundEmitter maintain a timestamp of when it stops being in use? Then every trimming cycle SoundManager could check the current time against the time the emitter stopped and destroy the ones greater than some lifetime? This all could be done in SoundManager too (marking first time !IsInUse and destroying if detected again longer than lifetime), but _could_ generate false positives depending on trim frequency.

UOP1_Project/Assets/Scripts/Sound/SoundEmitter.cs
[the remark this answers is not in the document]
DeivSky: Makes sense! Although that made me think about making a dispatcher... I might work on that 🙂

UOP1_Project/Assets/Scripts/Sound/SoundManager.cs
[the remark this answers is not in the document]
elocutura: Sounds good, I'll have a first pass implementation tonight.

UOP1_Project/Assets/Scripts/Sound/SoundEmitter.cs
```diff
@@ -0,0 +1,72 @@
...
+
+	public void PlaySound(AudioClip clip, SoundEmitterSettings settings, Vector3 position = default)
+	{
+		_audioSource.clip = clip;
+		ApplySettings(_audioSource, settings);
+		_audioSource.transform.position = position;
+		_lastUseTimestamp = Time.realtimeSinceStartup + clip.length;
```
davejrodriguez, line 31: This will fail for looped sounds. Trying to think of a way to detect when IsInUse resolves to false for the first time that doesn't involve a coroutine polling for it.

elocutura: isPlaying is always true for looped sounds, the places where timestamp is checked we also check if isPlaying so for now no issues at all.

davejrodriguez: > Using Mathf.infinity for loops would do the trick
 
Would this cause the looped source to never be trimmed even if a user called StopAllLoopedSounds?

elocutura: You are right, that's a problem. Do you think making the both the StopAllSounds and StopAllLoopedSound to just set the timestamp to te current timestamp to solve that?

davejrodriguez: Yeah, I think that's a good solution.

davejrodriguez: Hmm wondering if pooling and sound management are getting too coupled. Two thoughts:
1. Trimming is nice to have for memory but not a requisite for the object pool pattern. Maybe it's something that can be expanded upon?
2. I think it might be better to yield control of the SoundEmitter to the user. Then they could loop, play one shot, etc. to their heart's content. Once done, they can return it to the pool and you have a clear point at which the SoundEmitter is no longer in use.

elocutura: I don't understand first question, you say its not a requisite and its getting too coupled but want to expand on it?? I think the pooler is mostly done and I would not touch anything else for a first PR.

As for the second point, I programmed it centralized like this initially because that is what was asked in the forum thread, doing what you are purposing would be the complete opposite of what they are asking for.
From the thread:
> Characters, effects, events, can ask the audio system to play a sound ...

elocutura:
@ciro-unity why is this labeled as an enhancement?? this is an entire audio system for the game, I'm confused

UOP1_Project/Assets/Scripts/Sound/SoundEmitter.cs
davejrodriguez: Sorry. "Expand on" was a poor choice of words. I meant leave trimming functionality out for now and come back to it later.

And you're right, keeping control centralized fits the brief better. I was just thinking how easy it would be to Lerp position, volume, pitch, etc. if we had continuous control from user scripts.

Solid first PR I think. Nice work! 👍

davejrodriguez:
@elocutura It just marks it as a new feature. Not a bad thing. Just want to avoid that invalid label :)

UOP1_Project/Assets/Scripts/Sound/SoundManager.cs
```diff
@@ -0,0 +1,163 @@
+﻿using System.Collections;
+using System.Linq;
+using System.Collections.Generic;
+using UnityEngine;
+using UnityEngine.Audio;
+
+public class SoundManager : MonoBehaviour
+{
+	public static SoundManager singleton;
```
davejrodriguez, line 9: So I have one nitpick and one suggestion that probably needs to happen:

Nitpick: Conventionally, the singleton pattern calls its static instance `Instance`. At the very least, it should be `Singleton`, since it's a public field.
Suggestion:
 ```suggestion
       public static SoundManager Instance;
         
       [RuntimeInitializeOnLoadMethod(RuntimeInitializeLoadType.SubsystemRegistration)]
       static void Init()
       {
               Instance = default;
       }
```
This makes the static variable compatible with Fast Enter Play Mode, which is a requirement in the conventions doc.

File: UOP1_Project/Assets/Scripts/Sound/SoundManager.cs [hunk 12]
```diff
@@ -0,0 +1,163 @@
...
+	public int initialPoolSize = 1;
+	[Tooltip ("Frequency at which the sound manager will delete extra emitters that are not in use")]
+	public float trimFrequency = 5;
+	[Tooltip ("Extra time after instanced emitters are not in use anymore before destroying them")]
+	public float extraTimeBeforeTrim = 5;
+
+	[Header ("Mixer Groups")]
+	public AudioMixerGroup masterMixer;
+	public AudioMixerGroup sfxMixer;
+	public AudioMixerGroup musicMixer;
```
davejrodriguez, line 22: Are these used anymore?

UOP1_Project/Assets/Scripts/Sound/SoundManager.cs
```diff
@@ -0,0 +1,163 @@
...
+	{
+		DontDestroyOnLoad(this);
+
+		if (!SoundManager.singleton)
+			singleton = this;
+		else
+		{
+			Debug.LogWarning("One sound manager already present in the game, destroying " + this.name + ".");
+			Destroy(this);
+		}
```
davejrodriguez, line 37:
```suggestion
		if (SoundManager.singleton)
			Debug.LogWarning("One sound manager already present in the game, destroying " + this.name + ".");
			Destroy(this);
		else
		{
			singleton=this;
			DontDestroyOnLoad(this);
		}
```
Two reasons:
1. DontDestroyOnLoad should only be called on the one singleton instance. It's a waste to call it on an object you're about to destroy anyway.
2. I prefer to evaluate the most likely condition first. The reason for this is more apparent with if-elseif-else where there would be multiple conditions to evaluate. By evaluating less likely conditions first, you lose the possibility of an early out. In this case with just an if-else it makes no difference. Purely, just a preference. But thought I'd mention it.

DeivSky: The way I see it is, evaluate the condition that would make you not go through the usual process, that way you can remove the else altogether. It looks cleaner, and it's easier to understand that if that condition is met, you're just returning.
``` suggestion
		if (SoundManager.singleton)
              {
			Debug.LogWarning($"One sound manager already present in the game, destroying {name}.");
			Destroy(this);
                      return;
              }

        	singleton = this;
		DontDestroyOnLoad(this);
		
```
(Also added string interpolation)

davejrodriguez: +1 for string interpolation.

Yes, in your suggestion, I believe return is necessary. Destroy does not work immediately, so Awake would continue, singleton would be set, and then destroyed. I think I still prefer explicit else in this case, but works either way.

UOP1_Project/Assets/Scripts/Sound/SoundManager.cs
elocutura: I was under the impression that [RuntimeInitializeOnLoadMethod] was inconsistent and never used it, I'll take a look at it today and probably accept the commit.

UOP1_Project/Assets/Scripts/Sound/SoundManager.cs
elocutura: Dave is right, Destroy doesn't work instantly.
Will update the PR with these suggestions during the weekend since all are good changes.

UOP1_Project/Assets/Scripts/Sound/SoundManager.cs
elocutura: Not anymore, they were just common groups used for the first implementation. I will delete these variables in the next commit.

elocutura:
Currently the system is now using the Pooling from @davejrodriguez , there are a couple of issues known with the current implementation.
- Its less performant since SoundEmitters run a coroutine to return themselves to the pool after they are done playing a sound
- For some reason after the OnReturn() in the PoolSO class the member is never pushed back to the stack
Now the pool is not dynamic but since the systems are decoupled now it can be implemented in its own module later down the line if needed.

UOP1_Project/Assets/Scripts/Sound/SoundEmitter.cs
```diff
@@ -24,13 +24,29 @@ void Start()
...
+	public void OnRequest()
+	{
+		gameObject.SetActive(true);
+	}
+
+	public void OnReturn(Action onReturned)
+	{
+		StopSound();
+		gameObject.SetActive(false);
+	}
```
davejrodriguez, line 36: You need to call `onReturned.Invoke();` after you complete the reset of your object. This is why objects are not being returned to the pool. It's a callback to allow for resets that must be completed over several frames. Take a look at the Example scene and check out PoolableParticle for an example implementation. For a graceful return of a SoundEmitter, you likely want to stop looping if it's looping, wait for the current playback to be finished, Stop(), and then return.

I'm currently working on documentation for Factory and Pool. This is helpful. I'll be sure to emphasize the importance of calling the callback. Or maybe we can think of something more foolproof?

UOP1_Project/Assets/Scripts/Sound/SoundManager.cs
```diff
@@ -6,43 +6,51 @@
...
-    }
+	private void InitPool()
+	{
+		_factory = ScriptableObject.CreateInstance<SoundEmitterFactorySO>();
+		_factory.Prefab = soundEmitterPrefab;
+		_factory.Prefab.name = "SoundEmitter Factory";
+		_pool = ScriptableObject.CreateInstance<SoundEmitterPoolSO>();
+		_pool.name = "SoundEmitter Pool";
+		_pool.Factory = _factory;
+		_pool.InitialPoolSize = initialPoolSize;
```
davejrodriguez, line 52: I realize this is the example code given in the PR and LocalPoolTester example, but just so you know, you can create these factories and pools as assets of the project and plug them into the inspector as object references. Unless you _need_ to create these at runtime, I'd definitely recommend creating as project assets.

UOP1_Project/Assets/Scripts/Sound/SoundManager.cs
```diff
@@ -6,43 +6,51 @@
...
+	[SerializeField]
+	private SoundEmitter soundEmitterPrefab = default;
 
-	[Header ("Mixer Groups")]
-	public AudioMixerGroup masterMixer;
-	public AudioMixerGroup sfxMixer;
-	public AudioMixerGroup musicMixer;
+	SoundEmitterFactorySO _factory;
+	SoundEmitterPoolSO _pool;
+	public SoundEmitterPoolSO Pool { get { return _pool; } }
```
davejrodriguez, line 20: Do you need public access of the pool?

UOP1_Project/Assets/Scripts/Sound/SoundEmitter.cs
```diff
@@ -76,4 +92,10 @@ public float LastUseTimestamp()
 	{
 		return _lastUseTimestamp;
 	}
+
+	IEnumerator FinishedPlaying(float clipLength)
+	{
+		yield return new WaitForSeconds(clipLength);
+		SoundManager.Instance.Pool.Return(this);
+	}
```
davejrodriguez, line 100: This logic should be contained within the SoundEmitter's return sequence, rather than conditionally run if it's not a looping sound. Also, ideally, the pool object would not know about it's containing pool. I still think SoundEmitters _themselves_ should be decentralized to the users, with SoundManager as a provider. But if you're looking to keep it entirely decentralized, you can have SoundManager Return the SoundEmitter right after calling PlaySound().

davejrodriguez (commented):
Left some comments. I definitely need to add some documentation and more examples. I will continue to work on those.

MansyrevAY:
Hey, I've tried to pull a fork from @elocutura, but I have missing references from the UOP1 namespace, and I couldn't find them in his repo. Is there something I am missing?

ciro-unity:

ciro-unity:
Mega refactoring done. A couple of notes:
- I removed the singleton pattern, implemented the AudioManager as a simple component which will go in the Instantiation scene, and listens to an event Scriptable Object on which any object can request a sound to be played
- Implemented the AudioCue Scriptable Objects to play different sounds at the same time, as done by Peter77 on the forums

Merging now!

If you're on the Unity Discord, as usual: drop your username here and I'll add you the role Open Projects Contributor!

UOP1_Project/Assets/Scripts/Audio/SoundEmitters/SoundEmitter.cs
```diff
@@ -0,0 +1,100 @@
...
+	{
+		_audioSource = this.GetComponent<AudioSource>();
+		_audioSource.playOnAwake = false;
+	}
+
+	public void OnRequest() { }
+
+	public void OnReturn(Action onReturned)
+	{
+		StopSound();
```
davejrodriguez, line 27: With the current pool implementation, SoundEmitters will never be returned to the pool if they do not invoke onReturned. PoolableParticle in the example files shows how to do this.

Note: PR #159 makes it so resetting is entirely the responsibility of the pooled object or its manager and does not require any callback. This code will still need to be adjusted in the case that PR is merged.

PR #95, opened 2020-10-18
Update Unity Version to 2019.4.12f1 LTS

SoundGuy:
Just updated the version file to the latest Unity 2019.4.12f1


SoundGuy:
didn't realize this was not allowed

PR #98, opened 2020-10-19
FSM using MonoBehaviours. +Examples using SO and plain C# classes

STARasGAMES:
forum: https://forum.unity.com/threads/character-state-machine.979371/page-3#post-6432053
card: https://open.codecks.io/unity-open-project-1/decks/15/card/157-state-machine

This PR is not ready to merge. It is here just to show another way to do FSM.

I like SO a lot! But when you need Component-like behavior you're using MonoBehaviours, don't you?
That's why I implemented this quite simple example using MB instead of SO.
Also, there is a very generic StateMachine that can be used without both MB and SO. 

Cons:
- a little overhead from enabling/disabling whole GameObjects.
- someone may use FSM prefabs like SO and that really bad!!! **(no)**.

Pros:
- It can use standard Unity components out of the box 
- Everybody can just write Monobehaviours with OnEnable/OnDisable
- It can be built/stored either in a scene or a prefab
- It can be reused and modified via prefab variants
- It even supports nested state machines
- It doesn't require to do extra steps to instantiate the whole FSM graph
- It does not require any editor scripts/UI (state changes can be observed in the hierarchy)
- It would be possible to separate states from transitions if necessary
- It is possible to create a custom node editor for it (which is easy to add on top) but it is not necessary at all.
- **It just like SO but a lot easier to implement and maintain**

(thanks peaj_metric for this list of pros and cons)

### Content
In the root of `Assets/Scripts/StateMachines` folder you will find a bunch of interfaces that are enough to create any kind of FSM out of them:
- IState
- IStateAction
- ITransition
- ICondition

Also, there is a generic StateMachine that implements core FSM functionality.[here](https://github.com/UnityTechnologies/open-project-1/pull/98/files?file-filters%5B%5D=.cs&file-filters%5B%5D=.meta&file-filters%5B%5D=.unity#diff-8df56ec9e13227377df9130b58db155c0633e1222b462471e556dc06932385d1)

`Assets/Scripts/StateMachines/Mono` - FSM using MonoBehaviours. 
I want to draw your attention to MonoStateMachine which is not repeating FSM's core functionality and uses generic StateMachine instead. 

`Assets/Scripts/StateMachines/Scriptable` - FSM using ScriptableObjects. 
Again, ScriptableStateMachine doesn't implement FSM by itself. My SO-FSM uses a transition table rather than store transitions right in the state.

`Assets/Scripts/StateMachines/Plain` - FSM using plain C# classes. 
Nothing special, just a way to create StateMachine from code. Includes PlainStateMachineBuilder as a fluent builder. Just an example that this stuff is quite easy to implement if we really need to.

`Assets/Scripts/PlayerStateMachine` - FSM that implements current player movement using the MonoBehaviours approach.
Just a bunch of StateActions and Conditions. 
Important: I had to create CharacterMotor(that is very similar to Character class) as a data container for PlayerStateActions. 
Please also note, I spent on this very little time, so yeah... PlayerStateActions are poorly implemented, but who cares, it's just an example of a thing that no one interested in, XD

`Assets/Tests/PlayerFSM` - test scene where I did my dirty tricks. It contains 3 player characters: 
 - green - current, default
 - blue - FSM version
 - violet - prefab variant from previous(blue) one that adds ability to double jump

That how player's FSM looks in editor:
image

And that how looks FSM with added double jump:
image

Video: https://youtu.be/iPtt0snKloI

### DIFF
--- a/UOP1_Project/Assets/Materials/CharacterFSM DoubleJump.mat
+++ b/UOP1_Project/Assets/Materials/CharacterFSM DoubleJump.mat
@@ -1,0 +1,90 @@
+%YAML 1.1
+%TAG !u! tag:unity3d.com,2011:
+--- !u!21 &2100000
+Material:
+  serializedVersion: 6
+  m_ObjectHideFlags: 0
+  m_CorrespondingSourceObject: {fileID: 0}
+  m_PrefabInstance: {fileID: 0}
+  m_PrefabAsset: {fileID: 0}
+  m_Name: CharacterFSM DoubleJump
+  m_Shader: {fileID: 4800000, guid: 933532a4fcc9baf4fa0491de14d08ed7, type: 3}
+  m_ShaderKeywords: 
+  m_LightmapFlags: 4
+  m_EnableInstancingVariants: 0
+  m_DoubleSidedGI: 0
+  m_CustomRenderQueue: 2050
+  stringTagMap:
+    RenderType: Opaque
+  disabledShaderPasses: []
+  m_SavedProperties:
+    serializedVersion: 3
+    m_TexEnvs:
+    - _BaseMap:
+        m_Texture: {fileID: 0}
+        m_Scale: {x: 1, y: 1}
+        m_Offset: {x: 0, y: 0}
+    - _BumpMap:
+        m_Texture: {fileID: 0}
+        m_Scale: {x: 1, y: 1}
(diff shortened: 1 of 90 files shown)


ciro-unity:
Thanks @STARasGAMES , in the end we chose a different one mainly because we feel that having everything as a GameObject + MonoBehaviour has its big cost!

Issue #96, opened 2020-10-18
Upgrade to latest 2019.4 LTS version


Remwrath:
Latests LTS is now 2019.4.12f1 ;)

MEGATANK58:
I don't think we need a upgrade yet,there is no specific reason to do so and most of the people can't just keep updating and redownloading Unity every week or so just to be latest.

SoundGuy:
@MEGATANK58  you don't really need to update your installation between minor updates. I'm on 2019.4.12f but i don't commit the  version file. 

it would be nice if it was upgraded so i won't have that version file changed in my working copy - becuase i need to stash or discard it every time i check out another branch (along with a few other annyoying files i wish there was a way to just tell git to ignore because i will never commi them myself)

SoundGuy:
also the packages in my unity installation are newer and then change the manifest file every time i open the project

{
  "dependencies": {
    "com.unity.addressables": "1.8.5",
    "com.unity.cinemachine": "2.6.0",
    "com.unity.ide.rider": "1.1.4",
    "com.unity.ide.vscode": "1.2.1",
    "com.unity.inputsystem": "1.0.0",
    "com.unity.polybrush": "1.0.2",
    "com.unity.probuilder": "4.2.3",
    "com.unity.progrids": "3.0.3-preview.6",
    "com.unity.render-pipelines.universal": "7.3.1",
    "com.unity.settings-manager": "1.0.3",
    "com.unity.textmeshpro": "2.0.1",
    "com.unity.timeline": "1.2.15",
    "com.unity.ugui": "1.0.0",
    "com.unity.modules.ai": "1.0.0",
    "com.unity.modules.androidjni": "1.0.0",
    "com.unity.modules.animation": "1.0.0",
    "com.unity.modules.assetbundle": "1.0.0",
    "com.unity.modules.audio": "1.0.0",
    "com.unity.modules.cloth": "1.0.0",
    "com.unity.modules.director": "1.0.0",
    "com.unity.modules.imageconversion": "1.0.0",
    "com.unity.modules.imgui": "1.0.0",
    "com.unity.modules.jsonserialize": "1.0.0",
    "com.unity.modules.particlesystem": "1.0.0",
    "com.unity.modules.physics": "1.0.0",
    "com.unity.modules.physics2d": "1.0.0",
    "com.unity.modules.screencapture": "1.0.0",
    "com.unity.modules.terrain": "1.0.0",
[...]

PR #110, opened 2020-10-26
DRAFT: Build via GH Actions

kreghek:
**[for Generic PRs]**  
**Do you have a forum thread linked to this PR?**
No

**What will this PR bring to the project for everyone?**
You can dowload game, play and fast see/

**Why are these changes necessary?**
Add UNITY_LICENCE to github secrets https://unity-ci.com/docs/github/activation

**How did you implement them?**
main.yml is workflow to build project. After success build you can download artefacts under different arget planform.

### DIFF
--- a/.github/workflows/activation.yml
+++ b/.github/workflows/activation.yml
@@ -1,0 +1,19 @@
+name: Acquire activation file
+on: [push]
+jobs:
+  activation:
+    name: Request manual activation file 🔑
+    runs-on: ubuntu-latest
+    steps:
+    # Request manual activation file
+    - name: Request manual activation file
+      id: getManualLicenseFile
+      uses: webbertakken/unity-request-manual-activation-file@v1.1
+      with:
+        unityVersion: 2019.4.12f1
+    # Upload artifact (Unity_v20XX.X.XXXX.alf)
+    - name: Expose as artifact
+      uses: actions/upload-artifact@v1
+      with:
+        name: ${{ steps.getManualLicenseFile.outputs.filePath }}
+        path: ${{ steps.getManualLicenseFile.outputs.filePath }}
--- a/.github/workflows/main.yml
+++ b/.github/workflows/main.yml
@@ -42,7 +42,7 @@ jobs:
           restore-keys: |
             Library-${{ matrix.projectPath }}-
             Library-
-      - uses: webbertakken/unity-builder@<version>
+      - uses: webbertakken/unity-builder@1.3
         with:
           projectPath: ${{ matrix.projectPath }}
           unityVersion: ${{ matrix.unityVersion }}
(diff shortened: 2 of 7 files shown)


.github/workflows/activation.yml
```diff
@@ -0,0 +1,19 @@
...
+  activation:
+    name: Request manual activation file 🔑
+    runs-on: ubuntu-latest
+    steps:
+    # Request manual activation file
+    - name: Request manual activation file
+      id: getManualLicenseFile
+      uses: webbertakken/unity-request-manual-activation-file@v1.1
+      with:
+        unityVersion: 2019.4.12f1
```
Eyap53, line 13: You targeted version 2019.4.12f1, but you also pushed "ProjectVersion.txt" with 2019.4.13f1.

UOP1_Project/ProjectSettings/ProjectVersion.txt
```diff
@@ -1,2 +1,2 @@
-m_EditorVersion: 2019.4.11f1
-m_EditorVersionWithRevision: 2019.4.11f1 (2d9804dddde7)
+m_EditorVersion: 2019.4.13f1
```
Eyap53, line 1: Should not be pushed with your changes. See #96.

.github/workflows/main.yml
```diff
@@ -0,0 +1,45 @@
...
+  buildForSomePlatforms:
+    name: Build for ${{ matrix.targetPlatform }} on version ${{ matrix.unityVersion }}
+    runs-on: ubuntu-latest
+    strategy:
+      fail-fast: false
+      matrix:
+        projectPath:
+          - UOP1_Project
+        unityVersion:
+          - 2019.4.12f1
```
Eyap53, line 20: Same here.

Eyap53 (changes requested):
Can we find a way to keep the version defined in the workflow in sync with unity version of the project ?
I think we could add a step in it to detect the version, without writing the version explicitly.

kreghek:
> Can we find a way to keep the version defined in the workflow in sync with unity version of the project ?
> I think we could add a step in it to detect the version, without writing the version explicitly.

I will try

UOP1_Project/ProjectSettings/ProjectVersion.txt
kreghek: Ok.

.github/workflows/main.yml
kreghek: ok

UOP1_Project/ProjectSettings/ProjectSettings.asset
```diff
@@ -584,7 +585,7 @@ PlayerSettings:
   scriptingDefineSymbols: {}
   platformArchitecture: {}
   scriptingBackend:
-    Standalone: 1
+    Standalone: 0
```
kreghek, line 588: I draw **your** attention to the fact that I had to disable IL2CPP for the assembly. Otherwise, there will be an error that IL2CPP is not installed.

mohammad-ios (approved):

Issue #111, opened 2020-10-26
The state machine stops working when a state transition has many conditions


GFX5000i:
**Short description**
When a state transition has 4 or more conditions in a specific setup the exception _ArgumentOutOfRangeException_ is thrown and the character stops moving.

**Expected behaviour**
The state transition should allow as many conditions as needed and with any setting without throwing errors.

**To Reproduce**
1. Go to _Assets/ScriptableObjects/Protagonist/Transitions/BeginJumpDescent.asset_
2. Set the conditions as shown in the print screen attached
3. Enter Play Mode
4. The exception is thrown, the state machine stops and the character won't move

**Notes**
- The full stack trace: [Stack Trace.txt](https://github.com/UnityTechnologies/open-project-1/files/5441977/Stack.Trace.txt)
- The exception source: _Assets/Scripts/StateMachine/ScriptableObjects/StateTransitionSO.ProcessConditionUsages()_

![StateTransition Exception](https://user-images.githubusercontent.com/16827309/97237782-9c68a800-17c6-11eb-8a5f-bdb811b8f5ea.png)

DeivSky:
Hey, good catch! Can I ask you to see if you can replicate it without repeating any conditions?

PR #112, opened 2020-10-26
Fixed the StateTransitionSO error when grouping conditions

GFX5000i:
Issue: https://github.com/UnityTechnologies/open-project-1/issues/111

The source of the exception was the variable _idx_ following the condition index instead of the last index of _resultGroupsList_
This would give errors every time their values differ since it is expected to have less groups than conditions.

To test, reproduce the steps on the issue or add a bigger number of conditions with different operator combinations, specially chaining AND operators together and some OR operators between those chains.


DeivSky:
Tested and working!

ciro-unity:
I'll trust you on the testing, @DeivSky!

ciro-unity:
@GFX5000i thanks for the fix. I have assumed you are `Wagner GFX#8707` on Discord so I gave the OP Contributor role to that.

GFX5000i:
> 
> 
> @GFX5000i thanks for the fix. I have assumed you are `Wagner GFX#8707` on Discord so I gave the OP Contributor role to that.

Yup, that is me, thanks for role. 😁

PR #117, opened 2020-10-27
Generic (Scriptable) Object Pool

davejrodriguez:
## Forum Discussion
[Generic Object Pool Forum discussion](https://forum.unity.com/threads/generic-object-pool.990776/)

## What is object pooling?

Object pooling aims to alleviate the performance hit of instantiating/destroying many objects by activating/deactivating the objects instead, reusing objects when needed. A generic object pool could provide various systems in this game (audio, particles, character + projectile spawns) with consistent object pooling functionality.

## Usage

Because pooled objects are created on-demand, the pool must be initialized with a way to create the pooled object correctly. We can use a factory for this.

### Setup:
```cs
MyObjectFactory _factory = ScriptableObject.CreateInstance<MyObjectFactory>();
_factory.Prefab = _prefab;
MyObjectPool _pool = ScriptableObject.CreateInstance<MyObjectPool>();
_pool.name = gameObject.name;
_pool.Factory = _factory;
_pool.InitialPoolSize = _initialPoolSize;
```
### Use:
```cs
List<MyObject> myObjects = _pool.Request(10) as List<MyObject>;
foreach (MyObject object in myObjects)
{
	object.DoThatThing();
}
_pool.Return(myObjects);
```
## Extending
Custom poolable types implement IPoolable.
Custom factories implement IFactory\<T\>

### Stale State
One of the pitfalls of object pooling is stale state or state that survives from one use to another. This PR tries to address this is by enforcing the implementation of IPoolable.Reset for poolable types. 
### Gradual Reset
It might be desirable reset the state over time, so IPoolable.Reset takes a callback that notifies the pool once the reset is complete.
### Complex Resets
Some objects, particularly physics objects, may be difficult or impossible to reset properly. This PR does not attempt to address this issue directly. My thinking is that if the object cannot be properly reset at the object level, I do not know what can be done at the pool level. But perhaps someone has an idea for that?
### Continued Reference Access After Return
For example, calling `_pool.Return(myPooledObject);` will tell the pool to move the pooled object from the unavailable collection to the available collection. But the user still maintains a reference to the `myPooledObject`. It would be better to nullify the user's reference after returning. I experimented using the ref keyword, but was unable to get something working. If there is agreement that this functionality is desirable, would someone like to take a crack at it?
### Too Much Access
In my example, PoolableParticle inherits from Monobehaviour and lives on the same object as the particle system. A user could conceivably manually delete the pooled object. Perhaps, it would be better for pooled objects to be regular classes that just have a private reference to a generated Monobehaviour?


DeivSky (commented):
Nice, I'm leaving just a couple of comments

UOP1_Project/Assets/Scripts/Pool/Pool.cs
```diff
@@ -0,0 +1,58 @@
...
+namespace OP1.Pool
+{
+	/// <summary>
+	/// A generic pool that generates members of type T on-demand via a factory.
+	/// </summary>
+	/// <typeparam name="T">Specifies the type of elements in the pool.</typeparam>
+	public class Pool<T> : IEnumerable where T : IPoolable
+	{
+		public Stack<T> _available = new Stack<T>();
+		public HashSet<T> _unavailable = new HashSet<T>();
```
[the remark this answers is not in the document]
davejrodriguez: I know! Could've sworn I had them private. Even named them private. I think my IDE is playing tricks on me :)

UOP1_Project/Assets/Scripts/Pool/Pool.cs
```diff
@@ -0,0 +1,58 @@
...
+		{
+			this._factory = factory;
+
+			for (int i = 0; i < initialPoolSize; i++)
+			{
+				_available.Push(_factory.Create());
+			}
+		}
+
+		public T Request()
```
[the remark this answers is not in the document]
davejrodriguez: I thought about doing this, but wanted to keep first iteration light. Would definitely be handy to grab a bunch at a time.

UOP1_Project/Assets/Scripts/Pool/Pool.cs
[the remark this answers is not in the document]
davejrodriguez: So the reasoning here is that by maintaining a reference to unavailable objects, objects created outside the pool could not be added to the pool via Return.

However, I'm seeing now that I forgot to do that `_unavailable.Contains` check I meant to do in Return, so that's why that the collection looks useless.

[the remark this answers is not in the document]
DeivSky: I see! Although now I'm thinking, is it necessary to have that constraint? We could have a system in which object `A` sends `Foo` to object `B`, `B` pools `Foo` to then be able to send it to `C` or back to `A` or whatever? We would be capable to reuse the same instance when possible. And if there's a case in which `B` needs to send out one before `A`, it'd be capable of spawning it on its own. 
Thinking as I type, that'd be solved by having a shared pool for `A` and `B` :P

But my main question came from, what if `Foo` gets destroyed or lost reference of before returning it? GC would never run on it because of the reference inside the pool, yet it'd never be reused. I guess that'd be a misusage of the pool but it'd cause leaks that would be hard to track.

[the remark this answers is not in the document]
davejrodriguez: As much as I liked the idea of a closed system that could only pool the objects it made, you're right. The possibility of leaks outweighs the benefit of maintaining the reference. I've made the change. Pool now only has an available Stack.

davejrodriguez:
With the latest changes, there is a bit of a change in how objects are pooled. I removed the "parenting" functionality from ComponentFactory, because I felt like that was out of scope for the factory class. Therefore, objects currently pool into the root of the scene. I will re-implement this functionality as part of a ComponentPool class, where I feel it will be better suited. But I wanted to get latest stable changes out there for people to test.

davejrodriguez:
I accidentally added one of my scripts that depended on some ScriptableObject work I was doing, so I decided to add it all so there wouldn't be any missing dependencies.

Please test it out/look over the code. A little messy. And I would avoid creating a ParticlePool scriptable object instance, unless you'd like to debug it. It creates objects in the scene at editor time if you click on it. Still a WIP :)

davejrodriguez:
Alright, Pools/Factories as scriptable objects seem to be working well. I've edited the example scene to showcase how multiple user scripts could reference the same pool asset which lazily-instantiates a pool once it receives the first request. Each instance of the script requests 10 particles, creating a single pool of 20.

UOP1_Project/Assets/Scripts/Pool/Pool.cs
```diff
@@ -0,0 +1,63 @@
...
+	/// <typeparam name="T">Specifies the type of elements to pool.</typeparam>
+	public abstract class Pool<T> : ScriptableObject, IPool<T> where T : IPoolable
+	{
+		protected readonly Stack<T> _available = new Stack<T>();
+		public abstract IFactory<T> Factory { get; set; }
+
+		public virtual T Add()
+		{
+			return Factory.Create();
+		}
```
DeivSky, line 19: I don't get this method. It's called `Add` but it doesn't add the new instance to the pool? And I'm thinking maybe you meant to make it protected, not public. It would be very misleading to have a public `Add` method that returns a new instance just like `Request`. What's the difference, and, since it returns an instance, does that mean I can use it, or -as the name suggests- is it added to the pool and will I have problems if use it and then call `Request` and it hands me that same instance?

I see that you override it in the `ComponentPool` to set the parent when creating a new instance, so the idea is to be able to initialize the object when first instantiating it. Makes sense, but remember to make it protected, and I'd change the name to `Create` since it calls `Factory.Create()`.

UOP1_Project/Assets/Scripts/Pool/Pool.cs
```diff
@@ -0,0 +1,63 @@
...
+
+		public virtual T Request()
+		{
+			if (_available.Count <= 0)
+			{
+				_available.Push(Add());
+			}
+			T member = _available.Pop();
```
DeivSky, line 27: No need to `Push` and immediately `Pop` :)
``` suggestion
			T member = _available.Count > 0 ? _available.Pop() : Add();
```

UOP1_Project/Assets/Scripts/Pool/Pool.cs
```diff
@@ -0,0 +1,63 @@
...
+			member.Initialize();
+			return member;
+		}
+
+		public virtual IEnumerable<T> Request(int num = 1)
+		{
+			List<T> members = new List<T>();
```
DeivSky, line 34: When you know the number of elements, use either an array, or the `List` overload that takes in the capacity. 
``` suggestion
			List<T> members = new List<T>(num);
``` 
Depending on the size, the performance hit can be quite noticeable.

UOP1_Project/Assets/Scripts/Pool/ComponentPool.cs
```diff
@@ -0,0 +1,55 @@
...
+		{
+			T newMember = base.Add();
+			newMember.transform.SetParent(_poolRootObject.transform);
+			return newMember;
+		}
+
+		public override void OnDisable()
+		{
+			base.OnDisable();
+			DestroyImmediate(_poolRootObject);
```
DeivSky, line 52: Use `DestroyImmediate` only in the editor.
``` suggestion
#if UNITY_EDITOR
			DestroyImmediate(_poolRootObject);
#else
			Destroy(_poolRootObject);
#endif
```

UOP1_Project/Assets/Scripts/Pool/ComponentPool.cs
```diff
@@ -0,0 +1,55 @@
...
+
+		private void InitializePool()
+		{
+			_poolRootObject = new GameObject(name);
+			DontDestroyOnLoad(_poolRootObject);
+			for (int i = 0; i < InitialPoolSize; i++)
+			{
+				_available.Push(Add());
+			}
+		}
```
DeivSky, line 22: I might be blind but, is this method called anywhere? Looks like `OnEnable` stuff but I don't see it 👀

DeivSky (changes requested):
I really like the idea of having SO pools and factories! Leaving a few more comments this time, keep it up!

UOP1_Project/Assets/Scripts/Factory/ComponentFactory.cs
```diff
@@ -0,0 +1,23 @@
...
+	/// <summary>
+	/// Implements the IFactory interface for Component types.
+	/// </summary>
+	/// <typeparam name="T">Specifies the component to create.</typeparam>
+	public abstract class ComponentFactory<T> : ScriptableObject,IFactory<T> where T : Component
+	{
+		public abstract T Prefab
+		{
+			get;
+			set;
```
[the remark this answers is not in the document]
davejrodriguez: Sorry, I should have included an example of runtime pool/factory creation as well. Normally, I'd agree, but may be necessary in this case to support CreateInstance. I'll be sure to add an example.

UOP1_Project/Assets/Scripts/Pool/Pool.cs
davejrodriguez: I had it as Create. Not sure what I was thinking when I renamed. Makes more sense as Create.

UOP1_Project/Assets/Scripts/Pool/IPool.cs
```diff
@@ -0,0 +1,16 @@
+﻿using System.Collections.Generic;
+
+namespace OP1.Pool
+{
+	/// <summary>
+	/// Represents a collection that pools objects of T.
+	/// </summary>
+	/// <typeparam name="T">Specifies the type of elements in the pool.</typeparam>
+	public interface IPool<T>
+	{
```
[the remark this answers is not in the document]
davejrodriguez: While this PR _is_ using Factories in every Pool implementation, it is not necessary. Unlike Request and Return functionality, a pool could still pool without a factory. I think it would be a worse pool, but a pool nonetheless. So I'd prefer not to enforce factory usage from the interface.

UOP1_Project/Assets/Scripts/Pool/Example/ParticlePool.cs
```diff
@@ -0,0 +1,38 @@
...
+	public override IFactory<PoolableParticle> Factory
+	{
+		get
+		{
+			return _factory;
+		}
+		set
+		{
+			_factory = value as ParticleFactory;
+		}
```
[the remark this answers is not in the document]
davejrodriguez: This breaks CreateInstance. Though I'd like if you could expand on this: "add the setter to the derived class, but there's no need to enforce it from the base." If we can accomplish that without breaking CreateInstance, I'd be all for it.

UOP1_Project/ProjectSettings/ProjectVersion.txt
```diff
@@ -1,2 +1,2 @@
-m_EditorVersion: 2019.4.11f1
-m_EditorVersionWithRevision: 2019.4.11f1 (2d9804dddde7)
+m_EditorVersion: 2019.4.12f1
```
circa94, line 1: This is something you should not commit

UOP1_Project/ProjectSettings/InputManager.asset
```diff
@@ -4,196 +4,4 @@
...
-    altPositiveButton: 
-    gravity: 1000
-    dead: 0.001
-    sensitivity: 1000
-    snap: 0
-    invert: 0
-    type: 2
-    axis: 5
-    joyNum: 0
+  m_Axes: []
```
circa94, line 7: This was deleted by mistake?

davejrodriguez: Yeah, having a bit of a time with Git right now lol. Thought I got it reverted, but I guess not. I'll give it another try.

UOP1_Project/Assets/Scripts/Pool/Example/ParticlePool.cs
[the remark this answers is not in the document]
DeivSky: Well if you have a base class with a getter:
``` csharp
public abstract class BaseClass
{
    public abstract float Value { get; }
}
```
You can still add a setter to Value in any class that derives from it:
``` csharp
public class DerivedClass : BaseClass
{
    public float Value { get; set; }
}
```
So if you need a Factory setter for a specific pool, you can make that specific pool have it, and you'd be able to use CreateInstance on it.

UOP1_Project/Assets/Scripts/Pool/IPool.cs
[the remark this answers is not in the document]
DeivSky: I'd argue that that would simply be a Stack with different method names, but I guess you're right, there's no need to couple them together.

UOP1_Project/Assets/Scripts/Pool/Example/ParticlePool.cs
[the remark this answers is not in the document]
davejrodriguez: Hmm, that doesn't look right. In that code, you're shadowing (sans new keyword). And even if you do that, you still need to implement the proper override. And a proper override would error that the abstract `Value` does not have an overridable set accessor.

I might be missing something. Do you have this working? If so, please submit a pull and I'll take a look.

UOP1_Project/Assets/Scripts/Pool/IPool.cs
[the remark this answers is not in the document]
davejrodriguez: It doesn't even have to be a Stack. I just used the collection I thought was best for the job. Another Pool implementation might want individual index access and use a List. But that wouldn't be possible if I said all pools must use a Stack.

UOP1_Project/Assets/Scripts/Pool/Example/ParticlePool.cs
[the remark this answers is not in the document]
DeivSky: Huh, sorry, my bad, you're totally right! I mixed everything up with interfaces, I was so certain I didn't even check 🤦🏻.
``` csharp
public interface IInterface
{
	float Value { get; }
}
public class MyClass : IInterface
{
	public float Value { get; set; }
}
```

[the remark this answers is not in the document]
davejrodriguez: Ah, gotcha. No worries! I still don't think the factory belongs in the interface, but I'll leave this conversation open in case other reviewers would like to offer their opinion.

UOP1_Project/Assets/Scripts/Pool/Example/PoolTester.cs
```diff
@@ -0,0 +1,21 @@
...
+
+	private IEnumerator Start()
+	{
+		List<PoolableParticle> particles = _pool.Request(10) as List<PoolableParticle>;
+		foreach(PoolableParticle particle in particles)
+		{
+			particle.transform.position = Random.insideUnitSphere * 5f;
+			particle.Play();
+		}
+		yield return new WaitForSecondsRealtime(5f);
```
ciro-unity, line 18: Why using `WaitForSecondsRealtime` here? Since you're dealing with ParticleSystems, if you scale time they will go slowly, so you should used scaled time here as well (`WaitForSeconds`).

UOP1_Project/Assets/Scripts/Pool/Example/PoolableParticle.cs
```diff
@@ -0,0 +1,37 @@
...
+	public void OnReturn(Action onReturned)
+	{
+		StartCoroutine(DoReturn(onReturned));
+	}
+
+	IEnumerator DoReturn(Action onReturned)
+	{
+		if (_particleSystem.isPlaying)
+		{
+			yield return new WaitForSecondsRealtime(_particleSystem.main.duration - (_particleSystem.time % _particleSystem.main.duration));
```
ciro-unity, line 30: Same comment here, we should probably use `WaitForSeconds` rather than its -`Realtime` counterpart.

ciro-unity: Or maybe here it would be good to multiply it all by `Time.timeScale`, in addition to making the change above?

UOP1_Project/Assets/Scripts/Factory/Factory.cs
```diff
@@ -0,0 +1,16 @@
+﻿using UnityEngine;
+
+namespace OP1.Factory
```
ciro-unity, line 3: The namespace name should be `UOP1`, and so in all other classes.

UOP1_Project/Assets/Scripts/Pool/Example/Shared Global Particle Pool.asset.meta
```diff
@@ -0,0 +1,8 @@
+fileFormatVersion: 2
+guid: 73e856d683fe4d44aa05fc3a5fe882c0
+NativeFormatImporter:
+  externalObjects: {}
+  mainObjectFileID: 0
+  userData: 
+  assetBundleName: 
+  assetBundleVariant:
```
ciro-unity, line 8: Please, no spaces in filenames!

UOP1_Project/Assets/Scripts/Pool/Example/ParticleFactory.cs
```diff
@@ -0,0 +1,21 @@
+﻿using UnityEngine;
+using OP1.Factory;
+
+[CreateAssetMenu(fileName = "New Particle Factory", menuName = "Factory/Particle Factory")]
```
ciro-unity, line 4: No spaces in filenames

ciro-unity (commented):
I've done a review and added a few comments, but I'll fix them myself so we can go faster with the merge. It's a good solution for what I can see, we just need to use it in context to test it!

ciro-unity:
Thanks @davejrodriguez for the great PR, and @DeivSky and @circa94 for the code review!

I merged it. Now let's make sure it works with other systems. I will try to integrate its usage in one of the AudioSystems, as soon as I get to them.

UOP1_Project/Assets/Scripts/Pool/Example/PoolTester.cs
davejrodriguez: Yes, good point. My example was less about the specific particle system implementation and more to showcase how to use the pool. I have to admit I'm no particle system expert :) But I should have caught that.

davejrodriguez:
Thanks all for the reviews. Always good to have more eyes on the code.

The audio system (PR #90) is actually what inspired me to ask for/implement this feature. Not sure which one you were leaning toward, but IMO it's the only one that fully addresses the brief. I think the only things holding it back are:

1. Singleton pattern
2. Tight coupling between manager logic and pooling.

If that PR were modified to use this generic pooling solution and the singleton pattern replaced with a scriptable object implementation, I think it could be a great option.

elocutura:
> 
> 
> Thanks all for the reviews. Always good to have more eyes on the code.
> 
> The audio system (PR #90) is actually what inspired me to ask for/implement this feature. Not sure which one you were leaning toward, but IMO it's the only one that fully addresses the brief. I think the only things holding it back are:
> 
>     1. Singleton pattern
> 
>     2. Tight coupling between manager logic and pooling.
> 
> 
> If that PR were modified to use this generic pooling solution and the singleton pattern replaced with a scriptable object implementation, I think it could be a great option.

I'm sorry I haven't been able to work on my PR, this weekend I'll try to fix these couple of things and make use of your Object Pool, awesome PR

Issue #123, opened 2020-10-28
Character can climb the mountain with well timed jumps


GFX5000i:
**Short description**
Character can climb the mountain without sliding back down by landing the jump when the character is not moving (WASD keys are not being pressed)
The moment the character moves, it begins sliding.

**Expected behaviour**
The character should slide down when landing in steep slopes, even without moving.

**To Reproduce**
1. Jump towards the mountain and stop moving before touching the ground
2. Jump again and only move while in the air
3. Plant a flag on the top of the mountain 😆

**Notes**
- The Idle and JumpDescending states don't have a StartSliding transition.

### Reaching the mountain top
![Mountain Peak](https://user-images.githubusercontent.com/16827309/97508135-a1139480-195d-11eb-81cf-08fe041372e6.png)

### The FSM debug
![Mountain Console](https://user-images.githubusercontent.com/16827309/97508143-a7a20c00-195d-11eb-95ce-1bb574dec290.png)

davejrodriguez:
Sounds like a feature 🙂

PR #124, opened 2020-10-28
Allow switching from the Idle state to Sliding or JumpDescending

GFX5000i:
Issue: https://github.com/UnityTechnologies/open-project-1/issues/123

Changed the Idle state to support the transition to StartSliding and BeginJumpDescent.
This doubles as a fix and to support any future plans for moving or disappearing objects in the environment that the character can step on.

I haven't added a transition between JumpDescending and Sliding due to some strange behaviors when jumping across small, but steep, slopes.

To Test, check if the character is sliding in any steep slopes by following the steps in the issue.

Additionally:
- Add a dynamic platform in the environment (with collider and rigidbody) (suggestion: Cube(8))
- Enter Play Mode
- Move the character to step on the platform
- Activate the state machine debug option
- Rotate the platform on the X or Z axis until the character starts sliding
- Move or Delete the platform so the character can fall without moving


Dpagan1991 (changes requested):
review

ciro-unity:
Fix looks good, merging!

PR #128, opened 2020-10-29
Created Beach Skybox

HarshNarayanJha:
forum link: https://forum.unity.com/threads/skyboxes.987462/
I created and added a beach skybox using photoshop (just a beginning)
But I found a issue that top of the skybox looks weird. Please anybody can look upon it and modify it so that this weird thing can be solved?

I used photoshop for creating the panoramic skybox and imported into unity as a panoramic skybox material. I also used offset filter to create seamless clouds, but I can't figure out why the top part is looking weird. I also created six cubemap faces but the issue still exist.


ciro-unity:
Hey @HarshNarayanJha, in the end I decided to close the PR.
Your skybox is not bad, but I feel the style doesn't fit. Differently from code, when it comes to art we can only accept in the repo things that we think are in line with the project's art style, since they will show in all screenshot/videos/livestreams we do.

Thanks anyway!

HarshNarayanJha:
Ok, no problem. I will try with my other possibilities.

Issue #108, opened 2020-10-25
Baked geometry becomes black when no dynamic geometry is on screen


ciro-unity:
**Short description**
When using the toon shader, baked geometry becomes black when no dynamic geometry is present on screen.

**To Reproduce**
1. Pull the art-assets branch ([this commit](https://github.com/UnityTechnologies/open-project-1/commit/d2067b98aa10cee14a575c5a4e73176c4afcd1da), or more recent).
2. Open the ArtShowcase scene.
3. Make sure any object is marked as static (preferrably the ground too).
4. Bake lighting. Wait for Unity to finish.
5. Now look at the object, without having any non-baked object on screen.

**Notes**
- The issue seems to lie in how the custom toon shader treats shadows cast by dynamic objects on static geometry. When no object is in the view, the issue occurs as if the static geometry is now "all shadowed".
- The issue might be not in the shader, but in the pipeline or in the custom .hlsl code used in the shader.
- Video of the issue:

[![image](https://user-images.githubusercontent.com/20049224/97590534-01a6de00-19ff-11eb-9420-0af41805e934.png)](https://youtu.be/0HOBL3afYLM)

ciro-unity:
I already have a clue on why it happens. Will fix it as soon as I have time.

yopparai:
oops I created PR just right now

ciro-unity:
> oops I created PR just right now

It's ok, if your fix works I'll merge yours! :)

ciro-unity:
Fixed!

PR #129, opened 2020-10-29
Fix baked geometry becomes black when no dynamic geometry is on screen

yopparai:
Link: #108
There was missing `#else` condition
Now there are no black textures after light baking (look at reproducing issue section)

### DIFF
--- a/UOP1_Project/Assets/Shaders/CustomHLSL/CustomLighting.hlsl
+++ b/UOP1_Project/Assets/Shaders/CustomHLSL/CustomLighting.hlsl
@@ -18,13 +18,13 @@ void MainLight_float(float3 WorldPos, out float3 Direction, out float3 Color, ou
 
 	#if !defined(_MAIN_LIGHT_SHADOWS) || defined(_RECEIVE_SHADOWS_OFF)
 		ShadowAtten = 1.0h;
-	#endif
-
-	ShadowSamplingData shadowSamplingData = GetMainLightShadowSamplingData();
-	float shadowStrength = GetMainLightShadowStrength();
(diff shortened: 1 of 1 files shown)


UOP1_Project/.gitignore
```diff
@@ -23,6 +23,9 @@
 .vs/
 .vscode/
 
+# Idea cache directory
+.idea/
```
yopparai, line 27: I use [Rider](https://www.jetbrains.com/rider) so I dont want store `.idea` preferences folder

UOP1_Project/.editorconfig
```diff
@@ -1,4 +1,5 @@
 [*.cs]
+end_of_line = crlf
```
yopparai, line 2: Windows/Linux and VisualStudio/Rider have different line endings by default. So I installed `crlf` here.

UOP1_Project/.gitignore
ciro-unity: This doesn't belong to this PR though. Can you remove it and put it in a different one?

UOP1_Project/.editorconfig
ciro-unity: This doesn't belong to this PR though. Can you remove it and put it in a different one?
Also not super sure how the repo will react. Will all existing files be modified? I'm not an expert on this.

yopparai: Done

UOP1_Project/.gitignore
yopparai: Done

ciro-unity:
Fantastic!
Also, as a side effect, this fix improves the visualisation of Prefabs, giving proper lighting while in Prefab mode.

I'll merge it straight away.

Same as the other PR: if you are on the Unity Discord, I can give you the Open Projects Contributor role!

PR #131, opened 2020-10-30
Improved debugging for the state machine

GFX5000i:
[Generic State Machine - Post](https://forum.unity.com/threads/generic-state-machine.979371/page-4#post-6471092)

With this is possible to debug the details of how a state machine decides its state changes. The log is formatted to show:
- The GameObject of the state machine.
- The previous and new states.
- The name of the transition.
- Every condition evaluated by the transition and their results.
- If each result met the transition requirements.

A more in-depth debugger will prove invaluable when adding state machines to the game in more numbers and with increased complexity.
The logs will allow the developers and designers to pinpoint the source of any strange behaviors quicker.

**Main changes**
I've added the class StateMachineDebugger and a few small changes inside the core state machine classes:
- In StateMachine the debug variables were changed to a reference for the debugger class, while making sure the inspector options remained the same.
- Added references to the ScriptableObject of origin in StateCondition and StateTransition.
- Added a reference to the StateMachine in StateCondition.
- Added calls to the debugger in key points of the transition logic.
- Just like before, all calls to the debugger and its reference on the state machine are not carried over outside the editor.

**Note**
This will probably add many conflicts with @DeivSky's latest improvements on the state machine https://github.com/UnityTechnologies/open-project-1/pull/114 , might be better to merge his PR before this one.


GFX5000i:
Added to the log a list of the actions activated by the new State. This means StateActions now also holds a reference to their ScriptableObject of origin.

Implemented two extra options in the inspector to not log conditions or actions, avoiding excessive information when not needed.

DeivSky:
> Added to the log a list of the actions activated by the new State. This means StateActions now also holds a reference to their ScriptableObject of origin.

As you probably noticed, to debug the names of the states I injected the name of the SOs into them, that's why they have a Name property, we should either do that with actions and conditions as well, or change states to instead hold a reference to the SO. We could also have a mix of having them hold a reference to the SOs and a `public string Name => _so.name;`
Just to keep things consistent. I'll leave it up to you decide :)

As I said in the forums, if you want to make it compatible with the changes I made to transitions, just removing the reference to the SO inside the transition and the name of the transition from the log would suffice. It's kind of redundant for transitions to have names because they're usually something like StateAToStateB, and saying "Transitioning from StateA to StateB" is often more legible anyway.

GFX5000i:
> As you probably noticed, to debug the names of the states I injected the name of the SOs into them, that's why they have a Name property, we should either do that with actions and conditions as well, or change states to instead hold a reference to the SO. We could also have a mix of having them hold a reference to the SOs and a `public string Name => _so.name;`

The SO usually don't have any possibility of being destroyed before any of the FSM classes and a reference has a constant memory size, while a string will depend on the name size. With this in mind I think it's better to keep a reference to the SO than to create a copy of the name, might be a little bit lighter on memory and also doubles as some classes like actions and conditions might need to get configuration data from their respective SOs in the future.

For now I will change the State to hold the SO reference and remove the name property since it's only being used for the debugger. If the code gets more complex in the future then it might be better to add some properties.

I'll commit these and some extra changes tomorrow (including the documentation that I totally did not forgot to write 🙄)

ciro-unity:
Like for the TransitionTable stuff from @DeivSky, I've merged this PR into a new branch called `state-machine-improvements`... your new playground! :D

PR #135, opened 2020-10-31
Working dialogue and cutscene system.

rainaldisatria:
**Forum:**
https://forum.unity.com/threads/dialogue-system-timeline-track.980712/page-2

**Changes:**
- Improve and rename previous dialogue data (Coversation.cs to DialogueData.cs)
- Working cutscene system with trigger (showing dialogue box, disabling user input, advance the dialogue)
- Able to play DialogueData using timeline 
- Can move object (like camera) when waiting for user input.
- Will jump timeline to proper clip if player input is faster than the timeline.
- Will pause the timeline if user input is slower than the timeline 

**Demo:** 
You can play the demo in:
Scenes/CutsceneTest/Cutscene

There are 3 cutscene  that will automatically triggered + 2 manually triggered cutscene (talk with NPC):

or you can just watch it here: 
https://www.youtube.com/watch?v=sXAiXf3uiE8

**Note**
This is an improvement and a complete working generic Cutscene/Dialogue System inspired by the last 2 iteration of Dialogue/Cutscene System.

About the code:
To make it simple:
DialogueData consist of many DialogueLine and one timeline asset (you can leave timeline asset empty if the cutscene/dialogue doesn't use timeline, like a simple talk with NPCs).
DialogueLines consist of ActorData and sentence. 
**See Figure Below:**

image

Often, I found it hard to re write the actor name over and over again, so I decided to make reusable object of it. 
**See figure below.**
image

Then, there is CutsceneManager (Singleton) that hold reference to DialogueBox, CutsceneData, PlayableDirector, and etc. 
**See figure below**
image

CutsceneManager will get the CutsceneData from any controller in the scene (we call it CutsceneTrigger). 
CutsceneTrigger can be triggered using collider or automatic call on the start. 
**See figure below**
image
Note: That string field is an place holder for event (like changing scene)
That event should be triggered after DialogueData reached it's end. 
(We can have list of this CutsceneData)

Lastly, the custom timeline track for cutscene/dialogue.
image

Every dialogue clip tell timeline to show the dialogue box. (The control of user input itself still handled by the CutsceneManager). 
No dialogue clip mean = hide dialogue box.
Every clip has 'wait until' data. Rather than bind clip with dialogue id one by one, you just need to tell them the needed id.
Timeline will pause if current dialogue counter (user input) is less or equal than the wait time. 
Will Jump to end of the clip if current dialogue counter (user input) has become greater than the wait time.

You can still mimic 'bind clip one-by-one' by creating many dialogue clip with incrementing 'wait until'
image

I make video about the workflow:
https://www.youtube.com/watch?v=H3T3mm9Hq_s
 
btw,
I'm not really focusing on how to connect/trigger them, because I haven't check on how our game will be build up,
So rn, I use singleton DP (Its not a fixed design).

And because rn, we're using Singleton DP, one of the promised feature is still not delivered 
(able to see the dialogue in timeline preview mode).
I'll work on that later after we're not using Singleton DP. 

And also, I'm not exactly sure, how's our game dialogue will be, 
so I just make the generic one. So that we can improve this later. 

And... I'm not really familiar with the new input system, so let me know if I'm doing it the right way.

in the making, I have suggestion:
- Support multiple dialogue box (Conversation dialogue box, tutorial dialogue box, floating in the air dialogue box)


ciro-unity:
Great work @rainaldisatria, but a bit too complicated already and assuming a few things which we haven't defined. For instance, the character portraits. We don't have a plan to use them in the game.

Another thing I didn't particularly like is the decision to use `behaviour.WaitUntil` to control whether the timeline pauses or not. It's strange to have an int that needs to be manually set per clip, and is disconnected from the dialogue connected to the clip itself. Seems like a very good place to make lots of mistakes :)

So for now I can't accept this PR, but I'd like to keep it open as a reference because it contains a lot of good solutions too.

rainaldisatria:
Whoops, my bad. I make so many assumptions ehe

About the waitUntil, actually, I was inspired by last iteration (binding the clip with the id).
But, sure, we can have it in the DialogueData I guess, some boolean value.

Okay then, hope this'll help

ciro-unity:
Currently reviewing this and making a lot of changes, so nobody push changes to it, please.

ciro-unity:
More comments, as I refactor...

- The Cutscene system of the PR is controlling cutscenes, dialogues and UI for it. Too much. I'm splitting it into separate managers to allow for controlling and triggering dialogues which are not cutscenes.
- The system assumes that a cutscene is always a dialogue, and that an associated Timeline is only part of it, if present. We think it makes more sense to reverse it: cutscenes are Timelines, and _maybe_ they also contain a dialogue - in the form of a Dialogue track as specified on the task card.
- As the system was configured, it required assigning Timeline assets to the manager which didn't have a reference in the scene. This meant that you need to make sure to rebind all objects in the Timeline asset to the objects in the scene they are meant to control. A bit of a nightmare! Better to have the Timelines already in the scene, each in its own PlayableDirector component, and bind them there. This also allows previewing the Timelines, something which was impossible with your system without pressing Play and stepping into the triggers.
- The CutscenTrigger script was relying on `OnTriggerStay` to check if the user was interacting with the cutscene. Weird choice, what happens if the cutscene takes the player character out of the trigger?

ciro-unity:
Finally done! The refactoring was HUGE :D Took me a while, but I'm happy with it as a starting point for cutscenes.
Thanks for the PR though, it gave me a lot of ideas!

As usual, if you want to get the role of Open Project Contributor on the official Unity Discord, tell me the username you use there.

rainaldisatria:
@ciro-unity 
A HUGE refactor, indeed. :joy:
What a good news just after I finished my mid semester test.
Thanks for reviewing and refactoring my code, it gives me new perspective to implement it.

My Discord username is `sarai21#8040`

ciro-unity:
Happy to help :)
By the way, we're going to show it tomorrow during the livestream!

Issue #120, opened 2020-10-28
Spawn System not spawning the player when loading scene from the LocationLoader


Javimtib92:
**Short description**
Player doesn't get instantiated by SpawnSystem when Scene is loaded by the LocationLoader

**Expected behaviour**
When scene is loaded via the LocationLoader player should be spawned in the scene

**To Reproduce**
1. Go to LoadingTest/Beach Scene
2. Add the SpawnSystem and everything necessary to spawn the protagonist
3. Go to the SceneLoading Scene
4. Click Play
5. When the menu appears click Start
6. When Beach Scene is loaded check the screen to see if player was instantiated

**Notes**
- Maybe it's too early to create this issue, but I noticed this by playing a little bit with the LocationLoader, for the record a workaround is to go to the SpawnSystem class and Spawn on the Start method instead of the Awake, but maybe that is not a solid fix.

- Please @ciro-unity close this issue if you think it's too early to add it

ciro-unity:
It might be a tad early, but I'll keep it open as a reminder, why not!

HemangMaan:
Sir, I would like to contribute and fix this bug.

ciro-unity:
> Sir, I would like to contribute and fix this bug.

Sorry, there's already a PR which claims to be fixing it, I just linked it: #136

PR #136, opened 2020-10-31
Changed Awake() to Start() in SpawnSystem (#120)

treivize:
Issue #120 

When the scenes are additively loaded, GameObject instantiation should be put in Start(), because Awake() is exectuted before the new scene becomes active.

Reference thread: https://forum.unity.com/threads/additive-scene-loading-and-awake.660379/


ciro-unity:
Let's pull it in for now, then we'll figure a way to make the whole system more solid. Thanks @treivize!
Also, I gave you the Open Projects Contributor role - assuming you are @Treivize#1865 :D

PR #139, opened 2020-11-02
Added foam + refraction + depth fare to water shader

treivize:
Forum thread: [Water Shader](https://forum.unity.com/threads/water-shader.995380/)

This water shader is based on the following implementation:
[https://www.youtube.com/watch?v=MHdDUqJHJxM](https://www.youtube.com/watch?v=MHdDUqJHJxM)

It includes:
 - Water depth computation with shallow & deep water color gradient
 - Water reflection based on color distortion
 - Water foam based on Gradient noise
 - Scale adjustment to make the rendering not dependent of the water geometry scale

image

Result:
image


treivize:
The new result is the following:
image

The showcase update:
image

treivize:
Some improvements:
- Global effect fading for far water:
image

Result:
image

- Expose shadow controls (light probe contribution and Shadow intensity)
image

ciro-unity:
Really great work on this shader!
I have a few tweaks to ask you though, if you don't mind:

- One thing that would be great to fix is to make sure that planes put next to each other displace at the same rate, so that their edges match. Is that possible, or does it depend on geometry, size and position of the object?
You can see what I mean in the image below:

upload_2020-11-8_11-10-52

- I think it would be better to have only one displacement effect. Right now, the displacement is transparent and overlaid on top of the regular geometry render. This creates a ghosting effect that makes it a bit confusing.
In other words, even if the shader/material is in the transparent queue, it should completely covers the pixels of the existing geometry.

upload_2020-11-8_11-13-37

- And finally as mentioned before, it would be great to have sliders that limit the values within reasonable ones, so you can't break the shader involuntarily while tweaking it.

treivize:
Thanks a lot for the feedback and having spotted these issues.
I have been able to correct all of them.
- Water vertex displacement is based on World coordinate
- Ghost object behind the water level has been removed. Just the part rendered with the shallow water color change remains not affected by the refraction...
image
- Shader is full of slider now:
image

Finally, the water material settings are completely independent from the geometry scale!

ciro-unity:
Thanks for the super quick fixes!! 

> Finally, the water material settings are completely independent from the geometry scale!
That's also great!

There is still something I'd like to fix for the next iteration, which is:

- Right now the refraction is not doubled, but it still overlaps with the silhouette of what is not under the water:
image
I think the key is in that the alpha value of the colours is used to mask the refraction effect. I would need to investigate more on this...

- While fading away the refraction is a great idea to avoid pixelisation, I don't think we should fade away the wave movement and maybe not the foam either. But I also want to test this in a big scene before making this call!
image

For now, I want to merge this, then we can think of enhancements.

Thanks!!

PR #140, opened 2020-11-02
Created a simple Enemy System

HarshNarayanJha:
A simple Enemy System

Forum Link: https://forum.unity.com/threads/enemies.998704/
It's a enemy system in which I have sculpted a critter (not so good) and created a EnemyBase.cs . A Critter.cs sctipt extends EnemyBase and uses OnTriggerEnter, OnTriggerExit and OnColliderStay methods for detecting Player coming in detect range and moves towards it. If player collides with the enemy, a Health.cs script does TakeDamage method which decreases player's health by damageAmount. if health is 0 (or less than 0), player dies, then we can end the game or anything else.


ciro-unity:
Hey @HarshNarayanJha, two things:
- We already have a 3D model of the StoneCritter, you can find it in the `art-assets` branch.
- If you want to create enemy behaviour, you should use the state machine that we currently have in the game. However, there are also numerous PRs changing its behaviour, so I'd suggest to hold off for a little before using it. Or you can do it now, but be ready for some refactoring later!

For now, I'll put this PR on hold.

ciro-unity:
Closing this because it's not using the StateMachine at all.

PR #141, opened 2020-11-03
Input checks from Phase.Started => Phase.Performed

kgc00:
This PR changes input checks inside the InputReader.cs file from the InputActionPhase.Started phase to the Performed phase.

The input system provides the ability to define 'interactions' - different actions which fire off on the same button press depending on the length of the press (Hold, Multi-Tap, etc).

Actions are detected in the Started Phase, and considered complete in the Performed phase.  Currently we only look at the started phase, which will cause issues if we ever want to use the interaction system.

Should not affect any existing functionality, but just conform to best practice (all the official guides show performing callbacks from performed phase, not started phase).

More details here: https://forum.unity.com/threads/controls.980073/#post-6484114


ciro-unity:
Thanks, great that you spotted it. I'll merge it.
Are you on Discord? I'll grant you the role Open Projects Contributor if you give me the username!

PR #143, opened 2020-11-03
Added cached conditions to the State Machine

GFX5000i:
[Forum > Generic State Machine > Post](https://forum.unity.com/threads/generic-state-machine.979371/page-4#post-6480297)

With these changes the conditions will be cached on the first evaluation of every frame, this will allow the state machine transition logic to be more complex without increased processing in subsequent evaluations on the same frame.
This feature is active by default and can be changed independently for each instance of the condition Scriptable Object.
The caching is based on each instance of the state machine and of its respective SO.

There is a lot of changes on the code, but most of them are from refactoring:
- The abstract function Condition.Statement() changed from public to protected so it can only be used inside the internal Condition.GetStatement(). that is where the caching happens.
- Changed the createdInstances Dictionary from being a "nomad" variable to a readonly field inside the StateMachine. For better code readability.
- Both signatures of StateSO.GetState() unified in a single internal function. Since using the public method outside the FSM could lead to unwanted results.
- Added documentation regarding all Get methods for the abstract Scriptable Objects to clarify their behaviors.

For now this PR is marked as a draft since it depends on other PRs first. I might need to redo the code for a cleaner commit.
https://github.com/UnityTechnologies/open-project-1/pull/114
https://github.com/UnityTechnologies/open-project-1/pull/131

### DIFF
--- a/UOP1_Project/Assets/Prefabs/Pig.prefab
+++ b/UOP1_Project/Assets/Prefabs/Pig.prefab
@@ -941,8 +941,9 @@ MonoBehaviour:
   m_Script: {fileID: 11500000, guid: 1eeda163c70b3cb4ebb88ba92f608fd8, type: 3}
   m_Name: 
   m_EditorClassIdentifier: 
-  CurrentState: 
-  debug: 0
+  _debugger:
+    debugTransitions: 0
+    CurrentState: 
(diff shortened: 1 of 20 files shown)


GFX5000i:
cleaner version on PR https://github.com/UnityTechnologies/open-project-1/pull/164

PR #145, opened 2020-11-04
Beach whitebox

jamercer:
originally by @itsLevi0sa (on forums), optimized by @jamercer
https://forum.unity.com/threads/whiteboxing.987468/#post-6491652


amel-unity:
Hey, I swapped the missing tree prefab so now you can use it in the scene. I don't know what happened but the current position of almost all the trees is set to (0,0,0) so if you want to use the new prefab with the right initial positions used in @itsLevi0sa demo or I can merge the changes without the trees if you prefer.

jamercer:
okay. swapping the prefab seems to have confused unity and it zero'd out all the tree transforms when in the 'whiteboxing' branch,(but displayed transforms correctly when in the 'main' branch). i was not able to resolve this even by importing the original scene from the .unitypack. to get this to work i went into 'main' got all the trees, unpacked every tree prefab, saved that to a scene, brought it over to 'whiteboxing'
normally if you were to do this without unpacking the prefab it would assign the tree your swapped model but zero the transform, but since i unpacked it, it preserved the transform but was not able to swap the model. thankfully it was straightforward to manually reassign all the mesh data. so now the beach has trees, they arn't tree prefabs, but they're trees.

jamercer:
i should have imported the original tree prefab when i originally made this commit. not sure why i didn't. anyway.

amel-unity:
Looks great! Thanks a lot. I just made the trees prefabs so we can apply changes on them easier in the future :) I will merge the PR now.

PR #156, opened 2020-11-08
Added wind sensitiveness (green channel) on vertex color of nature FBX assets 

treivize:
Codecks cards:
- [Plants/Leaves wind effect](https://open.codecks.io/unity-open-project-1/decks/24/card/177-plants-leaves-wind-effect)
- [Shader add-on effects](https://open.codecks.io/unity-open-project-1/decks/61/card/17h-shader-add-on-effects)

Forum thread: [Toon Shader](https://forum.unity.com/threads/toon-shader.980874/)

The aim of this PR is to add the wind sensitivity on FBX assets by setting the green channel of their vertex color. Material switches to ToonPlusWind are also made.

image


ciro-unity:
Top contribution! Thanks again, @treivize!

ciro-unity:
Aaaaaaaarggghhhhh... I think this PR broke all of scaling for the nature objects! :(
They look the same size, but they are scale 0.01...

You need to be careful when you export .fbx, the unit in your 3D modeling software counts and will have effects on the scale once imported into Unity. I need to think of a solution... either I re-export the objects from Blender or something, or we might have to ditch the green channel for now and paint them from scratch.

ciro-unity:
This happened to
Tree
Palm
Bamboo
Bush

but not to
Fern
Grass
Flower

?

PR #158, opened 2020-11-10
Added an Editor assembly for the FSM editor scripts

GFX5000i:
This assembly will mark the editor code for removal when building the project since the Editor folder does not do this automatically inside another assembly.


ciro-unity:
I'm merging this without even verifying. Hope it works!!! :D

Ogungodspower5 (commented):
Sell

PR #162, opened 2020-11-10
Dialogue system event fix

kgc00:
https://github.com/UnityTechnologies/open-project-1/issues/163#issue-739547015

https://forum.unity.com/threads/dialogue-system-timeline-track.980712/page-2#post-6507001

Used named methods to subscribe and unsubscribe from events.  Also, took the opportunity to run this system through the input reader, rather than have it bypass the input reader.

1.) put a console log in some event subscription
2.) trigger multiple cutscenes
3.) see that the event was only fired the expected number of times (1)


ciro-unity:
I think these fixes make a lot of sense, I'll just make a couple of small tweaks (tabs instead of spaces?? 😨 ), and merge.

ciro-unity:
Ah nevermind, I know why you used the spaces. Let's leave them like that then!

PR #164, opened 2020-11-10
Added cached conditions to the State Machine v2

GFX5000i:
[Forum > Generic State Machine > Post](https://forum.unity.com/threads/generic-state-machine.979371/page-4#post-6480297)

With these changes the conditions will be cached on the first evaluation of every frame, this will allow the state machine transition logic to be more complex without increased processing in subsequent evaluations on the same frame.
This feature is active by default and can be changed independently for each instance of the condition Scriptable Object.
The caching is based on each instance of the state machine and of its respective SO.

There are a lot of changes on the code that come from refactoring and good practices:
- The abstract function Condition.Statement() changed from public to protected so it can only be used inside the internal Condition.GetStatement(). That is where the caching happens.
- Added documentation regarding all Get methods for the abstract Scriptable Objects to clarify their behaviors.
- Removed the script StateTransitionSO since it was replaced by TransitionTableSO
- Changed public TransitionTableSO.GetInitialState() to internal.
- Changed a few SO references to their specialized classes.
- Disabled the debugger on Pig.prefab so it doesn't spam the console.


ciro-unity:
Merging on the trust :D 
Let's finish this state machine!!

PR #170, opened 2020-11-11
Earring for Townsfolk

Smurjo:
Added an earring for townsfolk to make them different from one another.

It was build with Probuilder, exported to obj for some finishing touches with Blender, then saved as obj again. The material is a simple color.

The earring possibly needs to be placed somewhere else once the townfolk is rigged. Which begs the question whether they can move their ears...


ciro-unity:
Thanks for the PR!
A couple of things:
- The earring was waaaay too high poly. 768 triangles. Consider that an ENTIRE Townsfolk is 5576. So I reduced the polygons and reimported, as .fbx, to 120. I might have exaggerated a bit, but the visual difference is really not that much.
- The Prefab was an empty GameObject with another object inside. I think it's due to how you exported it from the DCC tool. For such a simple decoration, one object is enough, so I simplified the hierarchy.
- I renamed the folder, in general don't use spaces in folders and filenames.

PR #172, opened 2020-11-11
Added fire shader and campfire

treivize:
Forum thread: [Fire effect](https://forum.unity.com/threads/fire-effect.1001794/)
Codecks card: [Fire card](https://open.codecks.io/unity-open-project-1/decks/24/card/158-fire)

Pot over fire update:
image

Fire Shader graph:
image

Fire Material:
image


ciro-unity:
The solution you used is very interesting, and probably quite cheap.

I think it does lose a bit of performance in that it's using a particle system to make it work. You use it only for the billboarding, right? Isn't it possible to do that in shader too? (not 100% sure about this)

In any case, I will approve the PR. **But**, I still think we can improve the fire by having a real hand-drawn flipbook, and as I said previously I'd love to measure myself with it later if I have time. This might mean that later we'll remove this shader... hope you don't mind!

treivize:
Indeed, it was for the built-in billboarding feature. 
It might be possible with a shader, with vertex displacement to move the quad and face the camera...

No problem to replace it later, I might try myself to do it, and learn along the path. 
One thing that I feel not realistic is if the fire is moving (like on the critter head): the fire should trail a bit I guess. it is something not emulate with the current shader

ciro-unity:
Yeah, that's why for next iteration I'd like to have proper particles. But we can do it in a bit, there are so many other things to do :D

Issue #178, opened 2020-11-12
Dialogue Line not shown in custom Timeline track and also not paused at the end of the clip.


rainaldisatria:
Sometimes Dialogue Line is not showed by Timeline if dialogue clip is sit side by side.
The skipped clip doesn't wait for user input even if PauseWhenClipEnds set to true.

**How to reproduce:**
1. Go to CutsceneExample scene (in Assets/Scene/Examples/)
2. Select CutsceneExample game object in hierarchy
3. In the Timeline Window, Move SecondLine next to FirstLine
4. Play
5. Boom, Second Line is not printed in console.

**Expected Behaviour:**
SecondLine is printed in Console.

Or just watch it here:
https://youtu.be/vfSMV7tkgh8

rainaldisatria:
So, after doing simple debugging, the problem is:
If the clip sit side by side, OnBehaviourPlay is called at the end of previous clip (not at the start of that specific clip).
Since OnBehaviourPlay() is called only once, that causes the second dialogue is not called (or called but still not ready yet).

![image](https://user-images.githubusercontent.com/57592497/98943333-e7522180-2521-11eb-84c4-bd37e8a598a9.png)

I fixed the problem by changing OnBehaviourPlay() with ProcessFrame(), 
I'll make the PR ASAP.

rainaldisatria:
Bug fixed. Closing the issue.

PR #180, opened 2020-11-12
Fix/dialogue line not shown in custom timeline track

rainaldisatria:
Issue: https://github.com/UnityTechnologies/open-project-1/issues/178

I change OnBehaviourPlay() to ProcessFrame() in DialogueBehaviour.cs.

OnBehaviourPlay() is called only once as soon as the playhead reached the clip. 
But the thing is, when two clips are side by side, the OnBehaviourPlay() of the second clip is also called at the end of the first clip.
that causes the second clip is called when it's still not ready (IsCutscenePlaying is still set to false). 

ProcessFrame() is called every frame.
We're checking if the clip is ready to shown every frame, but displaying the dialogue line is called only once.

I make sample scene in Assets/Scenes/Examples/Cutscenes.unity to make sure it's working.


ciro-unity:
Great catch, @rainaldisatria, and I see it now: since Timeline is a tool that doesn't work on discreet increments, the OnBehaviourPause of the first clip is only called **after** and not **when** the clip is over. So for instance if the clip ends at 00:30.150, it will be called at 00:30.152 or something.
So in that sense yes, we're already in the second clip, so it makes sense that the OnBehaviourPlay is called.

It makes a lot of sense now.

I was very much against using OnProcessFrame() since I didn't want to do the check every frame, but I guess a simple if(bool) is a light enough check to do - especially if it's only once a frame. (i.e. only one dialogue can play at the time).

I'll merge now.

ciro-unity:
You're also the first completed item in 0.4 :)
Screenshot 2020-11-12 at 20 00 40

PR #181, opened 2020-11-12
Remove unused ReorderableList implementation

neon-age:
This script was added with Dialogue system #135 and was unused.

It has no namespace, so when I was writing custom editor it used this script instead of UnityEditorInternal.ReorderableList.

Also it has a different API that is actually harder to use, and no real benefits over internal version.


rainaldisatria:
We changed how dialogue is implemented at times.
That's why there's some unused script.

Also, don't forget to delete this folder Dialogues/Editor/ 
It's not used either.

With that, all set.

Maybe also change the title of this PR to "clean up unused code"

ciro-unity:
Thanks @rainaldisatria for the hint to the other script.

Merging!

PR #183, opened 2020-11-13
Added glass shader and applied it to lantern

treivize:
Forum: [Toon shader](https://forum.unity.com/threads/toon-shader.980874/)
Codecks card: [Transparency Support](https://open.codecks.io/unity-open-project-1/decks/16/card/174-transparency-support)

The solution proposed to support transparency is to add a new transparent Unlit Shader and a new Material called both "Glass".
The shader is inspired by the refraction trick done on water shader to make the inner content deformed.

Glass_shader
Glass_Material

The lantern mesh has been split in two parts: glass part and metallic caps, to be able to apply 2 distinct materials.

Lantern_prefab

Result:
Glass_Result


ciro-unity:
I don't think we need to split the mesh in two, we just need to assign two materials to the same mesh. After all, you will never see the glass on its own or the two ends without the glass in the middle.

Also not sure about refraction. The Phoenix Chick is an important character, and have it refracted at a lower resolution is a bit of a big loss. Plus, how thick is that glass? It probably isn't, so the refraction is probably not necessary.

PR #186, opened 2020-11-13
Added Transparent support on ToonShading and Glass Shader for the lan…

treivize:
Forum: [Toon Shader](https://forum.unity.com/threads/toon-shader.980874/)
Codecks card: [Transparency Support](https://open.codecks.io/unity-open-project-1/decks/16-3d-shading/card/174-transparency-support)

Sorry about this PR, because first it is a new one and not an update of the previous one and because it is covering 2 aspects of the transparency support:
- An update of the M=main Toon shader transparency cutoff feature that is driven by the alpha channel of the color map texture.
- A new shader/material: Glass for the particular case of the lantern. The lantern fbx has been udpated to assign 2 materials to the mesh.

Both results together:
image

Glass shader:
image

ToonShading:
image

Note: I have also remove a useless input from ToonSpecular which was a mistake from the shader refactor from me before...


ciro-unity (changes requested):
A couple of things need to be fixed before I can accept this:
- The mesh is now rotated, and requires to be in 90, 0, 0. This needs to be fixed
- Size has also broken, somehow. The model is 0.01 scale, needs to be 1

Can you fix the model? Thanks!

image

treivize:
Sorry about that! It should be fixed now (FBX and Lantern prefab to include the glass material)

image

ciro-unity (approved):

ciro-unity:
No problem! Thanks for the quick fix. I merged it.

PR #188, opened 2020-11-14
Simple menu controls

kgc00:
https://forum.unity.com/threads/controls.980073/page-2#post-6507162

brings base functionality for navigating and interacting with menu UI.

This solution will need to be expanded upon to handle tab views.

Players should be able to interact with ui elements to progress through the game.

How did you implement them?  
#Overview
MenuInput class acts as manager for menu input logic
- it consumes events provided by the InputReader class
- adds custom selection and component caching behaviour 
- must be present on the same transform as the SelectableUI components it controls

SelectableUIELement send messages to the MenuInput when a mouse is over them or selection has been moved onto them
- they have events for mouse enter and exit
- they have events for on selection
- they find the MenuInput component by checking the root transform for children with the MenuInput class (can be refactored later)
- does not handle submit events (must use CC_Button if you'd like submit behaviour)

CC_Button (chop chop button) implements the same methods as SelectableUIElement, however it also inherits from UnityEngine.UI.Button implements an OnSubmitMethod which queries whether it should call base.OnSubmit() or stop the interaction early.  Used for our custom menu input handling to prevent an edgecase where mouse clicks can call submit on the wrong UI element.

Scenario goes as follows:
- mouse is over button one
- keyboard moves selection to button two
- user hits left mouse button without moving the mouse
- - without custom handling, system would submit on button two, which is not good.  custom handling stops the submit action before it hits the button's onClick list.

Also included are some debugging scripts / assets like DebugMenu.cs and some debug menu prefab in Prefabs/Menu


ciro-unity (changes requested):
Great submission, but there are a couple of things I'd like to change (mentioned in the comments).

Also, it had me confused FOR A WHILE, since the whole behaviour of the menu was relying on the `DebugMenu.cs` executing `OpenMenu()`, which calls `_inputReader.EnableMenuInput( )`. Only then the menu would work!!

BUT

The thing is, the built-in `InputSystemUIInputModule` script on the "EventSystem" GameObject also has a reference to an Input Action Asset, so it was also reading the input... but not all of it! That was very confusing. I think we should disconnect that value, like this:
image

At this point, the above mentioned script is completely controlled by your manual event relay. Otherwise, I fear there would be conflicts of events (?)

The other thing I don't fully understand is when the mouse is hovering on a Button and you press the keyboard, leaving the hovered button in a Selected state (green). Why so? Shouldn't be, right? It should be un-highlighted (normal state).

kgc00:
Hi @ciro-unity, thanks for checking this out!  ...As you may have been able to tell from my activity on the forums I was unsure of my approach and requirements for this feature.  I'm really glad you took a look at it, and that was honestly the main point of doing a PR at this state.  

One thing I do want to apologize for, is that after reading your comments, I realized I had done a poor job of documenting my work & intention.  I will definitely try to be more explicit next time =).  Ok, onto my response:

> Also, it had me confused FOR A WHILE, since the whole behaviour of the menu was relying on the `DebugMenu.cs` executing `OpenMenu()`, which calls `_inputReader.EnableMenuInput( )`

Yes! (sorry for that, I'm sure it was frustrating to review).  

My thought was that I didn't want 'create' the actual menu, just some temporary way to test input...  since my PR was just to handle the menu input.  I was thinking we could create a proper menu including opening and closing later on in another PR. I don't think I communicated that clearly, so that is definitely a mistake on my part.

If you are fine with it I can rework the names / functionality to make the DebugMenu the actual pause menu, and we can swap out the UI elements when needed.  (It sounds like this is what you are asking for)

> BUTThe thing is, the built-in `InputSystemUIInputModule` script on the "EventSystem" GameObject also has a reference to an Input Action Asset, so it was also reading the input... but not all of it! That was very confusing. I think we should disconnect that value, like this

Yes, I saw that you could assign an `InputActions` asset to the `InputSystemUIInputModule`...  I actually thought it was required. 
 I'm surprised that it still receives events from the ChopChop `InputActions` asset, even when the field in the inspector is cleared, but it does, and I agree that this is the correct way to go.

Also, whether the `Action Asset` field is assigned or not on the Event System Gameobject, the **event system** **is reading input independently** of our `MenuInput` & `InputReader`.  This is just how the UI + Event System architecture works.  For example:
    - go into the input reader and comment out all the method contents.  You will still see the selection logic work (according to the vanilla UI navigation rules).
    - Vice versa, turn off the event system, and keep our input reader's method's intact.  You will see we don't get any kind of UI selection/highlighting functionality.
This is what made it so hard to implement the custom logic, we could not control the event system's native calls... which makes sense.  I just had to do some reading to figure out the correct approach, e.g. extending button.

> Otherwise, I fear there would be conflicts of events (?)

from my testing there is not a conflict of events (i had logs in all the receivers and they were being fired as they should, buy i was also worried).  However, i agree 100% with your suggestion of removing the actions asset, and think it is the better approach.  No reason to have redundancy and obfuscation in our codebase.

> The other thing I don't fully understand is when the mouse is hovering on a Button and you press the keyboard, leaving the hovered button in a Selected state (green)

you should not be seeing that behaviour, i cannot recreate it on my end.  

I will post an updated video into the thread which contains the logic as it is now in my envirnoment and you can see if that meets your standards.  If you were messing around with the event system & input reader (which it sounds like you were) that could have caused this. 

Also, I will double check on my end that the git PR is up to date (could be a prefab config thing)

> I think having a special treatment for the Menu input vs. Gameplay and Dialogues is REALLY confusing.

> I'd incorporate Menu input in here with all its callbacks, for now. If it grows further we can do some kind of refactoring, but it has to be the same for all 3 Action Maps.

I agree 100% that this is really confusing.  Sorry about that, Ciro.  I should have kept with our current architecture for the PR, then submitted an additional 'cleaning' PR to separate out these changes.  The cleaning pr would have this applied to **all** action maps, not just the menu.

Here's my thinking on why I did it:

It is already the case that the number of actions in the `InputReader` is already too big, and each action map should be split to it's own class which should be managed by the `InputReader`.  I ran into naming conflicts when I did this PR in `InputReader`, for example:
        - Menu & Dialogue both have a `MoveSelection` action.  When the `InputReader` implements both `IDialogueActions` and `IMenuActions` there is a conflict, because both require a method called `OnMoveSelection`.
        - This script still compiles, but now, inside the implementation of `OnMoveSelection` we have to somehow tell whether we should fire the Menu event or Dialogue event.  IMO, this is a code smell and means that `InputReader` has too many responsibilities currently.
    - An alternative solution is to keep all the interface implementations in the `InputReader` and change the action maps and event names to reflect this
        - instead `Dialogue` `ActionMap` having a `MoveSelection` it would have a `MoveDialogueSelection`.
        - Then `InputReader`'s impelmentation would be:
            - the `IDialogueActions` interface would have an `OnMoveDialogueSelection` method, and the event we fire would be a `MoveDialogueSelectionEvent`
            - the `IMenuActions` interface would have an `OnMoveMenuSelection` method, and the event we fire would be a `MoveMenuSelectionEvent`.
        - No more conflict.
But, I thought it was cleaner to separate these out, and have a class subscribe to some `InputReader.Menu.MoveSelectionEvent` or `InputReader.Dialogue.MoveSelectionEvent`.

Also, would you prefer to (once all issues have been addressed) merge the PR in, or keep it on a feature branch and work out the tab view logic?

kgc00:
@ciro-unity  I've made the requested changes and confirmed that the functionality is working according to the requirements.

ciro-unity:
Argh, sorry, I was looking at the "Changes Requested" label, not realising you already did change stuff. I'll review it shortly, maybe immediately after the livestream so we don't mess up stuff in other areas (we want to show something which includes input - somehow).

kgc00:
> Argh, sorry, I was looking at the "Changes Requested" label, not realising you already did change stuff. I'll review it shortly, maybe immediately after the livestream so we don't mess up stuff in other areas (we want to show something which includes input - somehow).

hi ciro thanks for checking back in.  I was doing some other stuff so no worries =)

ciro-unity:
Ah ok, cool. We're in the usual Livestream panic today but I'll merge this soon enough :D

kgc00:
Sounds good! Thanks for all your work 🙂

ciro-unity:
Unbelievable but true, I finally managed to merge this. Sorry it took SO long. It's actually a simple enough merge, but I am still in my head thinking how many unforeseen problems will pursuing a menu for both mouse and keyboard/joystick bring us.
Also had an unexplicable merging issue which took me 2 hours to solve (not your fault) which I could only solve by literally copy-pasting lines inside the InputAction asset. 🤷🏻‍♂️ 

Anyway, it's in now!

kgc00:
> Unbelievable but true, I finally managed to merge this. Sorry it took SO long. 
No worries at all!  I have to say, I really don't envy being on the other end of this process... It must be super overwhelming to manage so many PR's =.=

Thanks for all your work (and sorry about the conflict)!

> I am still in my head thinking how many unforeseen problems will pursuing a menu for both mouse and keyboard/joystick bring us.

I do think the new input system will help us avoid this scenario... but they don't call it 'unforseen' for nothing =D

PR #192, opened 2020-11-16
Jumping, Landing and Walking Puff of Dust Particles

erizzoalbuquerque:
A PR containing puff of dust particles when the main character jumps, lands, and walks.

card: https://open.codecks.io/unity-open-project-1/decks/24-effects-particles/card/13s-puff-of-dust
fórum: https://forum.unity.com/threads/puff-of-dust-effects.1002936/#post-6521759

I created 3 particle systems for when the main character jumps, lands and walks based on Mario Odyssey's reference (see the forum post for the reference).

This is the result I got:

Updated Particles

I created 3d model for the particle so it can receive light and cast shadows as seen in Mario Odyssey. I also, created a material using the toon shader previously created in this project.
I can still polish it. Any feedback?

I also, created a temp particle controller script so people can test it easily. I intend to delete if the PR is about to be accepted.

All scripts and assets I created can be found in a folder named ErizzoalbuquerqueStuff. I will move/delete the files when/if the PR is about to be accepted.


erizzoalbuquerque:
So, is everything ok? I didn't get a chance to place each asset in the right folder before the PR was closed.

PR #194, opened 2020-11-17
Feature/game settings

Casey-Hofland:
Game Settings makes it easy to create setting and edit UI without writing additional logic! Check out the OptionsMenu - Preview scene for this.

As a test, try adding a Shadow Quality Dropdown to the scene and have it display the "Hard Only" option.

Support for URP and the Input System is coming.

- [https://forum.unity.com/threads/game-settings-system.980703/](forum)
- A Game Settings system that allows for easy creation of settings and coupled UI. I made that preview scene in 5 minutes and it can be changed in 5 minutes.
- The game will need an options menu. Furthermore, this menu should be easy to iterate on and not require the help of a developer.
- Game Settings are all Scriptable Objects with Save Load functionality and a property to the actual setting (e.g. QualitySettings.antiAliasing). These are then used by a SettingsDropdown or other in tandem with a SettingInterpreter. This approach allows GameSettings to be completely decoupled from the UI (in fact, if a project doesn't even import the ugui package, Game Settings won't throw any errors.)

Now: I know I have broken the rule of not adding any packages to the project, but Game Settings comes as a [https://github.com/Casey-Hofland/GameSettings](package). What I 'did' do was to make it Local to the project, so it should still be editable by anyone. I can of course also put it into the project as well. The only reason I haven't done so is... I would argue that this is cleaner. Having it as a package signals to the general nature of Game Settings and allows me to update the package directly.

### DIFF
--- a/UOP1_Project/Assets/Scenes/Menus/OptionsMenu - Preview.unity
+++ b/UOP1_Project/Assets/Scenes/Menus/OptionsMenu - Preview.unity
@@ -1,0 +1,6356 @@
+%YAML 1.1
+%TAG !u! tag:unity3d.com,2011:
+--- !u!29 &1
+OcclusionCullingSettings:
+  m_ObjectHideFlags: 0
+  serializedVersion: 2
+  m_OcclusionBakeSettings:
+    smallestOccluder: 5
+    smallestHole: 0.25
+    backfaceThreshold: 100
+  m_SceneGUID: 00000000000000000000000000000000
+  m_OcclusionCullingData: {fileID: 0}
+--- !u!104 &2
+RenderSettings:
+  m_ObjectHideFlags: 0
+  serializedVersion: 9
+  m_Fog: 0
+  m_FogColor: {r: 0.5, g: 0.5, b: 0.5, a: 1}
+  m_FogMode: 3
+  m_FogDensity: 0.01
+  m_LinearFogStart: 0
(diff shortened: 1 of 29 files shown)


STARasGAMES:
It looks like you've missed the actual code in your commit. I can see only one .cs file in your pull request

Casey-Hofland:
The code is inside this [package](https://github.com/Casey-Hofland/GameSettings).

davejrodriguez:
> The code is inside this [package](https://github.com/Casey-Hofland/GameSettings).

The code should not be in a package. It makes it impossible to do a proper code review. And how are people supposed to collaborate on this if it's a package referencing your own repo?

Casey-Hofland:
Thinking about it... people are actually not supposed to collaborate on the package. The package is supposed to be an all-purpose tool and it would be impractical to explain how the architecture fits together.

The thing is that I haven't thought this through, because I kind of want to feel out Ciro's and the forums response. Now that I'm evaluating the situation: no, I would not like people to contribute to this package, only on contributing its use in Open Projects.

Sorry if this is all kind of clumsy, I realize this whole thing has kind of become a push for my own agenda. I'd still like to explore it though because I believe the package is great, and I'll make a few tutorials on how to use it in the near future.

ciro-unity:
I'm just reading this. No worries, man, don't overthink it! :)

PR #198, opened 2020-11-18
Sombrero for townsfolk

Smurjo:
Sombrero for townsfolk to make them look more unique - suitable for one-sided rendering.

´These are the steps I did:

- I started with creating the texture in GIMP. 

- Then I created a cylinder in Blender, from which I deleted the top and bottom.

- I created UV on the cylinder in Blender with cylinder projection

- Then I created a material on the cylinder with texture and switched to textured view in Blender

- Using scaling and transformations, I shaped a the Cylinder into the rough shape of the sombrero

- Then I exported the townfolk from the project and imported it to Blender

- With scaling, transformation and rotations I shaped the sombrero fitting to the townsfolks head (which wasn't too easy given the ears and and cloth he already wears, I decided to have the cloth poke-through as if it was part of the sombrero)

- I also gave the Sombrero brim a slightly asymetric bend to make it look more natural

- Then, as we are not to build for double-sided redering:

- I mirrored the texture, 

- realigned the UV's in Blender to the modified texture, 

- duplicated the brim,
-moved the UV of the duplicated part to the mirrored part of the texture
- scaled the brim up minimally and moved it down a little, 
- flipped the UV and connected the 2 sides of the brim.

Sombrero

I know the triangle-count is 1472, but they are of comparable size as on the townfolk model - about 30-40% of the triangles were caused by having to make the brim doublesided.


ciro-unity:
> * I also gave the Sombrero brim a slightly asymetric bend to make it look more natural

Great idea!

ciro-unity:
Hey @Smurjo, thanks for the submission!
One thing: the prefab is great, but we don't need a new scene for every item. This way the scene count would skyrocket, and the project would become a mess. I'll remove it this time!

ciro-unity:
Actually, after looking at the model, I need to ask you for a few changes.
- Can you reduce the polygons a bit?
- At the same time, the cap has a hole inside. I would say we should cap it, so we can use the cap in another situation. Or they can take it off. You can put way less polygons inside.
- There is a lot of waste on the texture, which is currently a 2048x2048. Can we reduce it to 512?

Smurjo:
I will create new one..

Issue #193, opened 2020-11-17
State Machine flickers to Sliding if character walks into small objects


ciro-unity:
**Short description**
State Machine flickers to Sliding if character walks into a small object.

**Expected behaviour**
The character should stay in the Walking state.

**To Reproduce**
1. Open a testing scene, press Play
2. Walk to an object with a small collider (like the vase), or a very small step.
3. Keep pressing the direction towards the object.
4. Observe the character seems to jitter, while the state in the State Machine alternates between Sliding and Walking.

**Notes**
- Look at State in the gif below

[![Image from Gyazo](https://i.gyazo.com/928a1e40edbda03756b9c22a11d60ea1.gif)](https://gyazo.com/928a1e40edbda03756b9c22a11d60ea1)

mtakac:
At first I thought we could add some condition to the `IsSlidingConditionSO` that would prevent the player entering the sliding state if the collider of the body he collided with is less than certain height.

But after more testing, I think this might actually be connected to #167. The problem is that the Pottery object has a sphere collider and you're "climbing" and sliding from this sphere. What seems to be worse, is that you can still climb the pottery even if it's much larger (that's without any jumping, I am just holding down a movement key):

![Screen Recording 2020-11-19 at 04 17 33](https://user-images.githubusercontent.com/13066543/99583342-4df1a680-2a1e-11eb-8cd7-7f7dc54f98b2.gif)

I think limiting the slope that player can climb and maybe even matching it with the slope that's needed for player to slide would be a good solution.

ciro-unity:
Ah, if it's a matter of collider, it might be even better to assign it a different one (for instance a Capsule?). Again, no huge deal for now, but let's keep an eye on this one.

mtakac:
The problem from my previous comment seems to be fixed now. After I pulled a new version today, I can no longer reproduce it. I am not sure what fixed it, but now, if the pottery is larger it just blocks the player's movement and he's not able to walk over it, nor is there any flickering between walking and sliding states.

Regarding the original issue: yes, changing the collider would solve it, although maybe not Capsule, but rather Box collider. The thing is, Capsule collider still has a slope on its top so if it small enough it'll cause the same issue. Since Box collider has flat top, you either step through it if it's small enough or it will block your movement without changing to sliding state. In this case, the pottery objects are large enough to block the movement when using Box collider.

[...]

PR #201, opened 2020-11-19
Changed Pottery game object collider to box

mtakac:
This PR solves #193. See the discussion there for more details.

Before:

before

After:

after3


ciro-unity:
Great, thanks! I'll merge it, but I think we should keep Issue #193 open since other small objects would produce the issue anyway (?)

mtakac:
@ciro-unity yes, you're right. We'll have the same issue again if we use small sphere or capsule collider for some other objects, so it makes sense to keep the issue open. 

I think we can just wait until we have more objects and then decide whether it's okay to just use box colliders for small objects or if we need some more robust solution that would require updating the code.

sultanshb (approved):

PR #202, opened 2020-11-20
Add Explicit Prewarm to Pool

davejrodriguez:
Forum: https://forum.unity.com/threads/generic-object-pool.990776/#post-6541984

Added Prewarm to IPool
Implemented Prewarm in PoolSO
Removed private InitializePool in ComponentPoolSO
Added Lazy-Loaded PoolRootObject to ComponentPoolSO
Removed InitialPoolSize property from ComponentPoolSO
Modified Examples to use new workflow
Modified Audio scripts to use new workflow

### DIFF
--- a/UOP1_Project/Assets/Scripts/Pool/PoolSO.cs
+++ b/UOP1_Project/Assets/Scripts/Pool/PoolSO.cs
@@ -10,7 +10,7 @@ namespace UOP1.Pool
 	/// <typeparam name="T">Specifies the type of elements to pool.</typeparam>
 	public abstract class PoolSO<T> : ScriptableObject, IPool<T>
 	{
-		protected readonly Stack<T> _available = new Stack<T>();
+		protected readonly Stack<T> Available = new Stack<T>();
 		public abstract IFactory<T> Factory { get; set; }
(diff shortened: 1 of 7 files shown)


UOP1_Project/Assets/Scripts/Pool/PoolSO.cs
```diff
@@ -10,17 +10,32 @@ namespace UOP1.Pool
...
 			return Factory.Create();
 		}
 
+		public virtual void Prewarm(int num)
+		{
+			if (HasBeenPrewarmed)
+			{
+				Debug.LogWarning($"Pool {name} has already been prewarmed.");
+				return;
+			}
```
davejrodriguez, line 28: So I made the decision to only allow prewarming of a pool once. However, this doesn't _need_ to be. We could just allow this method to be called and add however many objects to the pool whenever. In that case, it may make sense to call the method `Warm`.

ciro-unity:
Thanks Dave, I merged it. However there are a couple of things about the pool, I'll write them [in the thread](https://forum.unity.com/threads/generic-object-pool.990776/)!

PR #203, opened 2020-11-21
Editor window multiple event trigger fix

MorDeCaii:
The handler method for `listView.onSelectionChanged` assignes every time the window got focus. The event triggers multiple times, so we need to clean it up on refresh.


ciro-unity:
Thanks @MorDeCaii! The fix seems to work, so I merged it. I am not 100% sure what is it fixing, but I'll trust you on this!

Tagging @DeivSky for awareness, since he made the custom editor.

DeivSky:
Good spotting!

PR #205, opened 2020-11-21
Created new HandTorch item

ciro-unity:
Added prefab and model, reused old material. The Prefab is not in the art scene yet.

Hope you like it.


ciro-unity:
Still making changes to this PR, bear with me!

PR #208, opened 2020-11-22
NatureAssets

Llaczky:
NatureAssets

3D Nature Assets assets, Mostly plants, and a Big House for review.


ciro-unity:
Suuuper cool contribution!!

I do however have a few points of feedback:
- As usual, keep an eye on the polygons. Some assets are ok, but sometimes I've spotted A LOT of polygons in very small things. Case in point: the centre of the flowers on the one called "FlowerBush"/"CloudBush", or the "BeachPalm". There might be others.
image

- The colour for the trees and the bush is... radioactive! They are **really** bright compared to the rest of the assets. Can they be toned down? Also, keep in mind that they all seem to have the Light Probe Contribution value set to 5 on the Material. It should be 3.5 by default.
image

- Many assets are imported with weird rotations or scaling. Often -90 on the X axis. Can that be fixed? What software do you use? If it's Blender, [I made a small guide that might help](https://www.youtube.com/watch?v=24DwOUWrKIg&list=PLZplUm29-Z-xOYY9Tw6t3tSIvlbhVhzUn&index=2).
image

- For the scaling: I feel like the trees are quite small. Specifically "FlatTree" "SwirlTree" and "MushroomTree". They could all probably be twice their size.

- The "FlowerBush" is using 3 materials! Can you condense them into one? We have mask tinting functionality (in a specific shader), so we can tint the flowers withouth needing 3 materials on one object (which triplicates the draw calls).

Lots of feedback, I know! Let me know if there is something you can't fix, and maybe I can help or we can offload it to someone else from the community. Again, thanks a lot for these great assets!

PR #210, opened 2020-11-23
Curls for female townsfolk and low quality sombrero

Smurjo:
Curls: I started with making the texture in GIMP, then I modelled the curls in Blender. I took a few screenshots while modelling:
- Modelling a single strand of hair
modelling curls 4
- Rolling it into a curl
modelling curls 7
- Duplicating and modifying (several times)
modelling curls 11
- Arranging curls in the hair tie
modelling curls 22
- joining curls in the hair tie - like branches of a tree
modelling curls 25
- adding hair that has been pulled together on the head
modelling curls 32
- adapting to townsfolks head and final tweaks, especially scaling and rotating curls slightly to make the curls more unique and to give the impression the curls are affected by gravity (albeit the curls largely defy gravity)
modelling curls 35

Sombrero: I added inside and reduced a lot of triangles

Both textures were optimized for smaller texture size and the UV remapped with overlapping UV


ciro-unity:
Thanks, merging! Just had to fix the sombrero's pivot, and there was a circle of vertices at the top which had some overlap, so I merged them.
👍🏻

Issue #206, opened 2020-11-21
Objects are black when the scene lighting is not baked


ciro-unity:
**Short description**
All objects (both dynamic and static) are black when dragged into a new scene.

**Expected behaviour**
Object should "work" automatically when dragged into a new scene.

**To Reproduce**
1. Create a new scene.
2. Drag any object from the `/Prefabs` folder (e.g. a character).
3. Observe that the object is black.

**Notes**
- Generating lighting, even if not the Baked GI, fixes the object. The object is probably becoming black due to missing ambient light information.

![image](https://user-images.githubusercontent.com/20049224/99889410-128ced00-2c55-11eb-8c1b-ccce0ce3d4d4.png)

Tatranskymedved:
I've just downloaded fresh repo, on Main branch I cannot reproduce this behaviour.

![image](https://user-images.githubusercontent.com/2912979/99910321-909cd280-2ced-11eb-9018-5cace8842715.png)

Tatranskymedved:
I'm not that experienced in Unity so I'm trying my best by test-failure errors, I'm warning that my founding might not be clear/make sense :wink: :

For dynamic objects: we're using Skinned Mesh Renderer which doesn't support lightmap (they will be always dark without direct light?) - https://forum.unity.com/threads/problem-with-skinned-mesh-renderer.366293/

However I've played around with static objects, I've found that when I've set constant `color` in shader to value `1` (Shaders/CustomHLSL/CustomLighting.hlsl), objects get proper color. I assume that the light at the position of that object has wrong color (black?) and thus it doesn't enlighten it.

I'd like to follow up on fix, but I would need some more hint what can be wrong & what to check. Thanks.

ciro-unity:
Hey @Tatranskymedved! Welcome to the project :)
Have you followed the exact steps?

For clarification, when I say dynamic objects I mean not static. But they don't need to be `SkinnedMeshRenderer`s. Both `SkinnedMeshRenderer` and `MeshRenderer` create the issue.

Tatranskymedved:
Hi @ciro-unity , I've added PR, once you have little time, could You please review it & confirm if we want to do it this way and if we want light on all objects or just the static/with Baked GI? Thanks!

treivize:
@Tatranskymedved, regarding this issue, I think that the smallest impact needed to fix it without having to use branch should be:

[...]

PR #213, opened 2020-11-24
Bugfix - Updated "Light and Shadows" subgraph for 0 light, added default

Tatranskymedved:
Updated "Light and Shadows" subgraph to check whether there is some light and if none, supply full light.

Card: https://open.codecks.io/unity-open-project-1/decks/13-bugs/card/193-objects-are-black-when-the-scene-is-not-baked
Issue: https://github.com/UnityTechnologies/open-project-1/issues/206

**Short description of issue**
All objects (both dynamic and static) are black when dragged into a new scene.

**Resolution**
Shader sub-graphs are currently taking light from Baked GI and Main Light color. If they are missing, there is no light on the scene and characters/objects are dark. This PR is adding step in the sub-graph of "Lights and Shadows" to check whether there is full black (0,0,0) and if so, add full light on the character/object.

image
image


Tatranskymedved:
To clarify, I was originally thinking to add this light value only to the Baked GI, however it would not produce light on the dynamic objects/characters. I assume we want to have them lightened all up thus added the light at the end of sub-graph.

treivize:
Why not moving your any + branch just after the Baked GI node? Because here you loose completely the directional light information.  And maybe reduce also the fake illumination contribution (1 -> 0.5) to not over saturate the final rendering.

Tatranskymedved:
> Not sure why it not working to move your any + branch just after the Baked GI node. Because here you loose completely the directional light information.
> 
> And maybe reduce also the fake illumination contribution (1 -> 0.5) to not over saturate the final rendering.

If I would move supply light (Any+Branch with Color(1)) just after Baked GI, dynamic objects and character (that are not static) would have 0 light and would be completely black.

I agree that I can limit the light provide just to 0.5, but this value should get to actual result only when there is complete 0 of main light/baked light. Will that every occur in game scenes?

treivize:
I thought you were maybe on Discord, but I have checkout your PR on my local and move the any+branch close to the Baked GI node. And it is working fine for static and dynamic objects on a new scene with a directional light: 
image
image
which other problem do you try to solve exactly?

Tatranskymedved:
As mentioned on Discord, if we put this Any+Branching pair just after Baked GI, dynamic objects (which are not Baked) will have 0 light. Once you remove Directional light, it will get completely dark.

Should there always be some Main light / Baked GI available? E.g. if we enter building or some cave? If not then the objects might get completely dark and it would make sense to move this Branch pair just after Baked GI. But if this will practically never occur, this solution shouldn't cause any troubles/prevent any light information loss.

So Q is: "What are we trying to achieve with this?". Afaik it is only that we want to see objects when there is no light.

ciro-unity:
Thanks a lot for the PR @Tatranskymedved!

As a first thing, I'd say that we don't need the fix for baked objects, because if they are baked, they work. If light is not baked, then they are dynamicly lit. So we need to fix it only for dynamic objects.

Note that here by static/dynamic I don't just mean whether the Static flag is on/off, I mean if the object has participated in a baking process. Obviously participating in the bake is done with the above checkmark, but until you press "Bake", all object are dynamicly lit regardless of their static flags.

ciro-unity:
The other thing is, I've been told that the Branch node is one to avoid, if you can. When the compiler is able to, it is transformed in a math operation. For instance if it can translate the `false` as `*0` and the `true` as `*1`, but if it can't, it will insert an `if` which apparently is very underperforming for shaders. So any fix we choose, I'd avoid the branch node.

Do you want to fix it or do you want me to do it?

Tatranskymedved:
Hey @ciro-unity ,

I hope I've understood this correctly. We just need to add some fake bakedGI if there is none. Doing this, it should add some kind of light/color for all objects and the rest of lightning will be done by the rest of light (main lights etc.). And as your last comment mentioned to avoid branching, I've changed it to use `Maximum` node. 

updatedBakedGI

treivize:
This problem remembered me a video I have seen few months ago: [Branchless Programming](https://www.youtube.com/watch?v=bVJ-mWWL7cE)
Based on this approach, I would go for this way of swapping Baged GI to a default global lightning contribution:

image

ciro-unity:
I assume that's the way, I can't be 100% sure if we don't have it in the graph and test it. If you'd want to make the change then we can see if it works.

ciro-unity:
Hey @Tatranskymedved, a million apologies for coming back to this so late. In the end there is a very similar PR from @treivize which I am going to pull in since it's a very simple solution - which is always preferable.

However, I've done a bit of Git gymnastics so I could pull your commit too, in order to show you as a contributor too. I think you deserve it, the discussion happening here was very interesting!
Hope to see you in another PR :)

PR #215, opened 2020-11-25
[Object Pool] Removed DontDestroyOnLoad & Add SetParent

davejrodriguez:
Removed DDOL
Added SetParent method that allows pool managers to tie the pool lifetime to their own

**Forum thread:** https://forum.unity.com/threads/generic-object-pool.990776/#post-6548521

_Please see LocalPoolTester.cs for implementation changes._ 

- DontDestroyOnLoad has been removed from ComponentPoolSO. Pools now do not survive scene changes by default.
- Added SetParent method to ComponentPoolSO. ~Objects in use do not immediately get re-parented, because: 1. The pool does not keep track of objects in use. 2. Part of their use may include being parented to a requester. However, re-parenting happens as soon as they are returned to the pool.~


davejrodriguez:
You know, as I'm drifting off to sleep, I'm realizing I complicated some of the parenting stuff for no reason. Pool objects can always be a child of the root object and I can just parent the pool root to whatever via SetParent. No looping through members re-parent each necessary.

Bah! I'll put together a fix tomorrow.

ciro-unity:
Thanks Dave, merging!

PR #222, opened 2020-11-27
Hollow Boat Effect

erizzoalbuquerque:
card: https://open.codecks.io/unity-open-project-1/decks/16-3d-shading/card/18x-hollow-boat-effect
forum: https://forum.unity.com/threads/hollow-boat-effect.1012723/#post-6565384

Effect for making the boat hollow when it's over the water.

image

The development details can be found in the forum post.

ATTENTION: 

There's no water prefab right now so it's important to remember that Water has to be on the Water Layer for this effect to work. (I didn't want to update the Artshowcase scene nor creating a prefab for Water without Unity team permission)

Also, as this effect works, the boat can "never sunk" (the water is never drawn inside the boat, even if it is totally underwater). However, I can't imagine a reason for having the boat totally sunken in the game. If this is needed, we can use the depth buffer instead of the stencil buffer. It should be very simple.


ciro-unity:
> There's no water prefab right now so it's important to remember that Water has to be on the Water Layer for this effect to work. (I didn't want to update the Artshowcase scene nor creating a prefab for Water without Unity team permission)

That's perfect!

I merged it :)

PR #225, opened 2020-11-28
hair color mask material, old lady hairdo and glass, moustache, and s…

Smurjo:
…hell bowl

Various art assets to make the townsfolk more unique. Bowl made from a giant shell.

I recolored the hair texture in red, blue and green for use with detail mask shader. This way the same hairdo can be done in different color.

For the new models I followed the same process I usually do:
- create texture with GIMP
- in Blender put texture on basic mesh, in case of the shell e.g. a plane
- create triangles with cut and subdivide, with a dense mesh where much is required and a wider mesh where there isn't much detail, here to be seen on the mesh of the shell:
grafik
- then shaping the mesh
- as last step setting the pivot point and transform location to (0,0,0). Usually I put the pivot point in the center of mass, but for the shell on the bottom, as it is supposed to sit on some surface.
- export to fbx, import into Unity
- create material with toon shader and put texture in

old lady hairdo: 471 triangles, no texture, reusing detail mask from curls (512x512)
spectacle frame: 222 triangles, no texture so far
spectacle lens: 96 triangles, no texture
mustache: 142 triangles, no texture, reusing detail mask from curls (512x512)
shell bowl: 466 triangles, texture 512x512


ciro-unity:
Thanks @Smurjo! A couple of notes for next PR:
- Try to avoid spaces in file names and paths. They can be annoying sometimes and make some scripts fail in very unexpected ways... better to avoid them altogether.
- If you can follow the naming conventions, with Capitalised names, camel case, etc.
- Thanks for making the prefab as a variant!
- Prefabs go in the `/Prefabs` subfolder, not in `/Art`
- Note that the hairstyle .fbx is currently scaled at 100, and the base rotation is wrong (it's -90 on the x).

I fixed most of these except the rotation/scaling (can't open Blender right now), so for the next one just keep an eye on these!

Thanks!

PR #227, opened 2020-11-28
State Machine - Proposed fix for Issue 228

jandd661:
Added Sliding to JumpAscending because player was unable to jump if sliding which caused the player to get stuck in geometry/holes.

https://drive.google.com/file/d/17W4yQk7EHbmIGfpJtbvGedsDUlmHogpp/view?usp=sharing

Issue: #228

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -2,4 +2,3 @@
 
 # Visual Studio cache directory
 .vs/
-*.meta
--- a/.gitignore
+++ b/.gitignore
@@ -2,3 +2,4 @@
 
 # Visual Studio cache directory
 .vs/
+*.meta
--- a/UOP1_Project/Assets/ScriptableObjects/StateMachine/Protagonist/Conditions/IsActuallyMoving.asset
+++ b/UOP1_Project/Assets/ScriptableObjects/StateMachine/Protagonist/Conditions/IsActuallyMoving.asset
@@ -9,7 +9,7 @@ MonoBehaviour:
   m_GameObject: {fileID: 0}
   m_Enabled: 1
   m_EditorHideFlags: 0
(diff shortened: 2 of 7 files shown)


ciro-unity:
I am aware of that spot in the map which makes the player stuck, and yes, this technically fixes it.
But it also kind of defeats the purpose of sliding? In the beginning of the project we created the sliding state as a way to prevent the player from walking up very steep geometry, but if they can jump while sliding now they can jump upstream (like a salmon :D).

So I'm reluctant to "fix" this. Maybe one way to fix it is to validate that, even if sliding, the player is not moving. Then you could allow a jump (or just put it in the Idle state). But then you would have to patch the Idle state to make sure it doesn't start sliding again in the following frame 🤔 
I think it's a bit of a rabbit hole.

Can't we just think geometry so that we avoid these edge cases, like the one which is in that scene now? We just need to make sure we don't have steep pits like that.

Thoughts?

ciro-unity:
Hey @jandd661, thanks for this fix! I'm just slightly worried now that with all these patches and checks, the StateMachine is hiding some nasty bugs... not in the code, but in the logic. But what can we do!
Luckily we're not trying to make a precision platformer!

Merging... and thanks!

ciro-unity:
Ah, small thing: when you compare a value with the `sqrMagnitude`, you should square the value too. I changed it in the comparison with the `_threshold`, so that the value that the designer sees in the Inspector is the one that is actually being compared against, since both sides of the equations are now squared.

jandd661:
> Ah, small thing: when you compare a value with the `sqrMagnitude`, you should square the value too. 

Good point. Thank you

PR #235, opened 2020-11-29
Added German Localization

Allstreamer:
- Added Files to Start Translating Into German

I added a new Local for German and translated the existing Text.

I will maintain the German-Translation as the story continues to develop.


ciro-unity:
Hey, thanks but for now we'd rather not include other languages. We're still ironing out a  few things with UI, dialogues and whatnot, and we'll add all languages together later on.

For now, I'll just label this as On Hold (or you can close it and reopen it later if you fancy).

Issue #229, opened 2020-11-28
Location Loader issue


gabagpereira:
**Short description**
I have spotted an issue with the Location Loader where, at some point, an error will cause two locations to be stacked, which results in critical bugs, like having two instances of the Player at the same time.

**Expected behaviour**
New scenes/locations must be able to load seamlessly, without causing errors. Therefore, previous scenes must be unloaded first.

**To Reproduce**
1. Start on the **Beach** scene.
2. Exit to **Forest**.
3. Exit to **Beach**.
4. Keep repeating several times, until an "Index out of range" exception is thrown and you'll have two different locations loaded at the same time!

**Notes**
- This is a repost of from the [Unity forums thread](https://forum.unity.com/threads/location-loader-issue.1012369/).
- Possible fix: I've noticed that the LocationLoader script loads new scenes before unloading the existing ones, which makes me wonder: shouldn't this happen reversely? I mean, it makes more sense to unload the existing scenes first, before setting the new ones to load.

**Screenshot**
![issue](https://user-images.githubusercontent.com/72309552/100515838-d686ea00-3176-11eb-9968-74349042b05d.png)

ErikBehar:
So this is happening because the two exit locations ( forest to beach & beach to forest) happen to be on top of each other. If you move one of the two so that its not overlapping then you wont have this issue. 

The loading order is random, since we do both load and unload Asynch. 

So this is likely to trigger in the case where the new scene loads before the older scene is unloaded, thus the old pig triggers the enter in the new scene. This could also happen the other way if the spawn point happened to be on top of where an old trigger is in the other scene. 

Overall this could probably be prevented by:
a) destroying the character prior to doing any load/unload  ( this could also prevent other issues regarding SO's surviving past scene load, between the two character loads ) 
b) making sure exit / enters don't overlap at spawn or at exit 
c)  adding a global cool down to exit locations, so that they can't be triggered so fast ( a few seconds should be enough?)
d) add more defensive code, like checking if the pig that triggered is in the active scene

ErikBehar:
tested option D) 👍 
add 
 to line 19 of LocationExit :  && other.gameObject.scene == SceneManager.GetActiveScene() 

seems to work fine

gabagpereira:
> So this is happening because the two exit locations ( forest to beach & beach to forest) happen to be on top of each other. If you move one of the two so that its not overlapping then you wont have this issue.
> 
> The loading order is random, since we do both load and unload Asynch.
[...]

PR #236, opened 2020-11-29
Proposed fix for issue 229 Location loader

jandd661:
Issue: #229 

Fixed issue with it not unloading the previous scene correctly/reliably. Added some TODOs for future expansion. Also provided some test scenes to load additively and update the _Beach_ scene _Scene Loader_ to load those scenes additively. Also provided the associated LocationSOs for those scenes.

Video: https://youtu.be/VQWs_yalU6o


ciro-unity:
Hey @jandd661, a couple of things:
- As soon as I checkout your branch, I have 7 meta files which get created by Unity.
image
I think it's because you haven't saved the assets in Unity before pushing, so the .meta files are not part of the PR. Always make sure to hit `File > Save Project` in the top menu before you commit for a PR.
- The LocationExit script in the Beach scene has a bunch of null references, which throws an error when testing it. Always make sure to test your solution as a very last thing before committing.
image
- I removed a couple of comments which were really superfluous.

But I'll merge now!

PR #237, opened 2020-11-29
Issue #229 : prevent old pig from triggering anything in new scene

ErikBehar:
Issue #229 

1) We check that the pig triggering the exit exists in the same scene as the trigger, this way we prevent old pig or new pig triggering exits in mid asynch load states.
2) To prevent old pig maybe taking damage or picking up items or any other effects, I destroy the old pig before loading.

This can be tested by loading from beach to forest to beach multiple times with out exception.

### DIFF
--- a/UOP1_Project/Assets/Scripts/SceneManagement/LocationExit.cs
+++ b/UOP1_Project/Assets/Scripts/SceneManagement/LocationExit.cs
@@ -1,4 +1,5 @@
 ﻿using UnityEngine;
+using UnityEngine.SceneManagement;
 
 /// <summary>
 /// This class goes on a trigger which, when entered, sends the player to another Location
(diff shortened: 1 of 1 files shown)


khush2o2o:
I have an idea to slove this bug

ciro-unity:
> I have an idea to slove this bug
Hey @khush2o2o, there's 3 solutions already (this, #236 and #238) and I'm about to merge one of them, so no new PR needed for now! :)

ciro-unity:
Hey @ErikBehar, I'll reject this one because in the end I chose to implement @jandd661's implementation. Also, I like the check in the if, but I don't like the fact that the trigger destroys the player. Shouldn't be his role!
Hope you don't mind!

ErikBehar:
I agree that maybe its not the triggers role to delete the player, but we should delete the old player to make sure it doesnt trigger other things... I'll guess I can split this into two PR's the extra check and another where the trigger sends out an event and some other system with proper responsibility destroys the player.

khush2o2o:
I have the same solution that he tells already...
Sorry....

PR #239, opened 2020-11-29
Fix Plant Critter mesh scaling and name

treivize:
Just a small change in plant critter which was not exported with scale 1,1,1 and change Cube to PlantCritter for the mesh name at the same time ;)


ciro-unity:
Blind-merging! :D

PR #240, opened 2020-11-29
[Shaders] Relocate fresnel properties & remove albedo influence on specular lights

daneobyrd:
### **This PR now addresses two problems.**
_Edit: I have created an issue for this PR found [here](https://github.com/UnityTechnologies/open-project-1/issues/246)._

First is relocating some properties out of the main subgraph, and second is reordering some operations to remove albedo influence on specular lighting.

### **1) Relocating Fresnel properties out of the main ToonShading subgraph**

@ciro-unity brought up this issue (QoL fix) in the main Toon Shader [forum thread](https://forum.unity.com/threads/toon-shader.980874/#post-6571615).

Currently the Albedo operation is in the main ToonShading subgraph. This would be fine normally however, the Toon_Fresnel shader variant needs to access the Albedo output to control **Fresnel Color**'s contribution using the **Fresnel Power** property.

This left **Albedo Post Hook Color** and **Albedo Post Hook Contribution** or unused/unset on the main subgraph in all other shader variants.
image

image

I solve this moving Albedo calculations from ToonShading to a new subgraph within the main Toon shadergraph. I then created the new subgraph Albedo_Fresnel, containin the Albedo subgraph and the fresnel lerp operation needed for the slimes. The Albedo subgraph has been added to all toon shadergraph variants.

image
image

This [eliminates](https://forum.unity.com/threads/toon-shader.980874/#post-6572002) any unused or unset inputs on the base ToonShading subgraph.

### **2) Albedo output prevents dark surfaces from having receiving bright specular light.**

The issue is first raised [here](https://forum.unity.com/threads/toon-shader.980874/#post-6571858).
Discussion then identifies the multiply node responsible for the albedo influence on specular brightness.
My solution is [here](https://forum.unity.com/threads/toon-shader.980874/page-2#post-6579487) and [here](https://forum.unity.com/threads/toon-shader.980874/page-2#post-6574750).

image
image
image


ciro-unity:
Thanks @daneobyrd!
One small thing: let's try to put one fix per PR, or things might get mixed up!

ciro-unity:
In the end, I made a couple of changes. One is to actually "explode" the Albedo_Fresnel Subgraph inside the Toon_Fresnel shader, like this:

image

First, because it really didn't make sense to have a Subgraph used only in one place, to contain 2 nodes. It creates more complexity than the one it hides. Plus, the inputs called "Albedo Post Hook Color" and "Albedo Post Hook Contribution" were really esoteric... what do they mean? So I would never really take that Subgraph and use it anywhere else. Finally it wasn't even doing a fresnel operation, the fresnel was done outside. It was more like "Albedo_Lerp" :D
Plus, we got a nice animated preview of the effect now in the main graph!

ciro-unity:
Another thing: the "Albedo" subgraph and the "Texture and Tinting" subgraph were now doing the same thing, so I removed the second by "exploding" it into the Albedo one.

ciro-unity:
There's more: I removed the alpha from inside the ToonShading subgraph. Before, it looked like this: first, the alpha was being split and discarded in the Albedo subgraph:
AlbedoSubgraph

But then, the alpha (which is now only from the colour texture) was injected into the ToonShading graph, were more nodes were first splitting it, then recombining it...
ToonShading

And the only purpose for that was because it was being used for the alpha clipping:
Overall

But it would have never really have to come out of the shading Subgraph, since it doesn't take part in the shading.

So I removed most of those nodes, and now bring the alpha directly from the Albedo subgraph to the alpha clipping into the Master Node. I also made the input and output of the ToonShading a Vector3 and not 4, for clarity (and renamed them):
NewMainToon

The ToonShading is also simpler as a result:
NewToonShadingSubgraph

Actually all of the above is more interesting for @treivize, I believe, since he was probably the one who made the alpha clipping thing in the first place (?)

daneobyrd:
Haha, thanks for all this cleanup. On the albedo subgraph change, I have no idea why that wasn't my first thought. Making an extra subgraph that is only used once makes no sense. And yes, I kept the alpha split within the subgraph to preserve what @treivize had done but I hadn't really thought about whether it should have stayed there. 👍

These are all great adjustments and apologies about including multiple issues in one PR. At first I tried making a second PR but then saw that my commits were added here after I pushed to my fork.

ciro-unity:
No problem!

PR #241, opened 2020-11-30
Added initial plant critter behavior

treivize:
Forum thread: [Plant Critter](https://forum.unity.com/threads/plant-critter.1005700/)
Codeck card: [Plant Critter Integration](https://open.codecks.io/unity-open-project-1/decks/15-code/card/186-animation-integration)

image

### DIFF
--- a/UOP1_Project/Assets/Scripts/Characters/StateMachine/Actions/FaceProtagonistSO.cs
+++ b/UOP1_Project/Assets/Scripts/Characters/StateMachine/Actions/FaceProtagonistSO.cs
@@ -1,0 +1,39 @@
+﻿using UnityEngine;
+using UOP1.StateMachine;
+using UOP1.StateMachine.ScriptableObjects;
+
+[CreateAssetMenu(fileName = "FaceProtagonist", menuName = "State Machines/Actions/Face Protagonist")]
+public class FaceProtagonistSO : StateActionSO
+{
+    public TransformAnchor playerAnchor;
+	protected override StateAction CreateAction() => new FaceProtagonist();
+}
+
+public class FaceProtagonist : StateAction
+{
+	TransformAnchor _protagonist;
+	Transform _actor;
+
+	public override void Awake(StateMachine stateMachine)
+	{
+		_actor = stateMachine.transform;
+		_protagonist = ((FaceProtagonistSO)OriginSO).playerAnchor;
+	}
+		
+	public override void OnUpdate()
+	{
+		if (_protagonist.isSet)
+		{
+			Vector3 relativePos = _protagonist.Transform.position - _actor.position;
+			relativePos.y = 0f; // Force rotation to be only on Y axis.
+
+			Quaternion rotation = Quaternion.LookRotation(relativePos);
+			_actor.rotation = rotation;
+		}
+	}
+	
+	public override void OnStateEnter()
+	{
+		
+	}
+}
--- a/UOP1_Project/Assets/Prefabs/Characters/PlantCritter.prefab
+++ b/UOP1_Project/Assets/Prefabs/Characters/PlantCritter.prefab
@@ -1,5 +1,167 @@
 %YAML 1.1
 %TAG !u! tag:unity3d.com,2011:
+--- !u!1 &460632930
+GameObject:
+  m_ObjectHideFlags: 0
+  m_CorrespondingSourceObject: {fileID: 0}
+  m_PrefabInstance: {fileID: 0}
+  m_PrefabAsset: {fileID: 0}
+  serializedVersion: 6
+  m_Component:
+  - component: {fileID: 460632931}
+  - component: {fileID: 460632933}
+  - component: {fileID: 460632932}
+  m_Layer: 0
+  m_Name: AlertZoneCollider
+  m_TagString: Untagged
+  m_Icon: {fileID: 0}
+  m_NavMeshLayer: 0
+  m_StaticEditorFlags: 0
+  m_IsActive: 1
+--- !u!4 &460632931
+Transform:
+  m_ObjectHideFlags: 0
+  m_CorrespondingSourceObject: {fileID: 0}
+  m_PrefabInstance: {fileID: 0}
+  m_PrefabAsset: {fileID: 0}
+  m_GameObject: {fileID: 460632930}
+  m_LocalRotation: {x: 0, y: 0, z: 0, w: 1}
+  m_LocalPosition: {x: 0, y: 0, z: 0}
+  m_LocalScale: {x: 1, y: 1, z: 1}
+  m_Children: []
+  m_Father: {fileID: 8221149608678237085}
+  m_RootOrder: 2
+  m_LocalEulerAnglesHint: {x: 0, y: 0, z: 0}
+--- !u!135 &460632933
+SphereCollider:
+  m_ObjectHideFlags: 0
+  m_CorrespondingSourceObject: {fileID: 0}
+  m_PrefabInstance: {fileID: 0}
+  m_PrefabAsset: {fileID: 0}
+  m_GameObject: {fileID: 460632930}
+  m_Material: {fileID: 0}
+  m_IsTrigger: 1
+  m_Enabled: 1
+  serializedVersion: 2
+  m_Radius: 2
+  m_Center: {x: 0, y: 0, z: 0}
+--- !u!114 &460632932
+MonoBehaviour:
+  m_ObjectHideFlags: 0
+  m_CorrespondingSourceObject: {fileID: 0}
+  m_PrefabInstance: {fileID: 0}
+  m_PrefabAsset: {fileID: 0}
+  m_GameObject: {fileID: 460632930}
+  m_Enabled: 1
+  m_EditorHideFlags: 0
+  m_Script: {fileID: 11500000, guid: 7eb40f37003e762489325a71829af192, type: 3}
+  m_Name: 
+  m_EditorClassIdentifier: 
+  _enterZone:
+    m_PersistentCalls:
+      m_Calls:
+      - m_Target: {fileID: 443910769}
+        m_MethodName: set_IsPlayerInAlertZone
+        m_Mode: 0
+        m_Arguments:
+          m_ObjectArgument: {fileID: 0}
+          m_ObjectArgumentAssemblyTypeName: UnityEngine.Object, UnityEngine
+          m_IntArgument: 0
+          m_FloatArgument: 0
+          m_StringArgument: 
+          m_BoolArgument: 0
+        m_CallState: 2
+  _layers:
+    serializedVersion: 2
+    m_Bits: 512
+--- !u!1 &575482847
+GameObject:
+  m_ObjectHideFlags: 0
+  m_CorrespondingSourceObject: {fileID: 0}
+  m_PrefabInstance: {fileID: 0}
+  m_PrefabAsset: {fileID: 0}
+  serializedVersion: 6
+  m_Component:
+  - component: {fileID: 575482848}
+  - component: {fileID: 575482850}
+  - component: {fileID: 575482849}
+  m_Layer: 0
+  m_Name: AttackZoneCollider
+  m_TagString: Untagged
+  m_Icon: {fileID: 0}
+  m_NavMeshLayer: 0
+  m_StaticEditorFlags: 0
+  m_IsActive: 1
+--- !u!4 &575482848
+Transform:
+  m_ObjectHideFlags: 0
+  m_CorrespondingSourceObject: {fileID: 0}
+  m_PrefabInstance: {fileID: 0}
+  m_PrefabAsset: {fileID: 0}
+  m_GameObject: {fileID: 575482847}
+  m_LocalRotation: {x: 0, y: 0, z: 0, w: 1}
+  m_LocalPosition: {x: 0, y: 0, z: 0}
+  m_LocalScale: {x: 1, y: 1, z: 1}
+  m_Children: []
+  m_Father: {fileID: 8221149608678237085}
+  m_RootOrder: 3
+  m_LocalEulerAnglesHint: {x: 0, y: 0, z: 0}
+--- !u!135 &575482850
+SphereCollider:
+  m_ObjectHideFlags: 0
+  m_CorrespondingSourceObject: {fileID: 0}
+  m_PrefabInstance: {fileID: 0}
+  m_PrefabAsset: {fileID: 0}
+  m_GameObject: {fileID: 575482847}
+  m_Material: {fileID: 0}
+  m_IsTrigger: 1
+  m_Enabled: 1
+  serializedVersion: 2
+  m_Radius: 1
+  m_Center: {x: 0, y: 0, z: 0}
+--- !u!114 &575482849
+MonoBehaviour:
+  m_ObjectHideFlags: 0
+  m_CorrespondingSourceObject: {fileID: 0}
+  m_PrefabInstance: {fileID: 0}
+  m_PrefabAsset: {fileID: 0}
+  m_GameObject: {fileID: 575482847}
+  m_Enabled: 1
+  m_EditorHideFlags: 0
+  m_Script: {fileID: 11500000, guid: 7eb40f37003e762489325a71829af192, type: 3}
+  m_Name: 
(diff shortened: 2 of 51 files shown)


UOP1_Project/Assets/Scripts/Characters/StateMachine/Actions/FaceProtagonistSO.cs
```diff
@@ -0,0 +1,37 @@
...
+	{
+		if (_protagonist != null)
+		{
+			_actor.LookAt(_protagonist);
+		}
+	}
+	
+	public override void OnStateEnter()
+	{
+		_protagonist = GameObject.FindObjectOfType<Protagonist>().transform;
```
DeivSky, line 31: You should cache this on `Awake`, or, actually, it should be in a field in `FaceProtagonistSO` so it can be set from the inspector. Also, I'd rename the action to `FaceTarget` since this could be useful for other things as well.

UOP1_Project/Assets/Scripts/Characters/StateMachine/Actions/FaceProtagonistSO.cs
```diff
@@ -0,0 +1,37 @@
...
+	public override void Awake(StateMachine stateMachine)
+	{
+		_actor = stateMachine.transform;
+	}
+		
+	public override void OnUpdate()
+	{
+		if (_protagonist != null)
+		{
+			_actor.LookAt(_protagonist);
```
DeivSky, line 25: Note that this is creating some probably undesired behavior!
image
😆 
Rotation should be only on the Y axis.

UOP1_Project/Assets/Scripts/Characters/EnemyColliderZoneController.cs
```diff
@@ -0,0 +1,26 @@
...
+		}
+	}
+
+	private void OnTriggerExit(Collider other)
+	{
+		if (other.GetComponent<Protagonist>() != null)
+		{
+			_exitZone.RaiseEvent();
+		}
+	}
```
DeivSky, line 25: This could be a very generic `ZoneTriggerController` if you detected the collision by layer:
``` csharp
[SerializeField] private LayerMask _layers = default;

private void OnTriggerEnter(Collider other)
{
	if (((1 << other.gameObject.layer) & _layers) != 0)
	{
		// stuff
	}
}
```

DeivSky (changes requested):
Hey, looking great! I'm leaving just a few comments.

And besides that, I have to ask, are you enjoying the StateMachine? Feel free to leave any feedback in the [thread](https://forum.unity.com/threads/generic-state-machine.979371/)! 😃

UOP1_Project/Assets/Scripts/StateMachine/Editor/Utilities/TransitionDisplayHelper.cs
```diff
@@ -46,7 +46,14 @@ internal bool Display(ref Rect position)
...
 				rect.x += 20;
-				LabelField(rect, SerializedTransition.ToState.objectReferenceValue.name, EditorStyles.boldLabel);
+				if (SerializedTransition.ToState.objectReferenceValue != null)
+				{
+					LabelField(rect, SerializedTransition.ToState.objectReferenceValue.name, EditorStyles.boldLabel);
+				}
+				else
+				{
+						LabelField(rect, "Unknown", EditorStyles.boldLabel);
+				}
```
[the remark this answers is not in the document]
DeivSky: Actually, I am checking and logging it in `TransitionTableEditor.cs`, but there's an error in the script!
``` csharp
if (serializedTransition.FromState.objectReferenceValue == null)
{
	Debug.LogError("Transition with invalid state found in table " + serializedObject.targetObject.name + ", deleting...");
	_transitions.DeleteArrayElementAtIndex(i);
	serializedObject.ApplyModifiedProperties();
	Reset();
	return;
}
if (serializedTransition.FromState.objectReferenceValue == null)
{
	Debug.LogError("Transition with invalid state found in table " + serializedObject.targetObject.name + ", deleting...");
	_transitions.DeleteArrayElementAtIndex(i);
	serializedObject.ApplyModifiedProperties();
	Reset();
	return;
}
```
I'm checking `FromState` twice!! 🤦🏻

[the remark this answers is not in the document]
treivize: My bad, indeed, I should not have added this null check in the middle of this PR... 
Indeed, I did that, because I experienced fatal exception in case of state deletion: Transition Table was crashing when opening it if a ToState was missing! 

I better understand why with your last comment! It is indeed better like that :) I will update my PR to remove the null check!

BTW: I did not follow actively the SM thread (I am already bothered on other topics and I am mostly using my free time to contribute), despite that it was easy to understand how it works and how to use it to build the behavior I wanted to put in place, especially without having to write lots of boiler plate code! Great job! 
BTW2: I have also learned lots of things reading the code you have written here!

UOP1_Project/Assets/Scripts/Characters/StateMachine/Actions/FaceProtagonistSO.cs
treivize: Actually, I initially put this in Awake, but it does not work because here the protagonist is instantiated by the spawning system after Awake is triggered. 

I was thinking also about having a target more generic not Protagonist specific, but how do you specify a target that has not been instantiated yet in the SO?

ciro-unity:
Hey, I merged the fix by @DeivSky on #244. Does that affect you in any way?

treivize:
> Hey, I merged the fix by @DeivSky on #244. Does that affect you in any way?

No, there is impact, the issue was when building the transition table when deleting or renaming states, but here everything is wired properly. And in my last commit, I have removed the null check I had introduced initially in TransitionDisplayHelper, but was not needed after the fix from DeivSky.

It is ok to merge this one, the implementation is simple and most of the things are right. But as described in my last post in Plant Critter thread, I have another version in the pipe, that might be more future proof. I have the code in another branch deriving from this one:

[Fork PR](https://github.com/treivize/open-project-1/pull/1)

UOP1_Project/Assets/Scripts/Characters/StateMachine/Actions/FaceProtagonistSO.cs
DeivSky: > Actually, I initially put this in Awake, but it does not work because here the protagonist is instantiated by the spawning system after Awake is triggered.
Hmm, I've actually had this at the back of my mind for a while... The **SM** is calling `Awake` on the initial state _during_ the **SM** `Awake`. I think it should be called on `Start` to solve this.

ciro-unity: @treivize that's a good question.

Actually, one of the last things I implemented might be an answer to this. [I mentioned it here](https://forum.unity.com/threads/decouple-cameramanager-and-spawnsystem-card-194.1016932/#post-6592348).

You could use one of those to connect this ActionSO to the player's transform, even if you don't know yet if it's been instantiated or not. You could just implement a check to see if it's null. If it's not, it will be the player's transform.

The check is also not super necessary, since it will only happen if the player has entered the TriggerZone, so the player exists at that point.

ciro-unity:
Honestly I think this one is better, the idea of using names and strings for finding the player is pretty dangerous.

I've also written it in one of the comment (on this PR), but regarding your hard time in how to find the player's transform, we might use the RuntimeAnchors concept I've implemented in the project as of late. [I explained it here](https://forum.unity.com/threads/decouple-cameramanager-and-spawnsystem-card-194.1016932/#post-6592348).

So if you want, I can go ahead and merge this, implementing the anchor in it.

treivize:
> Honestly I think this one is better, the idea of using names and strings for finding the player is pretty dangerous.
> 
> I've also written it in one of the comment (on this PR), but regarding your hard time in how to find the player's transform, we might use the RuntimeAnchors concept I've implemented in the project as of late. [I explained it here](https://forum.unity.com/threads/decouple-cameramanager-and-spawnsystem-card-194.1016932/#post-6592348).
> 
> So if you want, I can go ahead and merge this, implementing the anchor in it.

Definitely this one looks more robust, the only aspect that I do not like this fact that there is no link between the element entering into the alert zone and what the plant is facing which is always the Protagonist, assuming that it is necessary him who has entered the zone...  Maybe I should not care since it might be the case in the type of game we are building. 

Just regarding the other version, there is no hard coded prefab name expected in the SO, the name is fetched from the collider detection script and injected into the SO.

For the change to be made in the FaceProtagonistSO, I think I can make it, I was wondering if it could make more sense to retrofit my V2 approach using the new TransformEventChannelSO to send the ref of the element entering the collider to the facing script?

ciro-unity:
> For the change to be made in the FaceProtagonistSO, I think I can make it, I was wondering if it could make more sense to retrofit my V2 approach using the new TransformEventChannelSO to send the ref of the element entering the collider to the facing script?

I think it's a bit more safe if you go with the RuntimeAnchor stuff I was talking about. I wouldn't necessarily trust two actions in the StateMachine to send events to each other, if we change anything in the SM (i.e. introduce some kind of caching, recycling, etc.) that might introduce a bug.
With the Anchor way, if the protagonist is in the scene at all, that Transform reference will be there. And if for some reason it's not, there's a lot of other systems that will fail too giving us a lot of warning.

treivize:
No problem, I understand. I am fine with it. I am doing the change to use PlayerTransform SO in FaceProtagonist script, I thought that this SO was already set somewhere, but it seems it is not the case, am I wrong? I am planning to set it in SpawnSystem right after the player instantiation.

ciro-unity:
> I thought that this SO was already set somewhere, but it seems it is not the case, am I wrong?

Sorry, I should have really just linked it (since it's in the repo). The RuntimeAnchor is in this folder: https://github.com/UnityTechnologies/open-project-1/tree/main/UOP1_Project/Assets/ScriptableObjects/RuntimeAnchors

treivize:
> > I thought that this SO was already set somewhere, but it seems it is not the case, am I wrong?
> 
> Sorry, I should have really just linked it (since it's in the repo). The RuntimeAnchor is in this folder: https://github.com/UnityTechnologies/open-project-1/tree/main/UOP1_Project/Assets/ScriptableObjects/RuntimeAnchors

Actually I had found it, but what I meant is I did not find where PlayerTransform is initialized with the Player transform. So when I used it in my FaceProtagonist, the inner check isSet was always false. That is why I have added the PlayerTransfrom anchor to the SpawnSystem to be set right after its instantiation. 
image

ciro-unity:
Yes, I was confused for a second because I thought that I had done it already, but no... I just prepared the setup :D
So yeah, that's where you'd do it. Also because the spawn will change a bit (to support multiple entry points) so it's that which should take care of spawning and destroying the player prefab.

treivize:
Ok, I have just moved this logic in Spawn() method next to _playerInstantiatedChannel event raising, I think it makes more sense to be here instead of the gameObject instantiation from prefab method (InstantiatePlayer()).

I am not planning to push any further update to this PR, it should be fine now.

DapperDino:
What about if you have multiple instances of the same enemy? Won't the scriptable object events become an issue? If one enemy is alerted of the player and changes state, won't all enemies receive that event and switch state? Maybe I'm misunderstanding the use of events here?

treivize:
> What about if you have multiple instances of the same enemy? Won't the scriptable object events become an issue? If one enemy is alerted of the player and changes state, won't all enemies receive that event and switch state? Maybe I'm misunderstanding the use of events here?

oh god! you are perfectly right! Indeed the event is shared across all instances :( So they will act in sync what ever where they are from the player...

DapperDino:
Do you have any solutions for this that still uses Scriptable Objects or will we have to revert to using normal C# events?

treivize:
> Do you have any solutions for this that still uses Scriptable Objects or will we have to revert to using normal C# events?

I have no smart ideas right now... I guess I have to go using UnityEvent editable in the inspector and wiring the actions... Even for the conditions I am not sure how to implement it they are based on SO, so I have the feeling that using UnityEvent to trigger a function in it will do it for all the instances of the plant. I have to test it

DapperDino:
Let me know how it goes and I'll try coming up with a solution in the meantime 👍

treivize:
Ok! Here is my proposition to handle this problem:

I am storing the collider state into a new MonoBehavior script "Critter":
image

The data are set by each ZoneTriggerController associated to the collider via UnityEvent<bool>
image

Then the data are read by a revamp SM ConditionSO script:
image

treivize:
Should this PR target main or art-assets?

treivize:
Just an updated version of the implementation diagram:
image

ciro-unity:
Great PR, @treivize! I'm going to merge it soon, maybe with a few minor adjustments and also I want to get the other one you made and merge them together.

What is the big cylinder collider doing on the base object though?
image

ciro-unity:
Ah, I guess it's to impede the character from jumping over it? Never thought about this. Do we want to stop that?

ciro-unity:
Also another question: why all the getter/setter on the variables in the `Critter` script? `_isPlayerInAttackZone`, `_playerInAlertZone`... They are not serving a purpose right now, why not just make the variables public?

treivize:
Ah! So happy they fit with what is expected by the project!

Indeed it was made to prevent the player to jump on its head. Maybe it would be preferable to prevent it differently by repulsing the player if attempted. Just the plant critter attack will be a bit odd (not facing the player) if the player is standing up on its head. 

In the other PR I was thinking about disabling some exit time in the animator of the pig and the critter to make the attack and the hit more aligned.

treivize:
About the getter/setter no particular reason. I was looking at some guidelines in C# and how to expose data from a simple class and I found this pattern. Honestly, I feel the same, public is equivalent and less verbose.
I will push an update tonight accordingly.

ciro-unity:
No, don't worry about modifying this PR, I'm merging some stuff so I can do it.

ciro-unity:
I actually realised one use for the getter/setters: you're using them in the UnityEvent on the Critter's ZoneColliders, to feed the boolean value directly as part of the event being passed along. So I guess they can stay, maybe I'll add a comment in the script to explain it.

treivize:
Indeed but I think Unity event should be able to get/set public properties as well, right?

Are you planning to merge both PR together? Because I had in mind the following improvements for the combat part :
1. Improve the ActiveWeaponActionSO to activate the weapon only during a sub part of a state with a delay from start plus duration, since the attack animation contains a start and an end that should not be armful.
2. Change the Killed trigger by a Boolean in plant critter animator 
3. Add a "Fatal Hit" state followed immediately by the faint animation in the plant critter with a transition when killed equals true. It should replace the current faint state.
4. Add killed == false in the transition to GetHit in the plant critter animator and unchecked has exit time for all transition to get hit and fatal hit 
5. Unchecked has exit time in pig chef animator for the transition between walking and canehit 

Just in case you would be interested in including some of them in your merge otherwise it can be submitted in a later PR.

ciro-unity:
> Indeed but I think Unity event should be able to get/set public properties as well, right?

Yes, but apparently you still need to provide getters/setters... TIL 🤷🏻‍♂️ 
So I will just shorten them to the classic `{ get; set; }` syntax.



> Are you planning to merge both PR together? Because I had in mind the following improvements for the combat part :

Yes, I was going to merge it all before you make more adjustments, if you don't mind. I can tell you I had already implemented 2 and 5. For the rest, are you thinking of a random critical hit?

treivize:
Not a critical hit, even if it might be a good idea 👍 
I think the current animator for the plant critter is not optimal when fainting, because, currently the last hit before dying sometimes do not trigger the get hit animation, but directly the fainting one from alert, which looks a bit odd to me.
I will have a look and try to propose an enhancement from your merge.

BTW, I am very glad to see these PRs finally integrated in the project!

treivize:
I just pull the project after your merge and release that you actually implemented the change I just had in mind for the plant critter animator and the faint transition (making transitioning from BeenHit state only) I did some tests and I do not know yet why, but sometimes the plant is dying without going through the expected animator states, just stays in Alert and disappear after a while!
I am trying to understand why...

Issue #242, opened 2020-11-30
[Object Pool] HasBeenPrewarmed is not reset in between play sessions


davejrodriguez:
**Short description**
`PoolSO.HasBeenPrewarmed` is a boolean that prevents a pool from being prewarmed multiple times over its lifetime. It switches to true once `PoolSO.Prewarm()` is called. However, `PoolSO.HasBeenPrewarmed` remains true even after you exit play, so when Prewarm is called in a subsequent play session, the pool is prevented prewarming.

**Expected behaviour**
`PoolSO.HasBeenPrewarmed` should be reset to its initial value of false when exiting play mode. 

**To Reproduce**
1. Create an instance of a pool as an asset.
2. Create script that takes in a reference to a pool and call `pool.Prewarm(10)` in its `Awake` or `Start` method.
3. Attach the script to a GameObject in scene and set the pool reference to the pool asset you created.
3. Press play.
4. Exit play mode.
5. Press play again.
6. Notice you get a warning saying the pool has already been prewarmed, but no objects were created as members of that pool.

**Notes**

Should be an easy fix. Just need to reset `HasBeenPrewarmed` to false in the `OnDisable` method of `PoolSO`.

ciro-unity:
Ah! That's why! I was going nuts trying to understand why it was verified as `true`, it makes sense... it's part of the SO! 🤦🏻

davejrodriguez:
Yeah, my bad. That was an oversight when I added the check. The fact that playmode changes persist on SO assets is a bit of a double-edged sword 😄

PR #243, opened 2020-11-30
[Object Pool] Fix HasBeenPrewarmed Reset

davejrodriguez:
Issue: #242 
Forum thread: https://forum.unity.com/threads/generic-object-pool.990776/#post-6568468

I reset HasBeenPrewarmed to false in the OnDisable method of PoolSO.

To verify:

1. Open the Initialization scene in the project.
2. Press play.
3. Notice the AudioManager has a pool as a child and it has been prewarmed.
4. Exit playmode.
5. Press play.
6. Notice the AudioManager still has a prewarmed pool as a child and no warning is logged about the pool already being prewarmed.


ciro-unity:
Nice, clean and low-risk fix 👍🏻

Issue #245, opened 2020-12-01
Mouse pointer erratic behaviour


ciro-unity:
**Short description**
Sometimes, the camera spins very fast when using the mouse to control the camera, even if the mouse is moved by a tiny fraction.

**Expected behaviour**
The camera should follow the speed of movement of the mouse.

**To Reproduce**
1. Open the scene "TestingGround"
2. Press Play
3. Once the game is in Play Mode, hold down the RMB and move the mouse left or right by a very tiny amount.
4. Notice the camera spins around a full circle, before resuming at the right speed.

**Notes**
- [See this video for an example of the issue occurring](https://youtu.be/fGTnzSzSmrI)

kgc00:
hi @ciro-unity , i just saw this- we could just have mouse movement without RMB to sidestep this.  sound good?  I'm also down to debug if you'd like.

adoart:
It looks like it depends where the click starts, the further from the center bigger the rotation, @ciro-unity clicked once close to the center and rotation was little. That's what I got from the video...

khush2o2o:
I known this bus to how to solve...

drbier1729:
Just tried to check this out but I'm unable to replicate the issue. 

I'm guessing that maybe the InputReader or InputSystem isn't updating the cursor's position when it jumps to the center when the cursor is locked (in CameraManager.cs). This causes the mouse delta on the first frame when the mouse is moved after enabling camera controls to be crazy high since it is calculating the difference between the position where the RMB was clicked and the center of the frame.

I'm pretty new to Unity so this might be incorrect, but hopefully this noticing is helpful! I'll keep working on it but without being able to recreate it, I'm not optimistic about producing a fix.

ErikBehar:
> I have an idea to slove this bug...

Please provide the idea in writing or a link to the PR, stating that you know how to solve a bug with no further info is not useful to anybody.

ErikBehar:
ok so I think @drbier1729 has it correct ... you can more easily replicate the issue, by popping out the game window, then click outside so the window looses focus and then click back into the window to get focus, then rm on the edge of the window. You will get the crazy spin and yeah I think its due to the initial difference. 

with the above steps you can replicate the issue many times within 1 play session.

davejrodriguez:
Could someone who was able to reproduce the bug test #249?

drbier1729:
[...]

PR #83, opened 2020-10-12
Added Drag Composite

davejrodriguez:
Added Drag Composite to allow for mouse drag (or any other Button+Vector2 input)
Added a Right-click Mouse Drag as a composite for Rotate Camera
Camera now rotates with Right-click mouse drag

PR #249, opened 2020-12-02
[WIP] Simplified Input to use Composite

davejrodriguez:
Issue: #245
Forum: https://forum.unity.com/threads/mouse-pointer-erratic-behaviour.1014424/#post-6581881

The way we are currently going into and out of "rotate camera mode" via a separate right mouse button action that is _only_ for setups with a mouse is not really the best way to utilize the Input System. I'm unable to reproduce the bug on my machine, but my theory is that switching modes like this is causing the erratic mouse behavior.

A better way is to use a composite where you use the right mouse button as a trigger and output the mouse delta as the delta. Then you can set the composite as a binding of the singular camera rotate action instead of several mouse specific actions. So I've made a composite called DragComposite (I made this previously in #83, but we were unsure then if mouse drag was desired) which takes a button "trigger" and a Vector2 "delta". This is easily reused and modified.

I have also modified the CameraManager, InputReader, and TestingGrounds scene to work with these changes.

To verify:
1. Play TestingGrounds.
2. Right click and drag to rotate mouse.

### DIFF
--- a/UOP1_Project/Assets/Scenes/TestingGround.unity
+++ b/UOP1_Project/Assets/Scenes/TestingGround.unity
@@ -2096,6 +2096,27 @@ Transform:
     type: 3}
   m_PrefabInstance: {fileID: 69359815}
   m_PrefabAsset: {fileID: 0}
+--- !u!1 &73056683 stripped
+GameObject:
+  m_CorrespondingSourceObject: {fileID: 8745341641394998848, guid: 45632f0a227c860489bcba0eb1f4ec3e,
+    type: 3}
+  m_PrefabInstance: {fileID: 1810604289}
+  m_PrefabAsset: {fileID: 0}
+--- !u!114 &73056684
+MonoBehaviour:
+  m_ObjectHideFlags: 0
+  m_CorrespondingSourceObject: {fileID: 0}
+  m_PrefabInstance: {fileID: 0}
+  m_PrefabAsset: {fileID: 0}
+  m_GameObject: {fileID: 73056683}
+  m_Enabled: 1
+  m_EditorHideFlags: 0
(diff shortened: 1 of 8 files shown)


ErikBehar:
you removed the cursor lock, so I can't repro the bug, because the thing causing the bug is removed. I assume we still want the cursor lock / hiding the cursor.

davejrodriguez:
Sure, makes sense. I'll add it back in with this solution.

ciro-unity:
Where are we with this PR? I think @ErikBehar's fix solved the mouse issue (although we need some testing of course!)

davejrodriguez:
> Where are we with this PR? I think @ErikBehar's fix solved the mouse issue (although we need some testing of course!)

Still working on getting cursor lock and visibility back in. It's certainly harder without the extra action, but I think it's worth the trouble, especially when we start implementing rebinding logic. This PR should probably be considered an enhancement rather than a bug fix now that @ErikBehar fixed it in his PR.

ciro-unity:
Sounds good. Just be aware that there were some changes to the `CameraManager` recently (shouldn't conflict with your code).

ciro-unity:
Is it still a work in progress?

davejrodriguez:
Not active progress. I will close for now.

ciro-unity:
Just out of curiosity, so the issue with your implementation was that we lost the control over the cursor?
For the rest it was working...?

davejrodriguez:
The issue is that in my efforts to simplify the camera rotate actions (one specifically for mouse) with a single composite, I made it harder to enable/disable cursor lock and visibility at the right times. I think it's possible and still think it's cleaner than having a dedicated camera mode action. But until I can get it working with the desired cursor functionality, I'll refrain from clogging up the open PRs :)

PR #250, opened 2020-12-03
Issue #245 simple fix for camera erratic behaviour

ErikBehar:
Issue #245 

Description:
Camera behaves erratically when rotating it from edges of window.

Cause:
Cursor lock snaps pointer to center creating large diff for that frame

Solution:
Resolved by adding a small pause until allowed to move camera

Test:
1) pop out Game window
2) make window big 
3) move mouse to edge, click right mouse button, hold down and move
4) let go of mouse, notice that pointer is in center of screen now


ciro-unity:
Ignore the history of commits, I'm not sure how I managed to merge all of the commits I touched today into this PR :P

PR #251, opened 2020-12-03
Added details to furniture/tableware items and added colliders on furniture items

treivize:
Added some details on tableware and furniture items (fork, spoons, large bowl and large dish)
image

Added colliders to furniture items
image


ciro-unity:
Super nice details! Love the little bird in the plate!

PR #252, opened 2020-12-03
Added Frying Pan asset

erizzoalbuquerque:
card: https://open.codecks.io/unity-open-project-1/decks/16-3d-shading/card/17i-accessories-for-townsfolk
forum: https://forum.unity.com/threads/accessories-for-the-townsfolk.1007900/ 

Added a Frying Pan. (564 tris)

image

image


ciro-unity:
Really nice! I love the fact that it's a bit low-poly and not too round. The texture is great!

PR #253, opened 2020-12-03
BugFix Creating new Scene (Change to ProjectSettings)

DSivtsov:
**[for Bugfix PRs]**  
For creation new SOlocation & corresponding Scene files, demanding to add new Scene to "Scene in Build", but when I try to add Scene file (at Build Setting), I receive message "Unable to access service".

This forum thread [linked to an issue](https://forum.unity.com/threads/creating-new-scene.1015867/#post-6585751).
I resolve this issue:
- Made new Link to Services and create new Unity Project ID
- After that I'm unlink the Project from Unity Services and made commit based on changed file (UOP1_Project\ProjectSettings\ProjectSettings.asset)

After that all work fine, If you return to prev file version the problem is return.

### DIFF
--- a/UOP1_Project/ProjectSettings/ProjectSettings.asset
+++ b/UOP1_Project/ProjectSettings/ProjectSettings.asset
@@ -670,10 +670,10 @@ PlayerSettings:
     m_VersionCode: 1
     m_VersionName: 
   apiCompatibilityLevel: 6
-  cloudProjectId: eadb991e-24e4-432a-9a06-35e20daca02b
+  cloudProjectId: 
   framebufferDepthMemorylessMode: 0
-  projectName: Open Project 1
-  organizationId: cirocontinisio
+  projectName: 
+  organizationId: 
   cloudEnabled: 0
   enableNativePlatformBackendsForNewInputSystem: 1
   disableOldInputManagerSupport: 1


ciro-unity:
Hey @DSivtsov. I'm not sure I can accept this change. We need the project to keep the same project ID for all of its users.
As I mentioned on the forums, I don't see other people having the same issue as you, so can you try removing the project completely, including from the Hub, and redownloading the project from scratch in a new folder? Thanks.

DSivtsov:
> Hey @DSivtsov. I'm not sure I can accept this change. We need the project to keep the same project ID for all of its users.
> As I mentioned on the forums, I don't see other people having the same issue as you....

**Yes. BUT** Also no anyone who replied that he check it and all works ok - many participants, who testing the "creation Scene functionally" can simple created new project ID and continue they work, because it nothing changing.

> Hey @DSivtsov. I'm not sure I can accept this change. We need the project to keep the same project ID for all of its users.

As I propose the one variant - It's leave the project ID empty, in this case **it will work** also and **all users will have the same value.**

> ... so can you try removing the project completely, including from the Hub, and redownloading the project from scratch in a new folder? 

I can check it, but doesn't see any sense, because i twice checked, that problem begin when in ProjectSettings.asset exist the line - organizationId: cirocontinisio. And It's a logical, because the most users of this project use the Unity Personal version, where every Person it's "a separated organization", and will be strange if many different peoples ("organizations") can work with same name ("organization") - "cirocontinisio"?

ciro-unity:
> I can check it, but doesn't see any sense, because i twice checked, that problem begin when in **ProjectSettings.asset exist the line - organizationId: cirocontinisio.** And It's a logical, because the most users of this project **use the Unity Personal version**, where every Person it's "a separated organization", and will be strange if many different peoples ("organizations") can work with same name ("organization") - "cirocontinisio"?

I understand that users shouldn't be in that organisation, but we are also not using Services so the organisation shouldn't matter.
This has nothing to do with Unity Personal or Pro.

It's not that we need everyone to have the same ID, we need everyone to have **that ID** since we're checking how many people keep using the project.

Let me know if removing it from the Hub completely helps.

DSivtsov:
> Let me know if removing it from the Hub completely helps.
 @ciro-unity I did it. It was a [little challenge](https://forum.unity.com/threads/git-questions.985394/#post-6609151) :) .
But Nothing changed.
I'm fully remove the local files and also branch on origin. Make fork again and after cloned to local folder.
After this, When I load the project and when I trying to access the project "Build Settings", receive the same message & problems as I described early. All work normal only after I changed the (UOP1_Project\ProjectSettings\ProjectSettings.asset):

`$ git diff --oneline --graph main...BugFix
diff --git a/UOP1_Project/ProjectSettings/ProjectSettings.asset b/UOP1_Project/ProjectSettings/ProjectSettings.asset
index cf00737..f407560 100644
--- a/UOP1_Project/ProjectSettings/ProjectSettings.asset
+++ b/UOP1_Project/ProjectSettings/ProjectSettings.asset
@@ -670,10 +670,10 @@ PlayerSettings:
     m_VersionCode: 1
     m_VersionName:
   apiCompatibilityLevel: 6
-  cloudProjectId: eadb991e-24e4-432a-9a06-35e20daca02b
+  cloudProjectId:
   framebufferDepthMemorylessMode: 0
-  projectName: Chop Chop
-  organizationId: cirocontinisio
+  projectName:
+  organizationId:
   cloudEnabled: 0
   enableNativePlatformBackendsForNewInputSystem: 1
   disableOldInputManagerSupport: 1`

ciro-unity:
Closing since:
- The issue doesn't seem to happen to anyone by the DSitsov - so we can't fix it on our end. Sorry!
- The issue is a non-issue now, since we don't add scenes to BuildSettings but we use Addressables to include them in AssetBundles.

Thanks for reporting it anyway!

PR #256, opened 2020-12-06
Feature/code: Editor for ScriptableObject with events

Tatranskymedved:
Added editor for ScriptableObject with events. It allows to automatically show buttons for each event and raise it. Currently it is hooked only on the input SO (so on `InputReaderSO`), but can be attached simply to any other SO and allows user to invoke events from UI where argument are not needed.

I've used reflection to check the SO for event and via backing field they're invoked.

No forum or card link.

image


amel-unity:
Great contribution @Tatranskymedved ! Thanks a lot, this will be useful to help make tests faster in the future. I will merge it now :)

Issue #257, opened 2020-12-08
LocationSO not serialized properly


rainaldisatria:
**Short description**
LocationSO doesn't serialize some of it fields.

**Expected behaviour**
LocationSO should serialize all of it fields.

**To Reproduce**
1. Go to ScriptablesObjects/SceneData/Locations
2. Click on any of LocationSO. Let say Beach.
3. Add Music reference to the selected LSO.
4. Deselect the selected LSO.
5. LSO is not serialized. Music is from the reference is gone.

ciro-unity:
Nice find! Not super urgent to fix, but it can be a nice first task for somebody!

ciro-unity:
LOL ok nevermind, I didn't see you already put the fix in a PR :)
So fast :)

PR #258, opened 2020-12-08
Fix/LocationSO not serialized properly

rainaldisatria:
Is this PR linked to an issue? 
https://github.com/UnityTechnologies/open-project-1/issues/257

By applying serializedObject.Update() and serializedObject.ApplyModifiedProperties() in OnInspectorGUI().

Music is now serialized properly.


ciro-unity:
Very thoughtful to display the script reference in a non-editable field :)

I was thinking, we should try to abstract this and maybe put it into a base editor class, so we can use this "trick" with multiple Inspectors. What do you think?

rainaldisatria:
@ciro-unity Sounds like a good idea. 
Do you want me to create it in this PR?

ciro-unity:
No, for now I'll merge this one, then you can figure out how to make it a clean solution maybe - if you have time and will.

PR #259, opened 2020-12-08
Whiteboxing

NicknEma:
This is the updated version of my beach witeboxing proposal

Whiteboxing thread: https://forum.unity.com/threads/whiteboxing.987468/page-2#post-6569323

I added lightning, the Spawn System and the Camera Manager to make the scene easier to test.

(I may also have accidentally done something else in the process, if so my bad, tell me if something needs to be fixed before merging)


ciro-unity:
Thanks, I've merged it! But I think the beach is unwalkable...? Something to do with the ground collider, maybe. No big deal, I merged it anyway because I think it's still interesting to see it, even static.

PR #260, opened 2020-12-08
Added small internal cooldown to the PlayLandParticlesAction StateAction

MTrecozzi:
Addressing this open ticket:
https://open.codecks.io/unity-open-project-1/decks/15-code/card/18k-avoid-spamming-particles-on-sliding

Added a small solution to LandParticle Visuals being spammed when the state machine rapidly switches between the sliding and idle state, which was shown to happen in the ticket above.

After this change, Whenever the PlayLandParticlesAction is exited by the state machine, it checks if total elapsed time is greater then the total elapsed time of the last particle call + a small amount.

If so, only then do we tell the DustParticleController to play, and we also set the new time to beat to be the current time.


ciro-unity:
Thanks, nice fix! I just removed the unnecessary function call, I'm going to merge it now!

PR #261, opened 2020-12-09
Refactor Particle Effect Action

treivize:
Forum thread: [Puff of dust effects](https://forum.unity.com/threads/puff-of-dust-effects.1002936/)

I have come with this refactor when I tried to reuse the dust landing particle effect on the plant critter when it starts the alert mode.
The DustParticlesController which references all the dust effects attached to the PigChef prefab was not at all fitting with my need and also because the Landing effect action SO was triggered always on state exit whereas, I would have need mine to be triggered on state start. 

So I started to question the current mechanism to trigger the dust effects and this kind of one shot effect based on particle system in general.
Is it a performance requirement to have them predefined in the dynamic object prefab from the beginning? Could it be acceptable to instantiate the effect on the fly when needed and destroy it when it is done? Like it is done for dissolve effect for instance.
And finally replace the specific action script associated to each effect by a generic action for effect based on a particle system.

The proposed solution is one ParticleEffectActionSO script which will let the designer select the particle prefab to trigger, decide when to trigger it (on start or on exit) and finally force the stop of the effect at state exit or not:

image

image

image

This script is assuming that the particle effect will be instantiated as a child node of the state machine gameobject transform.


erizzoalbuquerque:
Nice work! This refactory is very necessary. I'll download your branch so I can test everything properly. For now, here some observations.

1) We could use this PR as an opportunity to solve this card:
https://open.codecks.io/unity-open-project-1/decks/15-code/card/18j-sync-amount-of-landing-particles-to-fall-speed
Some particles need to have some of their values sync before they play (or maybe while they are playing?). One way I can think of is using the SO configuration to override some values of the ParticleComponent. I'm using the OP1 Audio System as inspiration. Actually, I think the audio system can be a great source of inspiration for the particle effect system. (but it doesn't need to be as complex) 

2) I'm not sure if it's better to instantiate the particle effects when they are needed or to have them in the character prefab. If we're using the first option, maybe having an object pool would be a great idea (like the audio system)

3) Also, if we're instantiating effects when they're needed, we should think about a way for fine positioning them. Right now, it seems that the SO doesn't allow position and rotation to be set.

4) I'd change "When To Run" to "When to Play" since the method used to start a particle is called "Play" :) 

I still have little experience with the State Machine and Scriptable Objects since I barely code in this project. Sorry if I'm saying something very wrong.

I'm not trying to make this refractory very complex. I'm very fine if the scope of this refactory is small. Points 1) and 3) are the main ones, but maybe if this refactoring gets bigger it will deserve a forum post of its own.

erizzoalbuquerque:
I tested the branch. The character particles are behaving they used to. Great!

treivize:
Thanks a lot @erizzoalbuquerque for your review! Indeed all your points make sense actually.

about your point 1, I completely missed this requirement from the roadmap! Thanks to have brought it back. I do not know yet how to provide a generic way of controlling that... Maybe it is a bit too much to provide this control and having a MonoBehavior script attached to the LandingEffect prefab to tweak the settings based on its own speed might be more reasonable. I do not know if we will need dozen of particle property tweaking in the effect we are building... 

about 2 and 3, yes I also had them in mind :) for point 3 I was thinking about maybe adding it later incrementally based on the needs.

Indeed let's bring the discussion in the forum :)

ciro-unity:
Honestly, I'm tempted to reject this refactor altogether. I don't see much gain in making this system generic, on the contrary now the particle systems are being instantiated and destroyed for no reason, an operation which is generally costly and should be avoided if possible.

Maaaybe, we could consider a refactor like you did but, instead of destroying the particles when done, the Action just keeps a reference to them.
The issue with this approach though is that you can't pre-place the particles in the Prefab. That might be ok for jump particles that are generally at 0,0,0, but annoying for others. Also it become harder to preview them as you author them.

Another example: for the lantern, we might need particles that are parented to the lantern and play when the Pig swings it (i.e. when he burns things with it).
If the Particles are generated by the action, we would need a way to parent them to the right object. But the SO can't hold a reference to an object in the hierarchy of the Pig.

These are all things to consider before we move forward with this refactor...

treivize:
I definitely understand your regarding the placement of the effect on a specific sub object in the hierarchy, I discovered this problem indeed when looking at proposing an effect anchor solution. 
But regarding the instantiation/destroying of the particle system, I am not sure to understand, I though that using Pool/Factory feature developed in this project was solving this problem here. 
The particle effects are instantiated only when needed, otherwise they are re used from their dedicated pool

Honestly, I am ok if this PR is rejected, I am mainly here to learn what is good to do vs not. Looking at all the discussions regarding the feature put in place (SM/Pool/Event/etc...) I had the feeling that MonoBehaviour scripts with boiler plate code are evil and should be used in exceptional cases.

treivize:
I tried to find a solution to keep somehow the action to trigger an effect quite generic and still be able to place the effect properly from the object hierarchy.

First, like said in my previous comment, the effects are now based on Factory/Pool so there is no more Destroy of PS, they are just moved to a pool when finished or stopped.

Here is the addition to my previous approach: a new SO level has been added called ParticleEffectSO which reference a ParticleEffectPoolSO. The idea is to be able to reuse for multiple purpose the same effect and share the same pool.

The SM is now referencing this new SO instead of the Pool:
image

Now to be able to control the effect at runtime on the object instance, a MB controller has been put back, but a bit more in a generic way:

image

A list of effects can be defined associating a ParticleEffectSO and a Transform defined from the object hierarchy.

ciro-unity:
> But regarding the instantiation/destroying of the particle system, I am not sure to understand, I though that using Pool/Factory feature developed in this project was solving this problem here.
> The particle effects are **instantiated only when needed**, otherwise they are re used from their dedicated pool

"Instantiated when needed" is exactly the contrary of using a pool. When using pools, you normally want to instantiate them when the game is not doing much (so to speak) and thus you have time (in the frame budget) for the instantiation. Or, you usually do it all at the start (or better, in a loading screen) so you don't risk of incurring in any stutter during gameplay.

Pools solve another problem, which is to re-use ParticleSystems that are not used frequently for the same object, so they can be reused for many. But this is not the case for the player's jump particles! (or the critter run, or similar). We need them all the time and they are probably quite unique, so it's pointless to put them in a pool.

Think for instance of the fact that we want the landing particles to be measured according to the height of the jump (there's a PR for that). If you pool those particles and reuse them on another jumping character, now you need to make sure the right amount is emitted every time so you need to build that check into the other character's jump - even if they always jump the same height.

For these reasons, again, I don't see much point in making the character's particles pooled.
We can however have an action for the state machine to emit particles or effects that we really want to pool, but I can't right now pinpoint which ones they could be (the fire effect of the lantern when it burns an object? but that is maybe in the shader...)

treivize:
For your first point, I agree with you, but it seems just a matter of how the pool is configured at creation time, it can be initialized to create the element at start and not when needed the first time. 

Definitely, I understand your point about this particular case of the character effect, indeed, it might be not satisfactory to have this refactor. 

But, just to explain why I wanted to do this refactor to let me understand how to implement the next step I had in mind. 
When I played with Plant Critter behavior, I though about adding the same landing effect as the player has after a jump when the plant opens during alert state to emphases the animation, so I looked at the particle controller of the character and I found it really specific.

In addition, assuming that you may have 10/20 plant critters on the map, I felt that it might not be relevant to have on each a particle system attached to each instance that will play the same effect only one at a time when the player is coming closer. It sounds a good candidate for the pooling system. I was just planning to include the character effect since I was planning to reuse the same effect.

If you think this additional effect when the plant is opening, maybe we could just reserve this refactor for this usage. No problem.

ciro-unity:
> so I looked at the particle controller of the character and I found it really specific.

Hmm, it's just specific to characters that move and jump. In all fairness, most probably we will only have the Pig that does that... but that's ok too, it's fine to have specific scripts especially if they are related to the protagonist.

> In addition, assuming that you may have 10/20 plant critters on the map, I felt that it might not be relevant to have a particle system attached to each instance that will play the same effect only one at a time when the player is coming closer. It sounds a good candidate for the pooling system. I was just planning to include the character effect since I was planning to reuse the same effect.

If you want to create a one-off particle player Action for the StateMachine to use on the plant and maybe recycle on other NPCs, that's ok.

But if the particles belong to the specific character, I honestly don't see why pool them... sure, we might have 10-20 instances of plants in a level. That's not a big deal. We should only pool things that we plan to instantiate/destroy all the time. That is the real cost [in Unity], not having 10 particle systems that are not currently playing.

Plus, the particles played by the plant and character will probably be different in the end. Then we'd need to dismantle the system or split it into 2 pools, one of which will be a 1-element pool with just the player's particles... not ideal.

treivize:
Ok, thanks to have spent time explaining why it is not needed, I was not really clear about is costly in Unity about particle system. 
I will close this PR, since it is not adding real value. At least it helps me to better understand Pool/Factory system and what is important regarding particle system cost :)

PR #264, opened 2020-12-11
Outline improvements - New layers

treivize:
Forum thread: [Outline effect](https://forum.unity.com/threads/outline-effect.980034/)
Codecks card: [Outline doubling bug](https://open.codecks.io/unity-open-project-1/decks/13-bugs/card/18w-outline-is-doubled-when-two-dynamic-objects-overlap)

Sorry this update required a good amount of modifications. It solves the problem of the outline width changing depending of the dynamic object overlapping but it also improves generally the outline algorithm. 

First it introduce a new renderer feature called Outline Thickness:
image
The idea is to draw all the objects with outlines on a masking texture from the camera viewpoint. This masking texture will be read in the outline HLSL code to solve the doubling outline problem but also improve the outline computation.

Note:
I have moved all the existing objects that might be dynamic to dedicated layers:
 - Plant Rock and Slime critters into a new Layer called Critters
 - All props prefabs into a new layer called Props
 - All items prefabs into a new layer called Items
 - All the characters into the existing Characters layer

image

From this mask a first Roberts cross sampling is made to determine if the outline algo has to be run or not. 
- If all the sampling are in a black area, not part of an outlined object, no outline needed just return 0
- If at least one is in black and at least one is in red area, this is a border between outlined object and background object not outlined, always draw an outline, return 1
- If all the sampling are in the outlined object area, run the algorithm but with a thickness divided by 2 to keep the consistency with border thickness which is half displayed already. (The opposite could be an option (start with a sampling twice the provided size for the first pass)

Next, the current outline algorithm has been improved to include the biasing of the depth sensibility based on the surface normal and the camera view point.

Indeed, the algo looks for depth/normal jump over pixels and compares it to a sensibility to decide if an outline has to be drawn or not. But the depth difference can be high without meaning that there is an edge. A flat surface like the surface of a table seen from a grazing angle will have a strong depth gradient but we do not want to cover it with outline. It is what happens today with the current version of the algo. Actually having a unique depth sensibility global to the whole image whatever the camera point of view is not correct. See article from Roystan here: [Outline Shader](https://roystan.net/articles/outline-shader.html)

```
// Thresholding with view direction. 
// Balance normal difference based on the camera direction for flat surface viewed from a grazing angle
float NdotV = 1 - dot(normalSamples[0], viewDir);
float normalThreshold01 = saturate((NdotV - DepthNormalSensitivity) / (1 - DepthNormalSensitivity));
float normalThreshold = normalThreshold01 * DepthNormalThresholdScale + 1;
float depthThreshold = (biasedThickness / DepthSensitivity) * depthSamples[0] * normalThreshold;
edgeDepth = edgeDepth > depthThreshold ? 1 : 0;
```

The result:
image

Note: I have only fine tuned the PigChef Material to be properly outlined. The others are just using the default value for the new inputs (Depth Normal Sensivity and Depth Normal Scale)


daneobyrd:
Small note: make sure to fix the typo in the renderer feature. **Outline Thickness** rather than **Outline Tickness**. 😛

ciro-unity:
Really nice work, but I do see a few issues with it. I posted [on the forum](https://forum.unity.com/threads/outline-effect.980034/page-2#post-6615586).

treivize:
To improve the border outline, outline render pass is now including depth data based on object position from the camera.
Thanks to this information, the border outline is displayed only on the object the closest to the camera on the part overlapping other outlined objects.

In addition, a simple but efficient fading logic has been added to reduce the outline thickness depending on the camera distance:
image
The idea is that the thickness width is reduced by one every 10 distance unit from the camera.

Finally, a small proposition to try to make the outline independent from the resolution:
image

ciro-unity:
I left [some feedback](https://forum.unity.com/threads/outline-effect.980034/page-2#post-6689860) on the forum thread!

treivize:
@ciro-unity, I have pushed new updates still based on the approach based on Depth/Normal computation. But I found a bug which was responsible to the doubling width of the outline in the inner part. It has been corrected and I have also added the some AA  based on Harry technique shared in the Discord chat! 

The result looks much better now, I hope it could be considered better than the current implementation. Note, the PR is also including a proposal to support outline fading with the distance and a scaling of the outline thickness based on the screen resolution (the screen height since it drives the resolution of the game). These scaling work well with the AA technique from Harry!

Here is the Shader Graph including the fading, screen resolution independence and AA technique:

image

Results:
image
image
image

ciro-unity:
Closing as per discussion on #316

PR #265, opened 2020-12-11
Making GameSceneSO more reliable

dotsquid:
This pull request is partially related to this open task https://open.codecks.io/unity-open-project-1/decks/15-code/card/197-editor-window-to-quickly-open-scenes which I started to discuss here https://forum.unity.com/threads/editor-window-to-quickly-open-scenes-task.1019986/

Original solution which is referencing the scenes by their names is error-prone. Consider a situation when the scene asset being referenced is renamed or moved to another folder. This will lead to a broken link between GameSceneSO and the actual scene asset and will require manual resolving.

My solution is more reliable since it uses the scene asset directly. When the GameSceneSO is serialized (for instance during entering the play mode, building or editing the asset in the Inspector) a scene path (which is used in runtime) is retrieved from the scene asset (which is used in editor) and serialized along the asset itself.

Beside that my solution does not need any custom editor which makes code easier to read and understand (since this project has an education purpose, among others).

I also used scene path instead of its name since this makes the code more robust as well since allows to have several scenes with the same name. E.g. it's now possible to have 'Assets/Scenes/Locations/Beach.unity' and 'Assets/TestScenes/Locations/Beach.unity' with no problem.


DSivtsov:
> This pull request is partially related to this open task https://open.codecks.io/unity-open-project-1/decks/15-code/card/197-editor-window-to-quickly-open-scenes which I started to discuss here https://forum.unity.com/threads/editor-window-to-quickly-open-scenes-task.1019986/
I think you must make correction to your current preamble to this PR ("Making GameSceneSO more reliable" it isn't the same as PR for "Editor window to quickly open scenes") - mixing different type of changes to one big change, not good idea.

dotsquid:
> I think you must make correction to your current preamble to this PR ("Making GameSceneSO more reliable" it isn't the same as PR for "Editor window to quickly open scenes") - mixing different type of changes to one big change, not good idea.

Thank you for your concerns. These changes are necessary for the further work with the mentioned task (e.g. using scene path instead of scene name). I decided to put them in a separate PR because:
1) they solve the distinct problem and thus are self-sufficing;
2) they make the next PR lighter and thus make the maintainer's life easier since it would require to check less changes.

DSivtsov:
> ...These changes are necessary for the further work with the mentioned task (e.g. using scene path instead of scene name). I decided to put them in a separate PR because:
> ....

Sorry for my not clear sentence.
Idea of your changes - It's clear (I'm not saw in details your code yet).

I mention regarding your current preamble to this PR.

**IMHO** (and is commonly did here) Before making PR (or in the one time with it) create the corresponding separate thread on forum to discuss your proposal, to give possibility to community (to peoples who created the current version of "Scene loading") to be involved in your changes. Because architecture (and code) for loading scene discussed many month ago. And It have different card from RoadMap & thread (as variant you can find & renew that thread) than thread to which you linked this PR in your preamble.

ciro-unity:
Sounds like a good idea 👍🏻 

Before I get to it, one preliminary question: did you verify it works in a build?

dotsquid:
> Before I get to it, one preliminary question: did you verify it works in a build?
Yep.

dotsquid:
Just FYI the merge conflict is due to the changes in 'Assets\Scripts\Editor\GameSceneSOEditor.cs' while I removed it since it's not necessary with my solution.

DSivtsov:
> Just FYI the merge conflict is due to the changes in 'Assets\Scripts\Editor\GameSceneSOEditor.cs' while I removed it since it's not necessary with my solution.

Before it the "GameSceneSO " was a special "high level data type", which created for "real scene files", the most technical/internal scenes" haven't  "GameSceneSO ". The list of "GameSceneSO " populated from Build setting. 
@ciro-unity @amel-unity 
IMHO The previous architecture of use the GameSceneSO was more elegant - GameSceneSO  was a "simplete data", 
Assets\Scripts\Editor\GameSceneSOEditor.cs - script to automate work with GameSceneSO in Inspector.
In proposal variant "GameSceneSO " is mix data & code, which include preprocessor command and realization interface ISerializationCallbackReceiver.
And aim of these changes "...  it's now possible to have 'Assets/Scenes/Locations/Beach.unity' and 'Assets/TestScenes/Locations/Beach.unity' with no problem..."
How often it demands? And why not use some prefix or postfix in scene name  (Beach.unity and Beach_Test.unity or Beach_Alfa.unity)?

dotsquid:
@DSivtsov your permanent desire to criticize my work blinds you: the main purpose of the proposed changes is to exclude the possibility to lose connection between ScriptableObject wrapper and the scene itself in case the scene asset is renamed or moved. I'm not aware of your development background, but I know from my experience that this problem is not a phony issue - relying on names and paths is error-prone.

DSivtsov:
> the main purpose of the proposed changes is to exclude the possibility to lose connection between ScriptableObject wrapper and the scene itself in case the scene asset is renamed or moved.

In [the first post](https://github.com/UnityTechnologies/open-project-1/pull/265#issue-537045104) you did write it. But It doesn't work in current code, because your "architecture of Loading Level" based on **scenePath** value which stored in "GameSceneSO". 

When you move/rename SceneAsset corresponding "GameSceneSO" doesn't update, because your script which set value of **scenePath** work on event ISerializationCallbackReceiver.OnBeforeSerialize(), which occur when you change/view the corresponding Object/Instance. The corresponding "GameSceneSO" will have correct link to  SceneAsset (fileID: / guid:), but old data in field **scenePath**.

What is worse (It's a "Unity feature") - It'is, if you simple open in Inspector the corresponding "GameSceneSO" you will see the correct value (updated if you change it) name of AssetScene, but **scenePath** will have old data in "GameSceneSO".

And the most worst :) if you "unhide" the **public string scenePath**, You will see .... the new value of  **scenePath**, BUT if you simple after that close or continue work with other object ... the "GameSceneSO" values store in asset left unchanged.
You can update the **scenePath** value by current script, if you open "GameSceneSO" in Inspector and directly change AssetScene value in Inspector (open list, and select by double click).

The your Idea especial as you mention it in your last post - **"exclude the possibility to lose connection between ScriptableObject wrapper and the scene itself in case the scene asset is renamed or moved."** is very good.
But demands to change code to store the **scenePath** or replace it on Property/Method to receive the correct value before use by calling AssetDatabase.GetAssetPath(...) on actual Scene Asset.

And I agree with you that standard way for creation of "GameSceneSO"  through "Build Settings" is not good.
**But** the full remove of old script "GameSceneSOEditor.cs - script to automate work with GameSceneSO in Inspector." also not good.
More convenient way to have two variant of "work/creation of GameSceneSO": your variant and add to have additional check "in build setting" to exclude the issues that Game Designers put in build Scenes, the Scene which doesn't have correct GameSceneSO ("link build setting" and "list GameSceneSO").

dotsquid:
> When you move/rename SceneAsset corresponding "GameSceneSO" doesn't update, because your script which set value of scenePath work on event ISerializationCallbackReceiver.OnBeforeSerialize()

Yes, indeed it did not retrieve the scene path from the scene asset when play mode is entered. Somehow I missed this issue. Probably when I was testing this functionality that GameSceneSO was open in Inspector and that's the reason why it was serialized. Before that I thought ISerializationCallbackReceiver.OnBeforeSerialize() is called when Editor switches to play mode as it does when it builds the project.
Either way now it's fixed with a few lines of code.
Thanks for heads up.

DSivtsov:
> > When you move/rename SceneAsset corresponding "GameSceneSO" doesn't update, because your script which set value of scenePath work on event ISerializationCallbackReceiver.OnBeforeSerialize()
> 
> Yes, indeed it did not retrieve the scene path from the scene asset when play mode is entered. Somehow I missed this issue. Probably when I was testing this functionality that GameSceneSO was open in Inspector and that's the reason why it was serialized. Before that I thought ISerializationCallbackReceiver.OnBeforeSerialize() is called when Editor switches to play mode as it does when it builds the project.
> Either way now it's fixed with a few lines of code.
> Thanks for heads up.

:) You will check me or you will check your code before commitment :)

`if (prevSceneAsset != sceneAsset)` from `private void PopulateScenePath()` won't work if we rename or move Scene Asset because the Scene Asset wasn't changed - changed his "properties path".
Your script begin work if OnEnable() will be changed as following:

````c#
private void OnEnable()
	{
		prevSceneAsset = null;
		PopulateScenePath();
	}
````
And I can say more - he will work, but not correctly, because this is Unity ScriptableObject/Serialization features :) :
The correct value will have the scenePath variable which exist only in Runtime memory, if you check the corresponding SO file on disk, you will find old value :) (it's the great "Undo Unity" :).
If you want update any existen SO (also the same to serialized objects) on disk you must at first mark it as dirty. In this case it can be work this way:
````c#
private void OnEnable()
	{
		prevSceneAsset = null;
		PopulateScenePath();
		EditorUtility.SetDirty(this);
		AssetDatabase.SaveAssets();
	}
````
Again, I (IMHO) propose doesn't store the scenePath in "GameSceneSO" absolutely, store only Scene Asset, the scenePath  Get by `AssetDatabase.GetAssetPath(sceneAsset)` in corresponding scripts - LoadingLevel and so on?

OFFTOP:
shine always shine everywhere
that's my motto and sunshine @ciro-unity @amel-unity

dotsquid:
@DSivtsov sorry, but your last comment misses the target.

> if (prevSceneAsset != sceneAsset) from private void PopulateScenePath() won't work if we rename or move Scene Asset because the Scene Asset wasn't changed - changed his "properties path".

It does work: 'prevSceneAsset' is not serialized and thus has a default value (which is null) when the domain is reloaded, e.g. when the editor enters the play mode.

> The correct value will have the scenePath variable which exist only in Runtime memory, if you check the corresponding SO file on disk, you will find old value :) (it's the great "Undo Unity" :).

I have no idea what _the great "Undo Unity"_ means, but there is no need to save the those changes on disk when we are entering play mode. It's only required when we are making a project build and it's done when ISerializationCallbackReceiver.OnBeforeSerialize() is invoked by the build pipeline.

So to make things clear: we need 'scenePath' only when we are going to play the game. There are two cases of that:

1. When we are building the project. In this case we have to retrieve 'scenePath' and store it in the game bundle. That's done by ISerializationCallbackReceiver.OnBeforeSerialize()
2. When we are entering the play mode in the Editor. That's done by GameSceneSO.OnEnable()

In other situations we actually don't care what value is there.

> Again, I (IMHO) propose doesn't store the scenePath in "GameSceneSO" absolutely, store only Scene Asset, the scenePath Get by AssetDatabase.GetAssetPath(sceneAsset) in corresponding scripts - LoadingLevel and so on?

You may be surprised, but I did all these changes not to make things more complicated, but the other way round. The problem is that SceneAsset belongs to UnityEditor namespace and does not exist in runtime.

ThomasChen1220:
I pulled and reviewed the changes. 
I think @dotsquid  is on the right track because SceneAsset is giving us the ability to reference scenes directly rather than having to always save the path string. 
However, this PR is currently not working as expected. 
What I think is the desired behavior: 
1. when I set the sceneAsset to be scene1.unity, scenePath updates to the correct path
2. when I move scene1.unity to another folder, scenePath updates to the correct path
3. when I rename scene1.unity to scene2.unity, scenePath updates to the correct path

Currently, only 1 is achieved. I checked this by commenting out the [HideInInspector] (line 13) in GameSceneSO.cs. 
This happened because this condition: "if (prevSceneAsset != sceneAsset)" won't pass in the latter two cases. 

If we don't do the check, it would appear that it also works for 2 and 3. However, it's still not working under the hood. 
Here I have a logger that just prints out the scenePath every frame.
image
I moved the Beach1.unity to another folder. However, the update here didn't happen as soon as I moved the scene but when I opened the SO file in the inspector. This might be a nasty bug to find so I'm glad we noticed it here. 
I guess this happened because PopulateScenePath() is called only when we inspect the SO?

@DSivtsov Yes, if we can just get rid of scenePath entirely it should work. If it does, I'd say it's better than the previous solution. Right now I'm not sure if it will because SceneAsset appeared to be an editor thing? I can test it out. 

If not, then we have to somehow detect the OnSceneMoved and OnSceneRenamed events.

dotsquid:
@ThomasChen1220 
Hi. As I explained in my previous comment, we actually don't need to update 'scenePath' immediately after the path to that scene changed. We need a correct 'scenePath' only when we are going to switch to the game itself: either when entering play mode in Editor or when making the build.
[There is an API](https://docs.unity3d.com/ScriptReference/AssetModificationProcessor.html) to detect modification of assets. However it only works when the modification is performed inside the Editor. It won't help if that happens outside. And that's a quite possible scenario: e.g. reverting the changes in VCS.
So I prefer to update 'scenePath' as late as possible, close to the moment when it's going to be used.

DSivtsov:
> It does work: 'prevSceneAsset' is not serialized and thus has a default value (which is null) when the domain is reloaded, e.g. when the editor enters the play mode.

But if you change two times? I wrote it based on error in test (I simple move & rename Scene asset doesn't touch the SO) - the scene path was contain old value & can't load scene.

"great "Undo Unity" - It's mean, that Unity doesn't update SO self - give possibility to make undo to any changes of edited SO.


> there is no need to save those changes on disk 

Yes indeed, But it's not good when SO on disk contain value which doesn't corresponds with other.
And I doesn't understand the reason to store the scenePath? If we can use the "based object" - scene asset & get from it the value of path?

dotsquid:
> But if you change two times? I wrote it based on error in test
I can't imagine how the number of times the asset is renamed can affect this. Can you describe how to reproduce this error step by step?

> And I doesn't understand the reason to store the scenePath? If we can use the "based object" - scene asset & get from it the value of path?
>The problem is that SceneAsset belongs to UnityEditor namespace and does not exist in runtime.

DSivtsov:
> > But if you change two times? I wrote it based on error in test

UPDATE - Good news and bad news:
Good news - You are right
about 	//EditorUtility.SetDirty(this);
		//AssetDatabase.SaveAssets();
It doesn't demands
- If made the real build for game it correctly update scenepath
- If tun play mode in Editor also scenepath updated (but assent will contain old data, but it not create the issues)

regarding 
>if (prevSceneAsset != sceneAsset) from private void PopulateScenePath() won't work if we rename or move Scene Asset 
> because the Scene Asset wasn't changed - changed his "properties path".
>Your script begin work if OnEnable() will be changed as following:
You not right w/o these it doesn't work correctly [ see video](https://yadi.sk/d/nC3XZtx0j6wa3Q) - I also include current code & turn some debug message
````c#
private void OnEnable()
	{
		prevSceneAsset = null;
		PopulateScenePath();
	}
````

dotsquid:
@DSivtsov 
Hmm.. that's weird.
Maybe you have not default 'reload domain' options?
image

Anyway, I guess it's not the problem to add one line of code to make everyone happy.

DSivtsov:
> Maybe you have not default 'reload domain' options?
> image

FYI I try to set these options (it was unchecked) change nothing - the issues was the same. With last commit all work right, unrelated to values of Enter Play Mode Options

amel-unity:
As discussed on the forum, this is indeed a fantastic contribution! Thanks a lot @dotsquid. This will allow us to rely less on the error-prone solution based on names and make GamesSceneSo more reliable as you said. I also want to thank you all for this great discussion that helped to achieve this final result. I will merge this PR now

PR #266, opened 2020-12-13
Create custom base editor for custom inspector

rainaldisatria:
Proposition from Ciro in: https://github.com/UnityTechnologies/open-project-1/pull/258
Create base editor class that contain handy methods to use. 
Make things easier and no more writing repetitive code.

For now, it only contain:
- Draw Non Editable Script Reference()
image
image

This PR already implemented to some editor that needed it (no reference to the script):
- GameSceneSO
- AnimatorParameterActionSO

there are actually more:
- TransitionTableEditor
- StateEditor

But for some reason, it can't inherit from the CustomBaseEditor class. Maybe because they are using internal?


ciro-unity:
Amazing, thanks @rainaldisatria. Sorry for being late on this!

PR #268, opened 2020-12-16
Cards 195 and 197: Multiple Entry Exit Points and Editor Window To Ch…

georgefnix:
…ange Scenes(and Locations)

As discussed in:
https://forum.unity.com/threads/allow-multiple-entry-exit-points-in-locations-card-195.1019641/#post-6614431

I plan to make either a post with images or a video about this.

New:
1. Database to hold scene data. (Located in Resources/Database)
2. LocationSelection with custom property drawer.
3. Editor Window to load locations held in database during runtime. Can be found under Window/Designer Tools/Scene Selector
4. Locations are now places in scenes(Ex: Camp is a place in testing grounds). As such scene data scriptable objects accept location scriptable objects.
5. A tool called Location Editor can be added to objects in scenes to help designers fine tune locations visually instead of editing the scriptable objects directly. Beach, Forest, and Testing Grounds has some examples of this under Locations(blank scene object).

Changes:
1. Unified scene data into a single type.
2. SceneLoader loads a single scene.
3. Spawn position is communicated through a scriptable object.
4. Some renaming because the system changed.
5. Some moving of files because the system changed.

I hope I didn't forget anything significant.


ErikBehar:
suggestions:

* need to sign the CLA
* remove all the extra files that are not related to this commit its already a large change  
* don't like that it forces to always load from menu on start, can it be changed to allow load of whatever scene is currently loaded ? I know you can use the window, but saves time if you can just git play in the scene you are working on

georgefnix:
Thanks for the response. I was asleep, so my reply is delayed.
1, Thanks, done.
2. Good idea, The scenes have updated objects and some demonstration pieces, but you are right I should remove those and just do updates to freshly pulled scenes separately.(With exeception to the Initialization Scene)
3. The only dependency this has on the Initialization scene is LocationLoader(which existed prior to my work). That is a requirement for loading to work and I think is intended design. If I misinterpreted what you are saying then I am sorry. If are referring to how the menu is loaded first on play that can be changed in the LocationLoader by selected a different scene under "Main Menu Scene",

ErikBehar:
> Thanks for the response. I was asleep, so my reply is delayed.
> 1, Thanks, done.
> 2. Good idea, The scenes have updated objects and some demonstration pieces, but you are right I should remove those and just do updates to freshly pulled scenes separately.(With exeception to the Initialization Scene)
> 3. The only dependency this has on the Initialization scene is LocationLoader(which existed prior to my work). That is a requirement for loading to work and I think is intended design. If I misinterpreted what you are saying then I am sorry. If are referring to how the menu is loaded first on play that can be changed in the LocationLoader by selected a different scene under "Main Menu Scene",

Yes I'm referring to how the menu scene is loaded first on play... could you automatically set it to the current open scene instead ?

Issue #270, opened 2020-12-17
Build Fail Due To SceneAccessTool and SceneAccessHolderSO Errors


davejrodriguez:
**Short description**
Building a player fails with errors caused by SceneAccessTool and SceneAccessHolderSO.

**Expected behaviour**
I'd expect this editor tool to be put in an `Editor` folder so it does not get included with the build and cause errors when it can't find its UnityEditor types :)

**To Reproduce**
1. Try to build player.
2. Frown when it fails due to errors with the SceneAccessTool and its related SceneAccessHolderSO.

**Notes**
Some sample errors (they all have to do with missing editor assemblies):
- `Assets\Scripts\EditorTools\SceneAccessTool.cs(4,7): error CS0246: The type or namespace name 'UnityEditorInternal' could not be found (are you missing a using directive or an assembly reference?)`
- `Assets\Scripts\SceneManagement\ScriptableObjects\SceneAccessHolderSO.cs(12,31): error CS0246: The type or namespace name 'PropertyDrawer' could not be found (are you missing a using directive or an assembly reference?)`

amel-unity:
Thanks for testing @davejrodriguez ! The PR from @ThomasChen1220 fixes the issue, I will merge it now

PR #274, opened 2020-12-18
Fix SceneAccessTool build fail issue

ThomasChen1220:
**[for Bugfix PRs]**  
https://github.com/UnityTechnologies/open-project-1/issues/270 

Separated the custom attribute to its own file. 
Moved editor scripts to the editor folder. 

Try to build the project again and it should work now.

PS: not sure why these commits are shown: 
image
They should already be in the main branch. Let me know if I need to fix it.


amel-unity:
Thanks @ThomasChen1220 for fixing the bug. I tested it and now it's possible to build the game again. We are not sure if we will keep this version of the tool though, please check out [the thread](https://forum.unity.com/threads/editor-window-to-open-scenes.1019986/#post-6686869) and let us know what you think. I will still merge this PR now so we can build again on main, thanks 👍

PR #275, opened 2020-12-18
Multiple Entry/Exit

georgefnix:
Thread: https://forum.unity.com/threads/allow-multiple-entry-exit-points-in-locations-card-195.1019641/#post-6634054

This is a much less disruptive version of my last solution. I tried to minimize how many things were changed. This is the exact opposite tact I took previously.

I didn't include any scenes, so scripts within those scenes need to be updated for this to work.


amel-unity:
Thank you for the PR, I will need to close this one because we just merged [another PR](https://github.com/UnityTechnologies/open-project-1/pull/284) for this task as mentioned [in the thread](https://forum.unity.com/threads/allow-multiple-entry-exit-points-in-locations-card-195.1019641/#post-6712816).

PR #276, opened 2020-12-19
Control size of landing particles based on fall distance

davidhenshaw:
https://forum.unity.com/threads/puff-of-dust-effects.1002936/#post-6638764
I picked up on this issue through the Codecks card shown below. The forum thread for the puff effects hadn't addressed the card so I got to work on the solution. 
image

**The Solution**
I made an overload of the existing method `PlayLandParticles` that takes an argument called "intensity". I then rigged the `PlayLandParticlesActionsSO` to call the method with the intensity argument. The intensity is calculated based on how far the player fell during the State it was in. Since that StateAction is only called during the Jump_Decending State, the fall distance is always positive (but I still took the absolute value for safety).

As brought up by @ciro-unity in the [forum thread](https://forum.unity.com/threads/puff-of-dust-effects.1002936/#post-6638764), the particle effect's MinMaxCurve is reset using a coroutine. This was a work-around since resetting the curve just after calling `_landingParticles.Play()` was playing the effect with the original curve, not the new one. Another reason this was done was to retain the particle effect's settings in the Unity Editor instead of overwriting them with every jump. 

**Example of a short fall**
short-fall-before
short-fall-before-particleSize
short-fall-after
short-fall-after-particleSize

Just after the particle effect is finished playing, the `startSize` returns back to how it was set in the Unity Editor before the fall.
short-fall-way after-particleSize

**Example of a high fall**
high-fall-before
high-fall-before-particleSize
high-fall-after
high-fall-after-particleSize

The fall is so high that the intensity is 1. So the startSize maximum value is unchanged.

**Conclusion**
Please let me know if there is anything else I can do to improve this solution. Also this is my first ever PR, so please let me know how I could improve for future reference. Thank you very much in advance!

### DIFF
--- a/UOP1_Project/Assets/Prefabs/Characters/PigChef.prefab
+++ b/UOP1_Project/Assets/Prefabs/Characters/PigChef.prefab
@@ -320,6 +320,7 @@ MonoBehaviour:
   extraActionInput: 0
   movementInput: {x: 0, y: 0, z: 0}
   movementVector: {x: 0, y: 0, z: 0}
+  isRunning: 0
 --- !u!114 &6243045328629046901
 MonoBehaviour:
   m_ObjectHideFlags: 0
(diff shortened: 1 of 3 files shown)


BarthaSzabolcs:
Hi David!
I wanted to do this as my first task too, I thought it would be nice, if the dust cloud's size would be controlled by an animation graph, instead increasing linearly.

Since am not going to do the task I thought I will leave this suggestion here. :D

davidhenshaw:
> I thought it would be nice, if the dust cloud's size would be controlled by an animation graph, instead increasing linearly.

Hello @BarthaSzabolcs! Thanks for the feedback. The way I wrote the solutions doesn't necessarily change the size of all particles directly. The particle system still picks random sizes between the minimum and maximum value (the two values outlined in red). All my script really does is lower the maximum size of particles if the fall is short.

ciro-unity:
Thanks @davidhenshaw, great first contribution! Clean and self-contained.
I'm not super sure we need to reset the particles' size after the jump, since they belong to the character so they are overridden on the next landing. But for now it's ok. I've just made a note in the code after merging it.

PR #277, opened 2020-12-20
ReplaceTool selection and sibling index fix

neon-age:
That's it.


amel-unity:
Thanks for the fix @neon-age. Also, please keep in mind to add a description next time to describe what you are fixing especially that we don't have an open issue regarding this. I have noticed that the replace tool didn't keep the order of the object in the hierarchy when we use it and after testing with this fix, it now keeps the order. I will merge it now 👍

PR #282, opened 2020-12-28
Added new Accessory: Unibrow

BarthaSzabolcs:
forum:
https://forum.unity.com/threads/accessories-for-the-townsfolk.1007900/page-2

This is a new accessory, introducing more variety to the Townsfolk.
The only change is the added model and prefab, this was of course necessary.

The asset was intended to be similar to the mustache asset.
The texturing was done with creating an UV map, that will fit to the material that the mustache uses.

Preview image:
image


ciro-unity:
Perfect, thanks! Love that you recycled the texture and material for the moustache!

Issue #279, opened 2020-12-24
Camera clips tree crowns


davejrodriguez:
**Short description**
Even with the camera collider Cinemachine extension, the camera clips certain objects, notably trees.

**Expected behaviour**
I'd expect the camera to avoid clipping into the tree crowns OR gracefully fade objects out as the object approaches the camera's near clip plane. Since we're already using the camera collider extension, it might just be best (and easiest) to add sphere colliders to the crowns. 

**To Reproduce**
1. Play TestingGround
2. Walk around until there is a tree between the camera and Hamlet.
3. Use the player and camera controls to slowly maneuver the camera into the tree crowns.

![image](https://user-images.githubusercontent.com/1964335/103049621-47e06f80-4560-11eb-95ed-b5e63bb1bd06.png)

PR #283, opened 2020-12-28
Fixed colliders on the trees

Hey-Mikelangelo:
Added child GO to the trees named "leaves", added colliders to the leaves and set layer of leaves to "TreeLeaves" (So, also added this new layer. Set in physics properties to not collide with anything). In CameraSystem/FreeLook_VCam in "Cinemachine collider" added "TreeLeaves" layer in the field "Collide Against".

This PR is linked to the issue "Camera clips tree crowns #279" 

image
image
image
image
image
image


ciro-unity:
I like the solution with a new layer of leaves only, which obstruct the camera but don't collide. Good idea :)

PR #284, opened 2020-12-29
Allow multiple entry/exit points in Locations

rcarag:
Forum: <a href="https://forum.unity.com/threads/allow-multiple-entry-exit-points-in-locations-card-195.1019641/#post-6651727">Allow multiple entry/exit points in Locations (Card 195)</a>
Codecks: <a href="https://open.codecks.io/unity-open-project-1/decks/15-code/card/195-allow-multiple-entry-exit-points-in-locations">Card $195</a>

This PR adds a way to have multiple entry/exit points in Locations.

I looked at @georgefnix's solution and wanted to try an approach that avoids modifying the Protagonist class. In this approach, the SpawnSystem and LocationExit are modified instead. This way, the Protagonist class remains focused on its responsibility of handling player input.

How to set it up: https://imgur.com/a/kEiDOZm

The implementation was done by creating a PathSO scriptable object and a PathTaken runtime anchor. PathSO assets represent connecting paths between a LocationExit and LocationEntrance. They are basically empty scriptable objects with useful names, ex. "BetweenBeachAndForest", "BetweenBeachAndTestingGround", etc.

When LocationExit is triggered, it sets PathTaken to the PathSO assigned to it. The SpawnSystem checks the PathTaken, and searches its list of spawn locations for the first LocationEntrance with a matching PathSO. If a match is found, SpawnSystem will spawn at the matching location. Otherwise, SpawnSystem will fallback to the default spawn location.

In general, to connect a LocationExit to a LocationEntrance, just make sure they have a matching PathSO assigned to them. But it's also possible to have multiple LocationExits all leading up to the same LocationEntrance (ex. a maze, or where several roads lead to the same place). In the future, it may be possible for a game/level controller to dynamically change paths (ex. the player makes progress in a quest then comes home to find the village is burning).


amel-unity:
Thank you for opening this PR! I will merge it now as mentioned [in the thread ](https://forum.unity.com/threads/allow-multiple-entry-exit-points-in-locations-card-195.1019641/#post-6712816)

PR #285, opened 2020-12-29
Scene Selector window

dotsquid:
This PR introduces a yet another solution of this task https://open.codecks.io/unity-open-project-1/decks/15-code/card/197-editor-window-to-quickly-open-scenes which is being discussed here https://forum.unity.com/threads/editor-window-to-open-scenes.1019986/

It uses GameSceneSO's (modified in this PR https://github.com/UnityTechnologies/open-project-1/pull/265) to populate the list of the scenes to display.

Advantages:
1. Using GameSceneSO contrary to retrieving the scenes from the BuildSettings allows displaying the testing scenes which should not appear in the BuildSettings list. Also it does not require any manual actions when the scenes are created / duplicated / deleted. All changes to GameSceneSO's are automatically reflected in SceneSelector window.
2. Has a Preferences window which allows to reorder the scenes, modify their visibility and set a color marker to distinguish different types of scenes.
3. Preferences are stored in EditorPrefs. Contrary to storing the data in Resources this way is more cleaner since does not generate any redundant files in the project, simplifying the work with version control systems and keeping the final build smaller.
4. Correctly proposes to save the current scene prior to switching to another one.


amel-unity:
Thanks a lot for making the tool @dotsquid, I will merge the PR now as [I mentioned in the thread](https://forum.unity.com/threads/editor-window-to-open-scenes.1019986/#post-6713191).

PR #292, opened 2021-01-06
Plant Critter Behavior and Combat System applied to it

treivize:
[Forum thread](https://forum.unity.com/threads/combat-system.1032685/)
[Codecks card](https://open.codecks.io/unity-open-project-1/decks/15-code/card/19j-combat)

This PR derived from PR [Added initial plant critter behavior](https://github.com/UnityTechnologies/open-project-1/pull/241) 

Summary of changes:
- Added attack input tracking into Protagonist script to be used in State Machine (trigger animation and enable weapon during attack state)
- Added a simple Weapon script holding weapon strength and enabling state. Defined Frying Pan as a demo weapon 
- Added Critter script to hold critter instance states (Get hit/Killed/Max Health/Current health/Alert state/Fighting state and also keeping track of weapon hit to decrease current health and update the Get hit state.
- Added SM updates for Pig Chef and Plant Critter

What is draft/not done: 
- Weapon association to the player (here the Frying Pan is manually attached to the prefab at the right place in the Rig...)
- Maybe introduce SO to store Weapon/Critter general properties
- Critter attack on the player is not implemented

To test it, just add a Plant Critter in a test scene and run the scene. 

[Video demo](https://youtu.be/RhDWuYIlwm4)

Note: Attack is triggered when pressing J key in keyboard ;)


ciro-unity:
> Note: Attack is triggered when pressing J key in keyboard ;)

Small note before I forget: I changed the attack to F and added the LMB.

I changed F because I moved the alternative camera controls to IJKL. Even though almost nobody is going to use them, they can still be useful to have for people with movement disabilities or, for the same reason, to be remapped on accessible controls using stuff like the Xbox Adaptive Controller.

PR #294, opened 2021-01-07
General Map Whiteboxing_1

itsLevi0sa:
Added a General Map folder that contains the General Map whiteboxing scene.


ciro-unity:
Hey @itsLevi0sa, seems like you need to sign the CLA above, just to confirm that what you are providing is your creation.

amel-unity:
Thanks a lot for making the general map Whiteboxing @itsLevi0sa ! This is really super helpful to finally be able to visualize the different locations connected together 👍  I will merge it now

PR #298, opened 2021-01-11
Corrected plant critter state machine to be aligned with animator

treivize:
Sometimes when giving the last hit to the plant critter during the plant attack, it happens that the plant is killed without playing the expected BeenHit & Faint animations.

I have just figured out what is going wrong, the attack state in plant critter state machine has transition to dying state, so it happens that the state machine is ahead compare to the animator so some animations are missed. I am removing these transitions (Attack to Dying and Fighting to Dying, and adding BeenHit to Dying).


ciro-unity:
Yeah I noticed it, thanks for the quick fix!

PR #299, opened 2021-01-11
Added Critter SO - Item drop when killed - FR critter names

treivize:

The aim of this PR is to add the item dropping to the recently merged combat system. Basically when the plant critter is killed, some collectable items are dropped based on the configuration set up in a new Critter SO attached to the Critter MB script:

This SO will describe a variant of a critter setup (Localized name, Max HP, Max number of items it can drop, the drop rate table associating items to a drop rate (between 0.0 and 1.0)

image

I took this opportunity to add a Critter Name localization table, I tried to find funny critter names in FR ;) 
image

Finally in Critter script, the Critter SO is read / used to instantiate the drop result when the critter is killed:

image

A short demo of this enhancement: [here](https://youtu.be/eAR_zczE3pI)


amel-unity:
Great addition @treivize ! Thanks a lot, I will merge this PR now. I also love the critter names in FR :)

PR #300, opened 2021-01-11
Art assets - Oven

danten66:
https://forum.unity.com/threads/regarding-art-assets-and-music.980892/
An Oven is something that is needed to bake bread - and pizza!
These changes are not nessicary I suppose but they do add more flavor to the environment.
An FBX model and png Albedo Texture made into a prefab variant and material.

This is my first Pull Request and it looks like I may have added a backpack file which does not belong but I don't know.


ciro-unity:
Thanks, merging! Looks great!

Unfortunately the psd for the basket is in the PR. You put it in a folder above the project, but the folder was still tracked so you pushed it to the repo. Luckily it was under 100MB. I deleted it now.

So **CAREFUL**: if you pull, you will lose the .psd. Make a backup first! (by just copying it out of the project folder)
In any case, I copied it to my desktop in case you were to delete it.

PR #301, opened 2021-01-11
Added support to multiple overlapping interactions

treivize:
The aim of this PR is to propose an enhancement of the current Interaction Manager to be able to support multiple interactions at the same time when the player is close to several interactable elements (like pickup-able items)

In addition, the ResetInteraction() method has been fixed to avoid removing the on going interactions when any collider exit detection is triggered (like the weapon with the interaction manager). Only the interactable objects stacked during the OnTriggerEnter() is checked during the OnTriggerExit().


UOP1_Project/Assets/Scripts/Interaction/InteractionManager.cs
```diff
@@ -109,44 +106,63 @@ void OnInteractionButtonPress()
...
+		{
+			_toggleInteractionUI.RaiseEvent(true, _ongoingInteractions.Peek().Type);
 
-		if (_toggleInteractionUI != null)
-			_toggleInteractionUI.RaiseEvent(false, _potentialInteraction);
+		}
+		else
+		{
+			_toggleInteractionUI.RaiseEvent(false, InteractionType.None);
+		}
```
oddEventHorizon, line 166: How about if instead of creating new stack, you'd iterate over _ongoingInteractions and remove whatever interaction we've found. If the interaction we've found was the most recent one (the one we are currently interacting with) AND stack is not empty you call DisplayInteractionUI, othherwise _toggleInteractionUI.RaiseEvent(false, InteractionType.None);

This should be more optimal and slightly cleaner as you reuse DisplayInteraction method.

treivize: Actually, I did some searches and it seems not possible to delete an item in the middle of a stack in C# without having to create a new one. I am not "fluent" in C# so if you have a working proposal I will be happy to integrate it.

For the last part of your comment, it is done on purpose, when an interactable item has been consumed (pick up item), the collider is removed so we have to show the InteractionUI for the next one. Next indeed, in case of random collider exit (not related to interactable objects), the current implementation will inverse the interaction order and so propose another interaction. Not sure it is a problem, since all are valid, most of the case it might be the same type, the other cases might be rare (ie: pickup-able item close to an NPC and the player playing the attack action which will trigger a random collider exit with the cane)

treivize: I have updated the implementation to use a LinkedList instead of a Stack. It provides naturally remove item from the collection and simple accessor to get/remove first item.

amel-unity:
Thanks a lot @treiviz! The PR fixes the issue, I will merge it :)

PR #311, opened 2021-01-15
UI Fade in and out code and implementation

ZombieNub:
This implements the UI fader to black card using events and scriptable objects. To use, add a FadeChannelSO to the relevant script, assign the scriptable object, the call either Fade, FadeIn, or FadeOut. Has parameters for duration and target color. This is usable anywhere and at any time, but contains no safeguards besides preventing multiple fades from happening at the same time.

Note: This modifies the PersistentManagers scene and adds an Image UI element that it depends on.

Forum Thread: [https://forum.unity.com/threads/ui-fader-to-black.1038013/](https://forum.unity.com/threads/ui-fader-to-black.1038013/)

I would like to thank forum user [Swerik ](https://forum.unity.com/members/swerik.5425601/)for proposing the implementation plan I ended up using.

### DIFF
--- a/UOP1_Project/Assets/Scripts/Events/ScriptableObjects/UI/FadeChannelSO.cs
+++ b/UOP1_Project/Assets/Scripts/Events/ScriptableObjects/UI/FadeChannelSO.cs
@@ -13,21 +13,12 @@ public class FadeChannelSO : ScriptableObject
 	/// <param name="fadeIn">If true, the rectangle fades in. If false, the rectangle fades out.</param>
 	/// <param name="duration">How long it takes to the image to fade in/out.</param>
 	/// <param name="color">Target color for the image to reach. Disregarded when fading out.</param>
-	public void Fade(bool fadeIn, float duration, Color color)
+	public void Fade(bool fadeIn, float duration, Color color = default)
 	{
+		if (color == default && fadeIn) // If no fadein color is assigned, black is given as default. If we are supposed to fadeout the rectangle, default is simply passed through.
+			color = Color.black;
 		if (OnEventRaised != null)
 			OnEventRaised.Invoke(fadeIn, duration, color);
-	}
-
-	/// <summary>
-	/// Generic fade function. Communicates with <seealso cref="FadeManager.cs"/>.
-	/// </summary>
-	/// <param name="fadeIn">If true, the rectangle fades in. If false, the rectangle fades out.</param>
-	/// <param name="duration">How long it takes to the image to fade in/out.</param>
-	public void Fade(bool fadeIn, float duration)
-	{
-		if (OnEventRaised != null)
-			OnEventRaised.Invoke(fadeIn, duration, new Color(0, 0, 0, 1));
 	}
 
 	/// <summary>
@@ -35,24 +26,18 @@ public class FadeChannelSO : ScriptableObject
 	/// </summary>
 	/// <param name="duration">How long it takes to the image to fade in.</param>
 	/// <param name="color">Target color for the image to reach.</param>
-	public void FadeIn(float duration, Color color)
+	public void FadeIn(float duration, Color color = default)
 	{
+		if (color == default)
+			color = Color.black;
 		Fade(true, duration, color);
-	}
-	/// <summary>
-	/// Fade helper function to simplify usage. Fades in the rectangle.
-	/// </summary>
-	/// <param name="duration">How long it takes to the image to fade in.</param>
-	public void FadeIn(float duration)
-	{
-		Fade(true, duration, new Color(0, 0, 0, 1));
 	}
 	/// <summary>
 	/// Fade helper function to simplify usage. Fades out the rectangle.
 	/// </summary>
 	/// <param name="duration">How long it takes to the image to fade out.</param>
-	public void FadeOut(float duration)
+	public void FadeOut(float duration, Color color = default)
 	{
-		Fade(false, duration, new Color(0, 0, 0, 1));
+		Fade(false, duration, color);
 	}
 }
--- a/UOP1_Project/Assets/Scripts/Events/ScriptableObjects/UI/FadeChannelSO.cs
+++ b/UOP1_Project/Assets/Scripts/Events/ScriptableObjects/UI/FadeChannelSO.cs
@@ -1,0 +1,58 @@
+﻿using System.Collections;
+using System.Collections.Generic;
+using UnityEngine;
+using UnityEngine.Events;
+
+[CreateAssetMenu(menuName = "Events/UI/Fade Channel")]
+public class FadeChannelSO : ScriptableObject
+{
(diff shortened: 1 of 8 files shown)


UOP1_Project/Assets/Scripts/Events/ScriptableObjects/UI/FadeChannelSO.cs
```diff
@@ -0,0 +1,58 @@
...
+	/// <summary>
+	/// Generic fade function. Communicates with <seealso cref="FadeManager.cs"/>.
+	/// </summary>
+	/// <param name="fadeIn">If true, the rectangle fades in. If false, the rectangle fades out.</param>
+	/// <param name="duration">How long it takes to the image to fade in/out.</param>
+	public void Fade(bool fadeIn, float duration)
+	{
+		if (OnEventRaised != null)
+			OnEventRaised.Invoke(fadeIn, duration, new Color(0, 0, 0, 1));
+	}
```
TheSwerik, line 31: You can give parameters a default value so that you don't have to write the same method multiple times.
Here an example:
```suggestion
	public void Fade(bool fadeIn, float duration, Color color = default(Color))
	{
		if (color == default(Color)) color = Color.black;
		if (OnEventRaised != null)
			OnEventRaised.Invoke(fadeIn, duration, color);
	}
```
This would work for the FadeIn-Methods as well

FYI: default(Color) == new Color(0, 0, 0, 0) and that's the only problem: If you want to fade in to or out from transparent but I don't see how that makes any sense so I would ignore that

UOP1_Project/Assets/Scripts/UI/FadeManager.cs
```diff
@@ -0,0 +1,101 @@
...
+		while (totalTime <= duration)
+		{
+			totalTime += Time.deltaTime;
+			_imageComponent.color = Color.Lerp(oldColor, new Color(0, 0, 0, 0), totalTime / duration); // Sets the image's color to a mixture between the old color and  total transparency, and interpolates based on the amount of time to completion.
+			yield return null;
+		}
+		_imageComponent.color = new Color(0, 0, 0, 0); // Here to guarentee the image is fully transparent at the end of the loop.
+		IsCurrentlyFading = false;
+		yield return null;
+	}
```
TheSwerik, line 56: Here I said [in the Thread](https://forum.unity.com/threads/ui-fader-to-black.1038013/#post-6728329) that I would like to keep the ability to choose a FadeOut starting color

you could do this with a default value so that if you don't put anything as a parameter it automatically chooses the old color:
```suggestion
	private IEnumerator FadeOutEnum(float duration, Color color = default)
	{
		if (color == default) color = _imageComponent.color; // Stores the old color of the image component, as we can't assume the image will always be black, if no color is specified.
		float totalTime = 0f; // Total amount of time this coroutine has taken. Determines when the fadeout will end and what color the imageComponent should be at every frame.
		while (totalTime <= duration)
		{
			totalTime += Time.deltaTime;
			_imageComponent.color = Color.Lerp(color, new Color(0, 0, 0, 0), totalTime / duration); // Sets the image's color to a mixture between the old color and  total transparency, and interpolates based on the amount of time to completion.
			yield return null;
		}
		_imageComponent.color = new Color(0, 0, 0, 0); // Here to guarentee the image is fully transparent at the end of the loop.
		IsCurrentlyFading = false;
		yield return null;
	}
```

TheSwerik: You would need to change it in the FadeChannelSO.cs as well so that You can choose a FadeOut starting color there as well.

TheSwerik: and you'd need to change it in this file line 96 `StartCoroutine(FadeOutEnum(duration));` aswell

ZombieNub: Your two comments make sense. I'll implement the changes and update the PR.

UOP1_Project/Assets/Scripts/UI/FadeManager.cs
```diff
@@ -0,0 +1,101 @@
...
+
+public class FadeManager : MonoBehaviour
+{
+	[Header("Fadeout Event")]
+	[SerializeField] private FadeChannelSO _fadeChannelSO = default;
+
+	[Header("Fadeout Canvas")]
+	public GameObject _fadeoutCanvas = default;
+
+	private Image _imageComponent;
```
TheSwerik, line 14: why grab the canvas when you could just grab the image itself?
```suggestion
	[SerializeField] private Image imageComponent;
```
Just drag the FadeOut Rectangle in to get the image
then you can also remove the Start() method

UOP1_Project/Assets/Scripts/UI/FadeManager.cs
```diff
@@ -0,0 +1,101 @@
...
+using System.Collections.Generic;
+using UnityEngine;
+using UnityEngine.UI;
+
+public class FadeManager : MonoBehaviour
+{
+	[Header("Fadeout Event")]
+	[SerializeField] private FadeChannelSO _fadeChannelSO = default;
+
+	[Header("Fadeout Canvas")]
```
TheSwerik, line 11: Wouldn't just `Fade` be a more appropriate name?
```suggestion
	[Header("Fade Event")]
	[SerializeField] private FadeChannelSO _fadeChannelSO;

	[Header("Fade Canvas")]
```
also change the `[Header("Fade Canvas")]` to `[Header("Fade Image")]` if you apply my other suggestion

TheSwerik: I removed `= default;` because my IDE (Jetbrains Rider) told me this was redundant
A variable will always be set to the default value so `= default` only makes sense if you want to give a function-parameter a default value

ZombieNub: > 
> 
> Wouldn't just `Fade` be a more appropriate name?
> 
> also change the `[Header("Fade Canvas")]` to `[Header("Fade Image")]` if you apply my other suggestion

Will do.

TheSwerik (approved):
I don't see anything wrong here
Great implementation imo

ZombieNub:
> 
> 
> I don't see anything wrong here
> Great implementation imo

:)

ciro-unity:
Thanks for collaborating on this :)
I like it, I just have a couple of comments which I'm going to fix before merging (no action needed from you, just tell me if I misunderstood something and my changes will break it)

For instance, in the manager you have two functions which are 99% identical (FadeIn and FadeOut), so I unified them.

I'll merge for now. We will need to test it, but also, implement a mechanism so that the scene loading actually doesn't begin until the first fade out has fully finished.
So basically: Disable gameplay controls > Request fade out > Begin scene loading > Scene loading finishes > Begin fade in > Enable gameplay again

PR #315, opened 2021-01-20
Slime critter propelling attack movement

treivize:
Forum thread: [Slime Critter](https://forum.unity.com/threads/slime-critter.1005712)
Codecks card: [Critters](https://open.codecks.io/unity-open-project-1/decks/65-game-design-wiki/card/14z-critters)

The aim of this PR is to propose an implementation for the propelling effect during the Slime Critter attack.
This animation might be specific to this critter so a new MB script has been created (Slime Critter Attack Controller) which controls the movement of the critter during the attack.

image

Here is the result in the game:
SlimeCritterPropel3

### DIFF
--- a/UOP1_Project/Assets/Art/Characters/SlimeCritter/Animation/Attack.anim
+++ b/UOP1_Project/Assets/Art/Characters/SlimeCritter/Animation/Attack.anim
@@ -1,0 +1,11623 @@
+%YAML 1.1
+%TAG !u! tag:unity3d.com,2011:
+--- !u!74 &7400000
+AnimationClip:
+  m_ObjectHideFlags: 0
+  m_CorrespondingSourceObject: {fileID: 0}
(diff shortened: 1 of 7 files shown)


amel-unity:
Hey @treivize ! Thanks for opening this PR! I needed to attach the SlimeCritterAttackController to the SlimeCritter_Base prefab to see it working. So I was wondering if you created a prefab variant from the base prefab, to which you already added the script. In this case can you push it? or if the script is on the base prefab then maybe you need to apply changes after the script is added and then push the prefab again (I can also do it, just let me know).

treivize:
Hey @amel-unity ! My bad, I noticed that the prefab was part of the changed log of my PR so I assumed it was because of the new script attached to it, but I guess I have edit it through the scene and forget about saving it to propagate the change in the prefab.

I have just corrected it.

amel-unity:
Works great now, thank you! I will merge it :)

PR #316, opened 2021-01-20
Added AA, Distance Fading and Screen Resolution independency to Outline

treivize:
Forum thread: [Outline effect](https://forum.unity.com/threads/outline-effect.980034/)
Codecks card: [Make outline effect less resolution dependent](https://open.codecks.io/unity-open-project-1/decks/16-3d-shading/card/18u-make-outline-effect-less-resolution-dependent)

The proposed implementation is the following:

image

**Distance fading:**
- Outline Thickness is reduced based on the distance of an outlined object from the camera. Reduced by 1 every 10 world dimension unit

**Screen Resolution adjustment:**
- Outline thickness is scaled up or down based on the screen height size. The material Outline Thickness property will be the one for a HD resolution (1080px height).

**Antialiasing algo:**
- Based on Harry solution shared in Discord
image
  The main idea is to run the outline function twice with different thickness value and to lerp both outputs.

**NOTE**: The materials of outlined object might have to be adjusted following this implementation if they were not defined with the targeted HD resolution.


ciro-unity:
I've been testing it and it seems pretty good! Will take another, fresher look tomorrow.

ciro-unity:
Sorry I'm a bit lost: how does this PR relate to #264? Are they intended to be together, are you hoping to merge the other one first, or are they trying to do the same thing?

Also I think I already tried the other one (#264) in the editor and saw some problems, but now I can't recall if it was that one (we had many outline ones). If so, I'm happy to close it, and maybe merge this one.

treivize:
Sorry, indeed the attempts to make the outline nicer was a mess...

This one is independent from the previous one (#264). I guess we can close #264 for now, maybe later I will come back with a better solution to solve the doubling of the outline when outlined objects are overlapping (it was the main intent from the previous PR).

ciro-unity:
Lol, I didn't intend to merge this one yet... but I did, by mistake 😆 
Never pull a PR into `main` locally and forget to revert, I guess. Now let's hope it was a good change!! 🤣 🤣

ciro-unity:
(I'm saying it laughing because I think it's a good change, otherwise I'd revert it)

PR #319, opened 2021-01-22
Combat system refactor

treivize:
Codecks card: [Combat](https://open.codecks.io/unity-open-project-1/decks/15-code/card/19j-combat)

I am really sorry for this huge PR... I might have split some of the features in future PRs... (at least GetHit Effect, Player SM updates to support GetHit and Fainting states)

Here is the summary of the changes made:
- Split into multiple SO CritterSO
image
  - Attack (including the attack reload moved from SM here)
image

  - Dropping
image
The logic has been changed to control more precisely the number of drops per kill. Each drop has a drop rate and a list of items that can be dropped with a rate per item. So we can define for a kill a common drop item (100% of the cases) with 50% chance to be an apple or 50% to be a banana. An additional uncommon drop item (20% of the cases) with 100% chance to be a smoothie.

  - Roaming Strategy
image

  - Chasing Strategy
image

  - Get Hit Effect (Bonus ;))
image

- Move the logic in each of them in association with a local data holder class (I am not sure how conventional this way of dealing with global properties from SO and local properties for each GameObject)
- HealthAnchor for the player health anchor 
- Add GetHit state to PigChef
- Add Dying state to PigChef
- Add Critter Weapon to PlantCritter


treivize:
Hi, I hope you did not started to loose too much time reviewing this PR, because I am planning to close it. I released during the night that most of the changes should be managed through FSM and not badly wired as here between SO scripts and Critter script.
From my point of view, roaming, chasing, "GetHit" flashing effect, should be moved to FSM actions/conditions.

Still some parts should be fine, like the attack and the adding of player getting hit by critter. 

I will open new PRs soon with isolated parts. 

Next, I am still considering having separated SO to store the different config of the behavior and not being part directly to the action/conditions of FSSM, it will allow having variant of the behavior without having to duplicate the full FSM table and created new actions and conditions.
 
A good example is about the attack config (strength and reload time) having them out of the FSM makes the same FSM working for different enemy strength.

Issue #321, opened 2021-01-24
Water rendered way higher than it should


ciro-unity:
**Short description**
Water gets rendered way higher than it should.

**Expected behaviour**
Waves shouldn't go higher than the value specified in the Inspector (Wave Height) in any case.

**To Reproduce**
1. Go to an empty scene and put a great plane with the Water material.
2. Walk around (see video)
3. You will eventually find a spot where the water waves are above the camera, even though the Wave Height might be 0.

**Notes**
- See video for details. Observe water plane and character position on the Y, and the material parameters in the Inspector:

https://youtu.be/yi_yDTd8vZg

ciro-unity:
cc @treivize: you might want to see this :D

treivize:
Oh oh... Definitely!

ciro-unity:
By the way it's related to the Global Fading Distance, because when you bring it to lower values the water stays lower, if you put the value higher you will see the water raise above the camera any time.

treivize:
@ciro-unity, I have opened a PR to fix this issue, just maybe one comment on the setup of the water plane from your video and how I can see in the Beach scene, the scaling on Y should be kept to 1, otherwise it will be used as a factor for the Water height parameter.

PR #324, opened 2021-01-25
Issue #321 - Fixed water height not working as expected

treivize:
Fix for issue #321.

There was a mistake in the water shader graph, the wave height was attached to the noise scale which is not scaling the height of the water but the noise density... 
So the height was always height 1, and was only affected by global fading to be reduced.

I mainly just multiply the output of the noise by the Wave Height parameter to fix the issue:
image

Result: Height 0.5
image

Result: Height 3.0
image


ciro-unity:
Thanks for the quick fix! Seems to be working, and now we can see the "sea bed" too through the water. Before it would go completely blue, for some reason.

ciro-unity:
Hmm, nope, unfortunately if you put the fading distance high (to 45 for instance) the water will still rise way above its intended level:

image

Sure, I can lower it, but we should look into why that even happens in the first place...

treivize:
Just, what is the scale on the Y axis? is it 1 or 100?

ciro-unity:
Yes, it was 100 🤦🏻‍♂️ 🤦🏻‍♀️ Sorry! Seems to be working now!

PR #326, opened 2021-01-25
Combat System - Added player getting hit and killed states

treivize:
Codecks card: [Combat](https://open.codecks.io/unity-open-project-1/decks/15-code/card/19j-combat)

Another PR to add the impact of critter attack to the player.
Mainly adding new FSM states/conditions/actions.

Noticeable feature:
 - Adding armful part to slime and plant critters
 - Temporary player respawn action of the player after having been killed by raising SceneOnLoad event


treivize:
Hi @amel-unity, the PR is updated following the merge of the big PR on Combat refactor! Up to you and @ciro-unity to review it!

AlexandreGheraibia:
> 
> 
> Hi @amel-unity, the PR is updated following the merge of the big PR on Combat refactor! Up to you and @ciro-unity to review it!
> 
> There is a noticeable hack here:
> 
> ```
> 	public override void OnStateEnter()
> 	{
> 		// Hack to force Collider Exit to be triggered before destroying an object.
> 		_gameObject.transform.position += - Vector3.up * 1000;
> 		GameObject.Destroy(_gameObject, 0.1f);
> 	}
> ```
> 
> I did that to force the object to trigger a collider exit before being destroyed. I do not know if there is a better solution...
perhaps :
_gameObject.GetComponent<Collider>().enabled=false;
from : https://docs.unity3d.com/ScriptReference/Collider-enabled.html

treivize:
Hi @AlexandreGheraibia,
I tried it without success, apparently disabling Collider or even the whole GameObject is not triggering OnTriggerExit as per design.
I found this old post in Unity forum with actually recent reply from Unity and it seems to be an old pb with no built-in solution.

The hack I am using is one of the solution suggested. Someone implemented a cleaner solution, but will rely on his script (ReliableOnTriggerExit), not sure it brings added value compare to the hack proposed.

treivize:
Thanks for the merging! And all the updates made. Keep tracking the target and check it in FSM to get rid off the OnTriggerExit not raised when GameObject destroyed is a good idea :)

PR #327, opened 2021-01-26
Ground shader - using vertex colors and polybrush

daneobyrd:
[Forum Thread](https://forum.unity.com/threads/ground-shader.1041673/)
[Codecks Card](https://open.codecks.io/unity-open-project-1/decks/16-3d-shading/card/19k-shader-for-ground)
This PR adds texture painting meshes using vertex colors and Polybrush. 
My shader graph supports up to 4 textures, three using the RGB channels, and a base texture assigned to no vertex color (black). I use a noise texture as a transition between ground textures; this can be iterated on further.

Use:
Select the third tab in the Polybrush window (marked by an isosceles triangle) and select black, red, green, or blue to paint the corresponding texture.
Screenshot 2021-01-25 182022

The shader also has a debug view, displaying black, and RGB colors in place of the textures assigned in the inspector. The debug view includes a channel mixer property, allowing the user to control the visibility of R, G, and B vertex colors independently.
image

All progress updates and iteration thus far can be found in the linked forum thread.

[Demo video](https://youtu.be/G70FY-uGpxk)

Note: I moved the outline outside of the main ToonShading graph because the ground shader is not outlined. I only did this because I received errors about exceeding the number of keywords/variants when I tried to use the internal lighting subgraphs.


ciro-unity:
Looks great, merging!!

ciro-unity:
Two small things:
- Try not to use spaces in file names, they can result in issues with scripts (spaces need to be escaped in strings)
- Example scenes go into Scenes/Examples (and you can put the assets in them too) so later we can easily delete them

ciro-unity:
I'm also renaming the textures with the colour they use.

Issue #333, opened 2021-01-28
SpawnSystem no longer works as the default spawn location


rcarag:
The recent modifications in SpawnSystem to allow multiple entry/exit points seems to have broken the ability to use SpawnSystem on its own as a default spawn location.

The problem happens because the SpawnSystem's transform is not included in the results of `GameObject.FindGameObjectsWithTag("SpawnLocation")` in the `SpawnPlayer()` method. As a result, when a scene without other spawn locations is played directly in the editor (ex. TestingGround_Small), the SpawnSystem will fail to find a spawn location, and the player will not be spawned.

I'm creating a PR to workaround this issue.

AlexandreGheraibia:
Hi, 
I looked in the spawnSystem file and saw this call :
FindSpawnIndex (_pathTaken? .Path ?? null) why not just put FindSpawnIndex (_pathTaken? .Path)
According to stackoverflow, pathTaken? .Path means that if PathTaken == null then return null else return pathTaken.Path. 
 and A ?? B means if A! = null then A otherwise B.

ciro-unity:
Is this even still broken? I am playing the build from `main`, and I can spawn fine in any scene, including "Empty_Location" (which is like the simplest playable scene we have).

ciro-unity:
I think we decided that this issue is no longer valid. Closing!

PR #334, opened 2021-01-28
Bugfix: SpawnSystem not working as default spawn location

rcarag:
This PR is to address the issue described here: [SpawnSystem no longer works as the default spawn location #333](https://github.com/UnityTechnologies/open-project-1/issues/333)

This is a continuation to the changes introduced here: [Allow multiple entry/exit points in Locations](https://forum.unity.com/threads/allow-multiple-entry-exit-points-in-locations.1019641/#post-6712816)

The fix works by assigning the SpawnLocation tag to the SpawnSystem prefab, so it can use itself as the default spawn location (ex. when PathTaken is null, or when no LocationEntrance matches the PathTaken).

This can be verified by testing a scene that doesn't contain other spawn locations (ex. TestingGround_Small) and playing it directly within the editor. For scenes that contain other spawn locations (ex. Beach), the player will now spawn at the SpawnSystem's location when it is played directly from the editor (instead of a random spawn location).

This PR also contains a new Exit prefab to simplify the creation of exit points.


AlexandreGheraibia:
Hi,
I looked in the spawnSystem file and saw this call :
FindSpawnIndex (_pathTaken? .Path ?? null) why not just put FindSpawnIndex (_pathTaken? .Path)
According to stackoverflow, pathTaken? .Path means that if PathTaken == null then return null else return pathTaken.Path and A ?? B means if A! = null then A otherwise B.

Another thing , the AutoFill function exclude the SpawnSystem's transform

ciro-unity:
Is this even still broken? I am playing the build from `main`, and I can spawn fine in any scene, including "Empty_Location" (which is like the simplest playable scene we have).

ciro-unity:
Closing as the related issue seems to be invalid. Thanks anyway!

PR #336, opened 2021-01-28
Add a Stop and Finish option for the AudioManager

island212:
Forum thread: None
Codecks card: [Audio: Stop vs. FinishAndStop](https://open.codecks.io/unity-open-project-1/decks/15-code/card/19f-audio-stop-vs-finishandstop)

Added the function to control the audio that has been requested to be played
Did some modification in the AudioManager and AudioEventChannel

Noticeable feature:
- You get an AudioCueKey when you request to play audio
- You can use this key to either stop or finish the emitters from the previous request.
- Stop end the audio playback instantly and finish stop the looping

[Video](https://streamable.com/0ozroq)

I chose to use a key to control the emitter to avoid using a emitter that has been put back into the pool. The alternative was to let the user do the pooling. I find the first option cleaner. In any case, can't wait to hear your opinion.


ciro-unity:
Fantastic work! I'm integrating it.

Issue #297, opened 2021-01-10
Input is cached when not immediately consumed by a State


ciro-unity:
**Short description**
Input is cached when not immediately consumed by a State in the state machine, and then used when the next state "interested" in that input is reached.

**Expected behaviour**
The input should simply be ignored, if not related to the state the machine is currently in.

**To Reproduce**
1. Open a gameplay scene, press Play.
2. Jump and while in the air, press the button to talk (called "ExtraAction" as of now).
3. Observe the character doesn't talk in the air (as expected).
4. Observe that, once landed, the character talks, even if the ExtraAction button is no longer pressed.

**Notes**
- The functionality described above to talk is a temporary one, but the issue might happen with any action that doesn't interrupt the walk.

ciro-unity:
This also happens when attacking, and pressing the Attack input while the first attack is still happening.

treivize:
Finally, I have the feeling that there are not lots of input which will create a problematic behavior, so what about consuming quickly the input in Protagonist script after activation like this?

![image](https://user-images.githubusercontent.com/42570903/104169394-cada4480-53ff-11eb-9f41-96a839896f19.png)

It is a bit hacky but it seems good enough for the problem we want to solve the problem here.

ciro-unity:
Hmm, it does feel kinda hacky, but it's also kind of safe because all it does is reset a bool. If we go that route, I'd unify it and make it a constant on all actions that need to be instantaneous (I guess all of them).

DeivSky:
Someone mentioned in the forums about doing it in `LateUpdate`, I think that would be the simplest solution.
``` cs
private void LateUpdate()
{
    attackInput = false;
    talkInput = false;
    otherInput = false;
}
```
Another option I just came up with, and I think the least hacky one -because it's listening to what we actually care about in the state we're in, instead of randomly setting things to false,- would be to make an **Action** to dynamically subscribe and unsubscribe from events.

Here is the way I imagine it: 

- We modify `InputReader.cs` to invoke **EventChannelSOs** instead of UnityActions.  In turn, we'd also have to modify `Protagonist.cs` to take in and subscribe to **EvenChannelSOs**. 
- We write the new `BindEventChannelsActionSO.cs`. An **Action** that takes in two **EventChannelSOs**. In `OnStateEnter()` we subscribe raising one of the channels to the other, and in `OnStateExit()` we unsubscribe it. See where this is going? 
[...]

PR #338, opened 2021-01-31
Fix Input is cached when not immediately consumed by a State for jump…

AlexandreGheraibia:
Fix Input for jump and attack 

Issue : https://github.com/UnityTechnologies/open-project-1/issues/297
I added an explanation in the issue.
 
Card : https://open.codecks.io/unity-open-project-1/decks/13-bugs/card/18o-input-is-cached-when-not-immediately-consumed-by-a-state

**There is a bug in my solution for Attack
For fix it in the Transition from PreventAction_Attack to Attack** 
image
image

**There is another error for the  jump descending, I moved the PlayLandParticle Action
because I misinterpreted the Actions of states.**
New solution :
image

I have a new version pending which normally fix the bug.
I'm not comfortable with github so I don't know how to do the new PR or if this PR must be validated before to be able to make the new one

An idea for talk:
image

Demo attack and pick up :
Like you can see, there is no pick up animation when we attack and pickup
UOP1_Project - InteractionExample - PC, Mac   Linux Standalone - Unity 2019 4 18f1 Personal  PREVIEW PACKAGES IN USE  _DX11_ 2021-01-31 11-14-52_Trim

### DIFF
--- a/UOP1_Project/Assets/ScriptableObjects/StateMachine/Protagonist/PigChef_TransitionTable.asset
+++ b/UOP1_Project/Assets/ScriptableObjects/StateMachine/Protagonist/PigChef_TransitionTable.asset
@@ -50,22 +50,19 @@ MonoBehaviour:
       Condition: {fileID: 11400000, guid: 41dbff2af0b6f7141a32bb464309ba69, type: 2}
       Operator: 0
   - FromState: {fileID: 11400000, guid: c57ed1dc57a890943b2648ca8a581075, type: 2}
-    ToState: {fileID: 11400000, guid: 027d32476800b3543b2f5446a59054c8, type: 2}
+    ToState: {fileID: 11400000, guid: 390bddf5271c3024b8a375a0a62d0990, type: 2}
     Conditions:
-    - ExpectedResult: 0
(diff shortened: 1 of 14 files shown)


AlexandreGheraibia:
A question on the implementation of isHoldingAttackConditionSO :
why test _protagonistScript.attackInput in the Statement and not replace it by the following line :
```
protected override bool Statement()  => _protagonistScript.attackInput;
public override void OnStateExit()
{
		base.OnStateExit();
		_protagonistScript.attackInput = false;
}

ciro-unity:
I'm not super sure of what's the strategy you want to use to fix it... are you suggesting to add a new intermediate state between Jump and Idle, called PreventAction? So that when you press the button it goes to this state instead of going to (for instance) Talk or Attack?

If this is the case, I don't think it's a good idea. It feels like a hack and it could introduce issues later on. Not to mention that it's harder to maintain, because the more states we have the more transitions and edge cases we need to implement (to give you an example, we'd need a new transition from PreventAction to GetHit when you receive a damage while in the air and pressing the talk button?? terrible...)

AlexandreGheraibia:
I put a filter state in front of each state that should not receive an action.
for example we necessarily go through PreventAction_Attack before going to the attack state.

Issue #343, opened 2021-02-01
DialogueManager throw null reference exception if OpenDialogueUIEvent is not initialized.


rainaldisatria:
**Short description**
Just a minor bug that'll occur if DialogueLineEvent in DialogueManager is not initialized.

**Expected behaviour**
Should not throw null reference exception even if DialogueLineEvent is not initialized by UIManager.

**To Reproduce**
1. Go to Scene/Examples/
2. Click on Cutscenes scene
3. Press start button.
4. This (null reference error) happens since DialogueLineChannelSO is not initialized by UIManager.

ciro-unity:
Closing the issue because it's not happening anymore due to some null checks implemented in the script.

However, please note that the expected behaviour is misleading:
>Should not throw null reference exception even if DialogueLineEvent is not initialized by UIManager.

This is incorrect, the expected behaviour is that
> the DialogueLineEvent should be initialised

If not initialised, then it **should** throw a NullReferenceException - to warn the developer of an error in the setup.

PR #344, opened 2021-02-01
fix/DialogueManager throw null reference exception if OpenDialogueUIEvent is not initialized.

rainaldisatria:
Just a minor bug. 
But of course, safety  :D

Issue: https://github.com/UnityTechnologies/open-project-1/issues/343

I solved it by: Invoking the the channel's method ( RaiseEvent() ) rather than the event it self (OnEventRaised) in DialogueManager.cs

EDIT:
I also do the same thing in UIEventoryManager.cs

-----------------

To not make this issue happen again, do anyone know how to protect OnEventRaised to not be invoked outside of the EventChannels script but initialize-able by other scripts?


jeffreysmooth:
Hi,

We can make OnEventRaised as private, In order to stop invoking the OnEventRaised from outside instead of calling RaiseEvent.
and Create new public method for Registering a event. 

@rainaldisatria @ciro-unity

rainaldisatria:
Hey @jeffreysmooth , that is a good idea.
Do you want to make a PR to implement it for every channel?

I kinda busy at the moment.
You can use my branch or create new fresh start branch.

rainaldisatria:
I just checked that you already create a PR XD
Cool.

PR #345, opened 2021-02-02
Added Slime Critter droplets when moving

treivize:
Just a small graphical effect addition to the slime critter. When moving the critter let some slime droplet behind him.

image

The implementation is based on a particle system with mesh renderer using the same material as the critter. the droplets have a negative Y speed to make them fading out smoothly.


ciro-unity:
I merged them, but I switched them to be distance-based so they are not emitted if the Slime is still.
Also, I switched them to fade in size rather than Y position, it was visible that they were sinking in the ground when seen from close up.

treivize:
Indeed, it is much better now! I learnt something new on the particle system (distance based generation :) )

I am looking at the result of the massive merge of yesterday evening! Impressive :)

PR #347, opened 2021-02-03
Fixed white crispy outline in Play mode around object over water

treivize:
Playing with the Beach scene, I discovered an annoying white crispy outline around any objects rendered in front of the water plane.

Before the fix:
image

After the fix:
image

Solution:
Updated Shader graph of water shader to cut foam when depthFade is >= 1 (part selected added)
image


ciro-unity:
Crispy outline 🤣

treivize:
> Crispy outline 🤣

Glittering would have been more appropriated, but I did not find it back yesterday evening and a cooking word was funnier ;)

Issue #342, opened 2021-02-01
Interaction System doesn't follow the FSM


DeivSky:
**Short description**
Interactions don't happen along with their corresponding States (objects get picked up without the character being in the PickUp state, dialog starts without being in the Talking state, etc).

**Expected behaviour**
Drops get picked up when entering the PickUp state, dialogues start when entering the Talking state, cooking starts when entering the cooking state.

**To Reproduce**
1. Go to the InteractionExample scene, or kill any critter so it drops items.
2. Get close to an item on the ground.
3. Jump or Attack and then press the interact button
4. The item gets picked up while the character is mid-air/mid-attack, but the character never enters the PickUp state so it allows the player to skip the pick-up animation.

**Notes**
- Video of the bug:
<video src="https://user-images.githubusercontent.com/27371812/106408983-4be49480-641e-11eb-9a73-7399edc88bef.mp4" controls="controls" class="d-block rounded-bottom-2 width-fit" style="max-height:640px;"></video>

- Same thing goes for Talking and Cooking, the dialogue/cooking UI will pop up but the animations won't start since we don't get to transition into their states.

rainaldisatria:
This is it, the problem I'm facing when developing my hack and slash game with our State Machine.
It is really hard to maintain and sync two different state machine at the same time, so what I do is: 

Removing Unity's State Machine (Animator) completely.
![image](https://user-images.githubusercontent.com/57592497/106443715-64e95600-64af-11eb-807a-33ec5a643930.png)

Then, to get the ability to transition smoothly' through one animation to another is by using CrossFadeAlphaInFixedTime();
I make an Action for that called 'PlayAnimationActionSO' that'll play the animation OnStateEnter smoothly.
![image](https://user-images.githubusercontent.com/57592497/106444092-de814400-64af-11eb-8522-ab4ca35c1d64.png)
*ignore the error line there, I have slightly different code base.*

The only catch is that we need to manually assign the Transition Duration. 
Usually the transition duration is 0.25f tho.
![image](https://user-images.githubusercontent.com/57592497/106444199-0375b700-64b0-11eb-8b7b-7b675869e83c.png)

If you want to see the working example, you can check it out here:
https://www.youtube.com/watch?v=RouJPWyfP0Y
*(it also an open source project)*

DeivSky:
[...]

PR #348, opened 2021-02-03
Fix Interaction System doesn't follow the FSM

AlexandreGheraibia:
issue : https://github.com/UnityTechnologies/open-project-1/issues/342

**How did you resolve this issue?**  

In interactionManager, I divide the function OnInteractionButtonPress in two  OnInteractionButtonPress and OnInteractionResponseActivated. The last one is associated with a VoidEventChannel named InteractionResponse.

I also create an ActionScript and the associated action named ActiveInteractionResponse_OnEnter which raise the event InteractionResponse.

**How can it be verified that the issue has actually been resolved?**  

Try to pick up an object while you are jumping or attacking
Try to cook and see that you can't.


UOP1_Project/Assets/Scripts/Characters/StateMachine/Actions/ActiveInteractionResponse_OnEnterSO.cs
```diff
@@ -0,0 +1,32 @@
...
+{
+	protected new ActiveInteractionResponse_OnEnterSO OriginSO => (ActiveInteractionResponse_OnEnterSO)base.OriginSO;
+	
+	public override void Awake(StateMachine stateMachine)
+	{
+	}
+	
+	public override void OnUpdate()
+	{
+		OriginSO.interactResponse.RaiseEvent();
```
DeivSky, line 22: This should probably be in `OnStateEnter`

AlexandreGheraibia: Yes, this Should be o_O', Thank you

ciro-unity:
Hey @AlexandreGheraibia, I am merging this PR since I think the idea to break it into "before" and "after" is good. However, I am making some changes to clean up a lot of things (beyond this script).

I also think, if I'm not mistaken, that the PR as it is now breaks the cooking because Cooking is not a real state on the state machine, so nobody calls OnInteractionResponseActivated. But it's ok, I'm refactoring anyway! I'll explain more [on the forum](https://forum.unity.com/threads/world-interaction-system.1020709/unread).

AlexandreGheraibia:
> 
> 
> Hey @AlexandreGheraibia, I am merging this PR since I think the idea to break it into "before" and "after" is good. However, I am making some changes to clean up a lot of things (beyond this script).
> 
> I also think, if I'm not mistaken, that the PR as it is now breaks the cooking because Cooking is not a real state on the state machine, so nobody calls OnInteractionResponseActivated. But it's ok, I'm refactoring anyway! I'll explain more [on the forum](https://forum.unity.com/threads/world-interaction-system.1020709/unread).

Yes, you got it 🥇

PR #351, opened 2021-02-04
Whiteboxing of general map in medium size divided into 6 scenes 

Smurjo:
- contains minimalistic scene loading and unloading scripts only for trying out the whiteboxing
- contains small editor tool to create views. Don't delete deactivated objects in spawn scene if you want to update the views. 
- contains few additional art assets like shiitake mushrooms, big trees, bushes and creepers.


Smurjo:
Whiteboxing as shown in [https://youtu.be/xQrY5oHP1hE]

ciro-unity:
A few notes:
- Please don't put spaces in filenames
- You should try not to submit tons of graphic files with the whiteboxing. We're not taking the trees in at this point, they are very different in style from the game's graphics. Should I merge the PR and delete them? If I leave them in the project, people are going to use them.
- Same goes for some of the structures. I love the little umbrellas, but they feel too modern:

image

- The stage is nice. I think there are some issues with the normals not importing or being very flat, but we can use Unity's import settings to recalculate them so no huge deal. The white wood and the perfectly straight poles make it feel a bit too modern and realistic.
- The big tents are cool. Especially when lighting is baked:

image

- One thing though: please don't create complex patterns for the tents and umbrellas. I feel like people who dress like the Townsfolk and live in wooden/hay houses wouldn't decorate their tents with such fine and precise patterns. It's a matter of preference, sure, but to me they scream "modern" tent. I had a google (admittedly the only thing I could think of is "medieval fair tents") and [if you have a look around](https://www.google.com/search?q=medieval+fair+tents&sxsrf=ALeKk00yZg0AQkpMuAW-AKhcAFnJ3hZuog:1613338914456&source=lnms&tbm=isch&sa=X&ved=2ahUKEwjMx5fnq-ruAhVNxosKHd6KCIcQ_AUoAXoECBMQAw&biw=1920&bih=947), you'll see how they are often simple patterns, all white, or big bands of colour with white. I assume the white textile was easier/cheaper to produce, and the coloured one was for more rich people.

ciro-unity:
More notes in random order:
- I'm worried about rivers and waterfalls. For the rivers, we need some special version of the water shader, to simulate the flow of water. Otherwise they will feel very static. For waterfalls and springs, we need a good effect of water flowing out (maybe the river shader would do) and particles. And sound!
- The bridges on the map so far are very small, pretty different from the ones drawn in the Concept Art thread.
- Many areas feel very connected, not presenting the "prefab walls" (as I called them) which should provide a visual shield so that you can't see what's behind them, beyond the skybox. I can probably add them as I reconstruct the map. I need to go through and create the final ground meshes, and place a few rocks.

ciro-unity:
More notes:
- Place Prefabs in the prefab folder
- Remove the word "Variant" at the end of variants
- This looks cool:

image

- Remove the outline from materials for props and food. We only use it for things that move, like characters and tools.
- Try to capitalise names (as mentioned in the guidelines). This is actually quite important since if I try to fix capitalisation, we migth get "ghost" files since .git is not case sensitive. So I'd rename something but wouldn't be able to push the change unless I also add other letters or move the file. Can be quite tricky to understand what's going on at times.

Issue #340, opened 2021-01-31
Camera lerps into place at the beginning of a scene


ciro-unity:
**Short description**
The Cinemachine Vcam lerps into place when a scene is loaded.

**Expected behaviour**
While a small camera movement is ok, it needs to be intentional. As it is now, it's a bug that depends on the camera position at load time.

**To Reproduce**
1. Open the Beach scene. Press play.
2. Observe the camera moves a bit on entering play.
3. Turn around and enter the cave with the collider taking to the Glade. The level will load.
4. Observe the camera is pulled fast into place when the level loads.

**Notes**
- Here's a video of the issue:

https://user-images.githubusercontent.com/20049224/106399291-0de47200-6418-11eb-9cc0-e66cf3e87877.mp4

sergio-nnz:
Hi, I would like to fix this bug!

ciro-unity:
Feel free to give it a shot, @sergio-nnz! We don't "assign" tasks, but rather evaluate work that is done on them and pull it in if it solves the issue.

If you have a clear idea on how to fix it you can move forward and open a PR, or use this issue for comments, or if you want to bring the discussion to the forums feel free to do so by opening a new thread there.

vjs22334:
I think I have a fix. warping the vcam to the player after setting the target seems to work.
```
public void SetupProtagonistVirtualCamera(Transform target)
{
  freeLookVCam.Follow = target;
  freeLookVCam.LookAt = target;
  freeLookVCam.OnTargetObjectWarped(target,target.position-freeLookVCam.transform.position);
}
```


https://user-images.githubusercontent.com/30682201/106930528-a8f08b00-673b-11eb-817e-444cb1ef7476.mp4

I guess the distance can be adjusted a bit to warp the camera to a set distance away from the player.

Should I make a pr

PR #354, opened 2021-02-05
fix for camera Lerps into place

vjs22334:
fix for issue: #340 
link to forum thread: https://forum.unity.com/threads/camera-lerps-into-place-at-the-beginning-of-a-scene.1052282/

**How did you resolve this issue?** 
 After a new scene is loaded, the Camera Manager in the scene manager calls SetupProtagonistVirtualCamera() to set the target of the free look vcam. After setting the target, I have used OnTargetWarped() to warp the camera close the the player without lerping. I have also added a field in camera manager called _SpawnDistanceFromPlayer to set how far from the player the camera will warp. I found the most consistent results in terms of camera movement for a value of zero.

**How can it be verified that the issue has actually been resolved?**  
you may follow the steps to reproduce the bug mentioned in the issue, the vcam will now only move a small amount on location change

### DIFF
--- a/UOP1_Project/Assets/Scripts/CameraManager.cs
+++ b/UOP1_Project/Assets/Scripts/CameraManager.cs
@@ -11,11 +11,7 @@ public class CameraManager : MonoBehaviour
 	private bool _isRMBPressed;
 
 	[SerializeField, Range(.5f, 3f)]
-	private float _speedMultiplier = 1f; //TODO: make this modifiable in the game settings
-
-	//this is the distance from player the camera will warp to when a new scene is loaded. this can be set individually on each scene, zero is the most consistent
-	[SerializeField,Tooltip("the distance from player the camera will warp to when a new scene is loaded")]
-	private float _SpawnDistanceFromPlayer = 0f; 											
+	private float _speedMultiplier = 1f; //TODO: make this modifiable in the game settings											
 	[SerializeField] private TransformAnchor _cameraTransformAnchor = default;
 
 	[Header("Listening on channels")]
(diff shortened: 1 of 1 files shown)


AlexandreGheraibia (commented):
Hi, I don't understand the calculus of positionDelta.
In your code you calculate a difference between an existing distance and an arbitrary distance on the same vector. Why not just give an arbitrary distance ? 
Vector3 positionDelta =  (target.position.freeLookVCam.transform.position).normalized*_SpawnDistanceFromPlayer;

vjs22334:
Hi, the reason for that calculation is I wanted to move the camera to a point which is _SpawnDistanceFromPlayer away from player's position. So I calculated position delta as the vector between player and camera plus _SpawnDistanceFromPlayer  in the opposite direction. I cannot give a fixed arbitrary distance like: 
`Vector3 positionDelta = (target.position-freeLookVCam.transform.position).normalized*_SpawnDistanceFromPlayer;` 
as the distance between the camera and player and the camera is different in different scenes.
if we want the camera to just warp directly to the player I should simply go back to my previous commit which does:
`freeLookVCam.OnTargetObjectWarped(target,target.position-freeLookVCam.transform.position);`
with no option to move the camera a slight didtance away from player.

AlexandreGheraibia:
> 
> 
> Hi, the reason for that calculation is I wanted to move the camera to a point which is _SpawnDistanceFromPlayer away from player's position. So I calculated position delta as the vector between player and camera plus _SpawnDistanceFromPlayer in the opposite direction. I cannot give a fixed arbitrary distance like:
> `Vector3 positionDelta = (target.position-freeLookVCam.transform.position).normalized*_SpawnDistanceFromPlayer;`
> as the distance between the camera and player and the camera is different in different scenes.
> if we want the camera to just warp directly to the player I should simply go back to my previous commit which does:
> `freeLookVCam.OnTargetObjectWarped(target,target.position-freeLookVCam.transform.position);`
> with no option to move the camera a slight distance away from player.

Ok, you do like you want^^.
I meant you could get your result differently by get a proportion of the vector target.position-freeLookVCam.transform.position.

vjs22334:
I removed it. It was most consistent with _SpawnDistanceFromPlayer zero so might as well always move the camera directly to player position

ciro-unity:
Thanks @vjs22334, great contribution and a very elegant fix in one line 👍🏻

PR #355, opened 2021-02-08
To avoid developer to call "OnEventRaised" action instead of calling …

jeffreysmooth:
To avoid developer to call "OnEventRaised" action instead of calling RaiseEvent. This causing null pointer exception.

Also created new Class EventChannelGenericSO.cs to avoid repeatable code in event systems.

**[for Generic PRs]**  
1. Avoid repeatable code in Event systems if we want to create a new channel with new parameter type.
2. Created generic BaseEventGenericSO Class and moved RaiseEvent method inside to that class to achieve point 1.

**[for Bugfix PRs]**  
DialogueManager throw null reference exception if OpenDialogueUIEvent is not initialized
How did you resolve this issue?  make OnEventRaised as private and created public method to register

### DIFF
--- a/UOP1_Project/Assets/Scripts/Audio/AudioManager.cs
+++ b/UOP1_Project/Assets/Scripts/Audio/AudioManager.cs
@@ -83,16 +83,16 @@ public class AudioManager : MonoBehaviour
 
 	private void RegisterChannel(AudioCueEventChannelSO audioCueEventChannel)
 	{
-		audioCueEventChannel.OnAudioCuePlayRequested += PlayAudioCue;
-		audioCueEventChannel.OnAudioCueStopRequested += StopAudioCue;
-		audioCueEventChannel.OnAudioCueFinishRequested += FinishAudioCue;
+		audioCueEventChannel.RegisterAudioCuePlayEvent(PlayAudioCue);
+		audioCueEventChannel.RegisteAudioCueStopEvent(StopAudioCue);
+		audioCueEventChannel.RegisteAudioCueFinishEvent(FinishAudioCue);
 	}
 
 	private void UnregisterChannel(AudioCueEventChannelSO audioCueEventChannel)
 	{
-		audioCueEventChannel.OnAudioCuePlayRequested += PlayAudioCue;
-		audioCueEventChannel.OnAudioCueStopRequested += StopAudioCue;
-		audioCueEventChannel.OnAudioCueFinishRequested += FinishAudioCue;
+		audioCueEventChannel.UnregisterAudioCuePlayEvent(PlayAudioCue);
+		audioCueEventChannel.UnregisteAudioCueStopEvent(StopAudioCue);
+		audioCueEventChannel.UnregisteAudioCueFinishEvent(FinishAudioCue);
 	}
 
 	// Both MixerValueNormalized and NormalizedToMixerValue functions are used for easier transformations
--- a/UOP1_Project/Assets/Scripts/CameraManager.cs
+++ b/UOP1_Project/Assets/Scripts/CameraManager.cs
@@ -35,7 +35,7 @@ public class CameraManager : MonoBehaviour
 		inputReader.disableMouseControlCameraEvent += OnDisableMouseControlCamera;
 
 		if (_frameObjectChannel != null)
-			_frameObjectChannel.OnEventRaised += OnFrameObjectEvent;
+			_frameObjectChannel.RegisterEvent(OnFrameObjectEvent);
 
 		_cameraTransformAnchor.Transform = mainCamera.transform;
 	}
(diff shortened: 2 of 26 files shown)


rainaldisatria:
Will close this issue: https://github.com/UnityTechnologies/open-project-1/pull/344

ciro-unity:
I'm thinking: do we need all of this? It's a lot of extra code to get the same functionality.
Does it solve issue #343? Maybe. But **was that an issue?**

Honestly if there is an event channel property on a script and no SO referenced in it, I as a developer want to know that. So the error message was actually good. And it gives me all the info I need to find the culprit.

With this "fix", now event channels can fail silently. An object is meant to fire an event, and it doesn't. 😨 Now I can play the game and not realise that a certain sound effect is not playing or something is not happening in the game, because the system hides the missing connection. Not super good, no?

jeffreysmooth:
**Please find my inline comment.** 
I'm thinking: do we need all of this? It's a lot of extra code to get the same functionality.

**[Jeff: I think it is not extra code. The below code been used at most of the channel event. This has been modularized by creating new class called EventChannelGenericSO.cs]**

```
       public UnityAction<ActorSO> OnEventRaised;
	public void RaiseEvent(ActorSO actor)
	{
		if (OnEventRaised != null)
			OnEventRaised.Invoke(actor);
	}

ActionButtonClicked.OnEventRaised += ActionButtonEventRaised;
```
The above code change like this.

`public class DialogueEventChannelSO : EventChannelGenericSO <ActorSO>
{
}
ActionButtonClicked.RegisterEvent(ActionButtonEventRaised)
`



Does it solve issue #343? Maybe. But was that an issue?
**[Jeff: Yes it is an issue. Developer raising events directly using obj.OnEventRaised(data) instead of calling RaiseEvent method.]**

Honestly if there is an event channel property on a script and no SO referenced in it, I as a developer want to know that. So the error message was actually good. And it gives me all the info I need to find the culprit.
With this "fix", now event channels can fail silently. An object is meant to fire an event, and it doesn't. 😨 Now I can play the game and not realise that a certain sound effect is not playing or something is not happening in the game, because the system hides the missing connection. Not super good, no?

**[Jeff: I do not see any silent failure here. channel raise event and whoever subscribe/Register to that event will receive the event. May be we can add debug log if there is no subscriber for the event. Even now it happen if developer calls RaiseEvent correctly instead of calling like this obj.OnEventRaised(data) ]**

`if (OnEventRaised != null)
	OnEventRaised.Invoke(actor);
else
    Debug.log("No subscriber to catch event raised.")`

rainaldisatria:
I think the error message should be called in the EventChannel itself to say that no one is subscribing to that event yet,,,, just like jeffreysmooth said.

It's not an issue at all... just to respect encapsulation for other developers to use in the future.

ciro-unity:
Sorry, still don't understand the point of all these changes. You make private the Action, and expose two methods that do nothing more than subscribing to the Action. Why not keep the Action public then?

Second point: You made a generic class for channels with 1 parameter. But there's no generic base class for the EventChannel with 2 or 3 parameters. As you can see from the summary of the PR, this adds 152 lines and deletes 110, and you would still need to create the generic base classes.

Also, as I said before, the issue #343 is not a problem with the code, it's a setup issue. So this change wouldn't "fix" it. (in fact, I'll close it now)

So, if you want to create the generic EventChannel classes with all parameter configurations (1, 2, 3 at least) go on, otherwise I'll close this PR.

ciro-unity:
I finally, finally understood what we're trying to do here. Unfortunately issue #343 has a misleading description. The issue is **not** that the reference is missing, the issue is that the developer is invoking the OnEventRaised Action directly.
So this PR doesn't fix that issue.

So now, where are we.

Actually, to prevent the calling of `OnEventRaised`, can't we just add the `event` keyword to all actions? It saves us a lot of code (no need to register/unregister functions) and I like the `+=` `-=` syntax, when you see it in an `OnEnable` you know immediately what's going on.

So going from:
`public UnityAction<DialogueDataSO> OnEventRaised;`
to
`public event UnityAction<DialogueDataSO> OnEventRaised;`
This should prevent people from being able to call OnEventRaised directly from outside of the EventChannel that contains it.

Makes sense?

ciro-unity:
Hey @jeffreysmooth, can I close this PR and maybe implement the changes I suggested myself? Or you can also open a new one with those changes, if you prefer.

jeffreysmooth:
Hi, Sorry I was busy with my office work. Please close this PR. I will work on this when I get time..

ciro-unity:
> Hi, Sorry I was busy with my office work. Please close this PR. I will work on this when I get time..

Of course, no problem at all.

Thanks anyway!

PR #356, opened 2021-02-08
Added slash effect when player is triggering an attack

treivize:
A graphical addition to the combat system (prior to the addition of impact effect): The slash effect!

Slash2

It relies on a new effect controller script (should it be merged into a generic PlayerEffectController included dust effect anchors?):
image
Next, the trigger is based on animation clip event.


amel-unity:
Hi treivize, thanks a lot for creating the slash effect, looks great! 
Just added a commit to correct a typing error for the script name PlayerEffectController (instead of PlayerEffectContrioller) and yes it's good that you merged the effects into a generic PlayerEffectController 👍 
I will merge the PR now

treivize:
Oh! Merci beaucoup!

PR #361, opened 2021-02-11
Created a Shader Subgraph that applies Emission to a material using C…

Donkeylipsh:
…ustomEmission.hlsl, and added Emission to the Toon Shadergraph

**Shading Model - Emission Support**  
- [Forum Thread](https://forum.unity.com/threads/toon-shader.980874/page-2)
- This PR completes the "Emission support" task, by providing artists with the ability to add emission effects to their objects using the base Toon shader.
- [Codecks Task](https://open.codecks.io/unity-open-project-1/decks/16-3d-shading-animation/card/175-emission-support)
- Without these changes, the Toon shader is unable to support emission effects
- I created a Shader Subgraph that scales an input color, based on an emission flag, a texture mask to target emission, and an intensity value to scale the emission. If the flag and mask allow for emission, the pixel will have it's value scaled to a minimum of 1.5 (i.e. and input value of 0. (black) will have an output value of 1.5). Otherwise the color will output unchanged.
- I then inserted the Subgraph above into the Toon Shadergraph, just before the "Unlit Master" node
- To mask areas from emission, artists will need to set the color of the areas they don't want to emit to pure black (0., 0., 0.)

**Note/Future Improvement**
- The mask texture was designed to use grayscale values, but will also allow for artists to emit a different color than the surface material if they choose. If this is a desired feature, a slider to control the mix of surface color to emission color should be added to improve this PR


Donkeylipsh:
I've signed the contributor license multiple times, on multiple browsers, and multiple machines. Not sure what's going on here. Any ideas how to get this approved?

Eyap53:
Did you use the same email as the one you used in your commit ?

Donkeylipsh:
Unable to get CLA working. Resubmitting.

PR #362, opened 2021-02-12
Shading Model - Emission Support

Donkeylipsh:
### Shading Model - Emission Support
Created a Shader Subgraph that applies Emission to a material using CustomEmission.hlsl, and added Emission to the Toon Shadergraph

- [Forum Thread](https://forum.unity.com/threads/toon-shader.980874/page-2)
- This PR completes the "Emission support" task, by providing artists with the ability to add emission effects to their objects using the base Toon shader.
- [Codecks Task](https://open.codecks.io/unity-open-project-1/decks/16-3d-shading-animation/card/175-emission-support)
- Without these changes, the Toon shader is unable to support emission effects
- I created a Shader Subgraph that scales an input color, based on an emission flag, a texture mask to target emission, and an intensity value to scale the emission. If the flag and mask allow for emission, the pixel will have it's value scaled to a minimum of 1.5 (i.e. and input value of 0. (black) will have an output value of 1.5). Otherwise the color will output unchanged.
- I then inserted the Subgraph above into the Toon Shadergraph, just before the "Unlit Master" node
- To mask areas from emission, artists will need to set the color of the areas they don't want to emit to pure black (0., 0., 0.)

### Note/Future Improvement

- The mask texture was designed to use grayscale values, but will also allow for artists to emit a different color than the surface material if they choose. If this is a desired feature, a slider to control the mix of surface color to emission color should be added to improve this PR


ciro-unity:
Hey @Donkeylipsh! Welcome to the project, I think this is your first contribution, right? Cool.

Out of curiosity, why did you go with custom HLSL code? That looks like stuff we could do with Shadergraph. Plus, you're using an `if` statement which, as far as I know, is less than optimal in shaders. The funny thing is that if you were to port the same logic to Shadergraph, the resulting code wouldn't have an `if` because SG would use a multiplication instead, optimising it greatly.

Would you be able to convert it into a graph?

Donkeylipsh:
Hi @ciro-unity! This is my first contribution, and actually my first contribution to anything that wasn't part of a class project. Which pretty much explains all the decisions I made, they were rookie mistakes. I went with the custom HLSL simply because I'm new to Unity and was more comfortable in code than I was working with all the wires and nodes. Thanks for the tip on the `if` statements in Shaders, that'll help optimizing all my future work.

I'll get this converted to Shadergraph and get rid of that pesky `if` statement tonight and resubmit my PR.

FYI, my username on the Unity forums in **cdvano**, so you can take a look at the screenshot I posted to see if this is working as you expected.

Thanks for all your help and advice!

Donkeylipsh:
@ciro-unity Everything has been converted to Shadergraph and works as expected. 

image

ciro-unity:
Hahhaa what a carnival!! 😆 
Thanks a lot @Donkeylipsh. Will pull it as soon as I verify the implementation.

Donkeylipsh:
Sounds good. I wasn't sure if I should include the mask textures I used for testing in the PR, so I left those out. Here's a couple masks you can use if you wanted to save some time during testing.

bard_hare_Emission
PhoenixChick_Emission
Pig_Emission
PlantCritter_Emission
RockCritter_Emission
SlimeCritter_Emission
townsfolk_F_Emission
townsfolk_M_Emission

ciro-unity:
I ended up simplifying the graph a bit, no need for minimum emission and the other parameter. I reduced it to the texture (which provides the colour and the difference within the object) and 1 float, for multiplication. All is added on top of the base colour.

The reason is that when somebody uses the shader, they won't understand immediately what "minimum" does, so setup now is ultra simple for everyone. Thanks anyway, merging!

Donkeylipsh:
Sounds good. Thanks @ciro-unity !

PR #363, opened 2021-02-12
Combat System - Added reloading to player attack - Single and Double strike 

treivize:
Forum: [Combat System](https://forum.unity.com/threads/combat-system.1032685/)
Codeck card: [Implement second cane hit animation](https://open.codecks.io/unity-open-project-1/decks/16-3d-shading-animation/card/19v-implement-second-canehit-animation)

To make the player less angry when chaining the attack, here is a proposition to insert reloading time between two attack request from the player.

Here is the behaviour:
- Player performs just one click -> Single strike -> Reloading (1s)
- Player perform a long click -> Double strike -> Reloading (1s)

In between, there is an attack reloading that is driven by the AttackConfig (theoretically this reloading may depend on the weapon and for critter depend on the critter like the attack strength)

To achieve that I have moved the reloading from a pure FSM condition "a la" TimeElsapsedCondition to Attacker script, otherwise the reloading time was not absolute, it was depending of the state entering.
image
image

And in PigChef transition table, here is the changes to move from attack to Idle/Walking:
image
There is a minimal attack duration (single strike) and a maximum attack duration (double strike) that is used in conjunction with the attack input state.

Finally the reloading check is done when moving from Idle/Walking to Attack:
image

A future improvement would be to add an effect to notify when the reloading is over and the player can request a new attack from the Pig Chef ( An effect on the cane or on the Pig or the phoenix  with one quick light flashing tinting), what do you think?


amel-unity:
Hi @treivize , thanks a lot for opening this PR. After discussing this addition with the team, we prefer to keep the combat system fairly simple without the reloading time. 

As for what we currently have: the combat system is quite solid however we intend to tweak the way the input for the attack is consumed by the state machine so that we only see the attack performed once when we press the attack button. Currently, as long as we are pressing the button, the player keeps attacking so we prefer it more like the jump (you need to press the button again to jump again), what do you think? 

More on the combat system, in the future we will probably do some tweaks for the colliders and maybe some adjustments in terms of visuals like the hit effect but we will definitely share suggestions/updates in the combat system thread. Thanks :)

treivize:
Hi @amel-unity , no problem it makes sense to limit the chaining of attack by forcing the player to press the attack again.
I can update/open a new PR to adjust the behavior. But just to be refine what would be expected, what is the option you have in mind:

Option 1: The player clicks to attack and do not release the button, Pig Chef performs the double strike and come back to idle. The player has to click again to trigger a new attack. If the click is short, only one single strike is played.

Option 2: The player clicks to attack and do not release the button, Pig Chef performs one strike and come back to idle. If the player clicks multiple times (before the end of the first strike), the attack is chaining the strikes.

amel-unity:
More like option 2, and just to add that only the ''press'' matters. We don't need to detect if it's a long press ( and so we don't need to track when the button is released).

amel-unity:
@treivize Do you mind if I close this PR?

treivize:
@amel-unity, no pb, sorry I should have done it already.
I am quite busy these days (school vacations period + job position change in my company), so I have not been able to contribute a lot recently... I might be able to open a new PR during the we for this part.

PR #365, opened 2021-02-13
System for changing camera angles

vjs22334:
Card : https://open.codecks.io/unity-open-project-1/decks/15-code/card/1ak-special-camera-angles
Forum : https://forum.unity.com/threads/camera-cinematics.1051310/
I have implemented the approach I posted in the thread above. I have created a scene : Scenes/Examples/CameraSwapTest where you can test the new system. I have basically created a prefab with a trigger collider which when the player walks into will raise an event for the camera manager to swap the live virtual camera with the one set in the trigger.


UOP1_Project/Assets/Scripts/CameraSwapTrigger.cs
```diff
@@ -0,0 +1,45 @@
...
+            else{
+
+                //player exited on +ve z side which means we switch to our camera
+                VcamEventChannel.OnEventRaised(cameraToSwap);
+            
+            }
+        }
+   }
+
+}
```
ciro-unity, line 45: I feel like this is veery hardcoded. You're thinking of the "entering the trunk" scenario, but what if we just want a special camera for the centre of a very open area? In that case you could enter and exit from all sides.

Instead of doing this, why don't we use the idea of Compound colliders? https://docs.unity3d.com/Manual/CollidersOverview.html?_ga=2.247206724.151196618.1613240916-822635934.1589716631

UOP1_Project/Assets/Scripts/CameraSwapTrigger.cs
```diff
@@ -0,0 +1,45 @@
...
+using Cinemachine;
+
+public class CameraSwapTrigger : MonoBehaviour
+{
+    public CinemachineVirtualCamera cameraToSwap;
+    
+    [Header("Asset references")]
+    [SerializeField] private VcamEventChannelSO VcamEventChannel = default;
+
+    BoxCollider boxCollider;
```
ciro-unity, line 13: Might be better not to assume it's a Box, use generic `Collider` class instead?

UOP1_Project/Assets/Scripts/CameraManager.cs
```diff
@@ -18,36 +23,63 @@ public class CameraManager : MonoBehaviour
...
 
 	public void SetupProtagonistVirtualCamera(Transform target)
 	{
-		freeLookVCam.Follow = target;
-		freeLookVCam.LookAt = target;
+		foreach (CinemachineVirtualCameraBase vcam in vcamsInScene)
+		{
+			vcam.LookAt = target;
+			vcam.Follow = target;
+		}
```
ciro-unity, line 38: This centralised auto-populating list might break if for some reason the function is called from outside before OnEnable.

What about the cameras setup themselves? I know it's a bit against the idea of having the centralised manager, but it would work well.

UOP1_Project/Assets/Scripts/CameraManager.cs
```diff
@@ -96,4 +128,23 @@ private void OnFrameObjectEvent(Transform value)
...
 		SetupProtagonistVirtualCamera(value);
 	}
+
+	private void OnCameraSwapEvent(CinemachineVirtualCamera vcam)
+    {
+		// TODO: find a better way than +10 , -10 priority. may break if scene is setup incorrectly
+        if(vcam != null){
+			if(!cinemachineBrain.IsLive(vcam)){
+				cinemachineBrain.ActiveVirtualCamera.Priority -= 10;
+				vcam.Priority += 10;
```
ciro-unity, line 138: Isn't it safer to use absolute numbers? Like 100 and 0.
Feels safer in case for some reason this function is called twice (-20, +20), then you call the exit once  and it's not enough to bring the situation back to normal.

ciro-unity (commented):
Added a few considerations

UOP1_Project/Assets/Scripts/CameraSwapTrigger.cs
vjs22334: So in case of compund colliders should we use trigger entry and exit to decide which camera to use rather than which side player exits on ?

UOP1_Project/Assets/Scripts/CameraManager.cs
vjs22334: That method is part of a callback registered in onenable so I don't see a case where it can be called before on enable. But anyway I don't quite get what you mean by getting the cameras to set themselves up. Should I create a special vcam prefab with a script in it and do it there?

UOP1_Project/Assets/Scripts/CameraSwapTrigger.cs
vjs22334: So in case of compound collider should we make it such that if the player is in the volume enclosed by the collider then we activate a special camera and when he exits that volume we go back to the default camera? Or do we keep the same behaviour as now but just consider the z axis of the particular child collider of the compound collider the player exits?

vjs22334: Or I could setup the prefab to switch between both behaviours and the level designer can pick

vjs22334:
I have added an option to the prefab to use the entire volume of the collider along with the previous functionality along with generic colliders. I believe these 2 options will cover most of the scenarios which may arise. Is there a specific scenario which you see requiring compound colliders? because it can lead unexpected situations as the callbacks of each child collider will be independent

vjs22334:
Updated the system to work with compound triggers (colliders). The camera will now be swapped when the player enters/exits the volume of the collider. All colliders should be added as children of the parent gameobject of the prefab and should have the CameraSwapTriggerChild script on them. the child scripts will listen for OnTriggerEntry and OnTriggerExit call the parent which will keep track on whether the player is inside the compound trigger.

UOP1_Project/Assets/Scripts/CameraManager.cs
ciro-unity: 'm not 100% sure how to do it. But for instance, if we do it your way then any Vcam will always follow and look at the player. This is not necessarily what we want. The beauty of Cinemachine is that we can create Vcams that are static, on rails, etc. and really tweak each shot in a different way.

So maybe what I'm thinking of is a script which we apply to the Vcams that we want to find the player. Maybe the script is a simple `VCamRequiresPlayerTarget` script with 2 booleans: looksAtPlayer and followsPlayer.

The `CameraManager` would do something like you did here, but look for the `VCamRequiresPlayerTarget` in the scene, and provide the player's target. These scripts would use it for their LookAt, Follow, for both, or for none, based on the booleans above.

How does it sound?

vjs22334: sounds good. I will implement it.

UOP1_Project/Assets/Scripts/CameraManager.cs
vjs22334: implemented

UOP1_Project/Assets/Scripts/CameraSwapTrigger.cs
vjs22334: implemented.

UOP1_Project/Assets/Scripts/CameraSwapTrigger.cs
vjs22334: compound colliders implemented.

vjs22334:
All requested changes have been made

ciro-unity:
Sorry, needed some time to evaluate this. Some thoughts:

We might need to implement two types of colliders: one that is like, "turn this VCam on when entering, turn it off when exiting". This is good for colliders that you can access from all sides.

The other type (useful for situations like the trunk or tunnels) would be: "turn this VCam on when entering, do nothing when exiting". This could be used so that there's **two different VCams running on the same track**, like in the trunk example, and each one can follow the player from behind (so they'd need different settings). This way we can have two entrances (on both sides of the trunk) and depending on which side the player enters from, we activate the correct VCam. I believe it's the only way to avoid the Vcam doing weird gymnastics and going through walls, which happens now when you enter the trunk with the Vcam in front of you from the wrong side (imagine it's already above the trunk), which forces it to go through the trunk's roof to reach the position "in front of you but inside the trunk". If we had two cameras, as you walk inside the trunk and the camera is above it, it would do a spherical motion and transition to "behind you", so that it can follow you walking into the trunk.

The other thing is that we don't need the `CameraSwapTriggerParent.cs` and `CameraSwapTriggerChild.cs` combination to make compound collider work. We can use a script on the root that has an `int` variable inside keeping track of how many colliders are active at the moment. If for instance the player is going out of the last collider, this number would be 0, so the event can be fired.

Something like this:

```
private void OnTriggerEnter(Collider other)
{
	if (other.tag != "Player")
		return;

	colls++;

	if (colls == 1)
		VcamEventChannel.OnEventRaised(cameraToSwap);
}

private void OnTriggerExit(Collider other)
{
	if (other.tag != "Player")
		return;

	colls--;

	if(colls == 0)
		VcamEventChannel.OnEventRaised(null);
}
```

What do you think?

vjs22334:
If we add the second type of collider, we would need another collider to switch the camera back to default (free look) when we leave the trunk. so we would have one collider just outside the  mouth of trunk for free look on just inside for the rails cam on each side for a total of 4 triggers. I don't really see a better option than this. I could go back to my previous version of checking along the z axis of the collider but that feels hard coded and confusing to a level designer.

Regarding your 2nd point about using an int on the parent to keep track of collisions, that is what I did initially but when I put a rigidbody on the parent, the pig chef was always triggering ontrigger events twice. the capsule collider and character controller each seem to trigger it. while this won't be a problem as it will increment and decrement twice. I thought using child collider scripts and booleans is more robust in terms of avoiding this kind of issues. let me know what you think about this @ciro-unity

vjs22334:
I think I found a better way to resolve the Issue we had of needing "two cameras running on the same track". Since we had a child trigger system already. I thought maybe we could choose the camera based on which child trigger was entered. So I added  a camera field to the child trigger which is the camera which will be used if that child is entered. if that field is empty the parent's
camera will be used instead. I feel this system gives a lot of customization options. we can even add a field to for which camera to swap to on exiting a trigger in future if required. Let me know what you think.

Here's a video:


https://user-images.githubusercontent.com/30682201/111940447-c6fc2a00-8af4-11eb-9a2e-f14dccda08a5.mp4

ciro-unity:
Haven't forgotten about this, I'm just waiting until we have some more finalised scenery so we can put it to test in a real scenario!

PR #367, opened 2021-02-13
AudioManager UnregisterChannel fix

Hansen-L:
The `UnregisterChannel` method was not unsubscribing from the delegates properly.

Not sure if this is the right procedure for a small bugfix like this; I didn't think it was worth creating an issue for this


ciro-unity:
> I didn't think it was worth creating an issue for this

Yeah totally, no worries. Good catch too! This might explain some audio weirdness that I was hearing!

PR #369, opened 2021-02-14
Respawn when falling out of bounds

zvirak:
Rough time based detection of falling out of map, and teleporting protagonist to default spawn location.

---
https://open.codecks.io/unity-open-project-1/decks/15-code/card/1a7-respawn-when-falling-out-of-bounds
https://forum.unity.com/threads/respawn-when-falling-out-of-bounds.1057676/


ciro-unity:
Sorry, finally closing this because as mentioned on the forums, we decided to use location entrance/exits to handle falling out of bounds. It's a much simpler way since we tap into an existing systems without having to do modifications to anything, and also allows us to easily create multiple falling volumes that take you back to specific respawn points.

Thanks for the contribution anyway!

PR #370, opened 2021-02-14
Added NPC pathway movement strategy based on ScriptableObjects

treivize:
Forum thread: [NPC movement / Navigation / AI](https://forum.unity.com/threads/npc-movement-navigation-ai.1051322/)
Codecks card: [NPC moving on the Navmesh](https://open.codecks.io/unity-open-project-1/decks/15-code/card/1aa-npc-moving-on-the-navmesh)

The details are posted in the  forum thread here: [post](https://forum.unity.com/threads/npc-movement-navigation-ai.1051322/#post-6835883)

Just a short description:

Currently the moving critter is fully driven by the FSM with absolute conditions and actions, so there is no capability to assign different movement from one critter (or NPC) to another without duplicating the whole FSM table.

Here again, the approach could be to move the FSM configuration of the actions and conditions related to the movement into another SO attached to the game object like done with Damageable and Attack scripts. And next fetch this configuration in FSM to drive the conditions/actions.

[Demo](https://www.youtube.com/watch?v=5WhjrLeO5To)


treivize:
@ciro-unity and @AlexandreGheraibia, I have updated my PR to use the pathway editor tool. Some updates have been made to make it works with SO and also to fetch NavMesh (path computation and snap the waypoints on the navmesh lively)

AlexandreGheraibia:
I have a better solution

AlexandreGheraibia:
you are not probing the existence of NavMesh ?

AlexandreGheraibia:
you don't sounding the NavMesh existance. if NavMesh doesn't exist, your point fails!

treivize:
Not sure it makes sense to define pathways without having baked any NavMesh data available in the scene. Let it fail will tell the editor to bake NavMesh data first in his scene.

AlexandreGheraibia:
the better idea it's use the probe for replacing the point failed.

treivize:
At least I agree with you, I globally check that there is NavMesh data available before trying to run the pathway editor over the scene, it just log once when you click on the SO a warning message. 

Next, regarding the bad NavMesh baking data , it sounds a bit overkill to handle this one. I mean, the navmesh baking might occur before the pathway creation/edition and should be verified before the creation of pathway in the level design process.
@ciro-unity, do you think it would be required?

treivize:
image

AlexandreGheraibia:
In PathwayEditor.cs
you can remove this unnecessary method
```
private void SelectItem(ReorderableList list)
	{
		InternalEditorUtility.RepaintAllViews();
	}
```
and in PathwayGizmos you can remove the else if
```
else if (pathway.Waypoints.Count > 1)
		{
			for (int index = 0; index < pathway.Waypoints.Count && pathway.Waypoints.Count > 1; index++)
```

AlexandreGheraibia:
you can improve the performanaces by using GUI.changed to activate the calculations only when the handles move

treivize:
Hi @amel-unity and @AlexandreGheraibia,

I have updated this PR to take into account the latest version of the pathway editor from @AlexandreGheraibia.
I have replaced the usage of Gizmos.Sphere to Handle.SphereCap() which is compatible with SO custom authoring.

AlexandreGheraibia:
in PathwayGizmos .cs  you can write without [DrawGizmo(GizmoType.Selected)] because the gizmo does not exist with ScriptableObject.

```
public static void DrawGizmosSelected(PathwayConfigSO pathway)
	{
		if (!pathway.ToggledNavMeshDisplay)
		{
			DrawHandlesPath(pathway);
		}
		else
		{
			DrawNavMeshPath(pathway);
		}

		DrawHitPoints(pathway);
	
	}
```
And you can rewrite DrawHitPoints because Handles.SphereCap is obsolete

```
private static void DrawHitPoints(PathwayConfigSO pathway)
	{
		if (pathway.DisplayProbes)
		{
			float sphereRadius = pathway.ProbeRadius;

			for (int i = 0; i < pathway.Hits.Count; i++)
			{
				Handles.color = pathway.Hits[i] ? new Color(0, 255, 0, 0.1f) : Handles.color = new Color(255, 0, 0, 0.1f);
				Handles.SphereHandleCap(0, pathway.Waypoints[i].waypoint, Quaternion.identity, sphereRadius, EventType.Repaint);
			}
		}
	}
```

treivize:
@AlexandreGheraibia, thanks for the review. I have pushed the corrections as suggested.

AlexandreGheraibia:
there is no use case that does not work. it's more of a detail on the gismo. we can't turn off the gismo or we can't see the path in the game scene.

AlexandreGheraibia:
I forgot in PathwayConfigSO.c to put public List <WaypointData> Waypoints; outside the #UNITY_EDITOR directive

amel-unity:
Thanks a lot @treivize! Another great contribution! Thanks @AlexandreGheraibia for the review, great to see you both worked together on this solution :) I merged the PR locally, I am just adding an example using it for the slime critter to the Testing ground scene. So you should see it merged soon 👍

PR #375, opened 2021-02-19
Desserter Character

Smurjo:
Based on female townsfolk with skirt and texture modified. Additionally she has long hair (516 triangles) and a vanilla orchid (740 triangles) in her hair.

desserter_walk_2


ciro-unity:
As mentioned on the forum, can the vanilla orchid be reduced a bit in polygons? 740 sounds like a lot for such a tiny model.

ciro-unity:
Still a few issues I'd like to fix if we want to use this character:
- Can we reduce the flower to less polygons? It really doesn't need more than 150 at best.
image

- I think the animation clip is using the Male townsfolk as a base
- The walk seems fast... has time been scaled?
- The idle has some artifacts (like the eyebrow going through the head)

Thanks!

Smurjo:
I didn't change anything on the animation. I used the male animation because the female animation didn't work at all. I will wait until that is fixed.

PR #376, opened 2021-02-19
Boat with modified sail

Smurjo:
Modified the mesh of the sail since it wasn't suitable for the windshader:
- now 158 triangles
- added slight bulge as if the wind was caught in the sail
- increased thickness in the middle to avoid front and back of the sail overlap with wind

SailMesh
BoatWithModifiedSail


TJHeuvel:
Thanks for this, i'll look at the wind stuff in the weekend!

TJHeuvel:
https://dl.dropboxusercontent.com/s/g89ggspmhu55ds5/w4w6OZ6gMl.mp4
https://dl.dropboxusercontent.com/s/9512irisp0c7ndf/ST85WAdrlj.mp4

Can confirm it already looks a lot better, even with the existing wind solution!

TJHeuvel:
Again, thank you for improving the model, but there are a few things i noticed:

- There seem to be some problems with importing, that you noticed too. In the prefab you've changed the scale of the wind sail, and rotated it around. This causes issues with the wind-shader, would it be possible to fix these in the modeling program?
image

- The boat is made of 3 different objects, the original fbx too. Why is this? Its inefficient to render this as three different draw calls. It could be one for the entire thing, or perhaps two so we have a separate material for the sail. I dont see why the mast has to be another object.

- I think its best to overwrite the current boat fbx, instead of making a new one with your name appended. With GIT we always have history of the project, so we can go back to the old version, or even see that you've made it. Like this we'd have a _ton_ of models real fast, and we wouldnt know which one to use!

- The model comes without vertex colors for the weight of the sail, do you think it makes sense to do so in the modeling program instead of in Unity?

ciro-unity:
Checked it out in the editor: as I understand you're still working on the shader?

Smurjo:
Just cleaning up my Github..

PR #377, opened 2021-02-20
Added a small, simple Tag class that holds tags.

alimertcetin:
I dont have a forum thread linked to this PR
The purpose of this PR is avoid misspelling while comparing tags
These changes are not necessary. I just think that it is easier to compare tags like this;
if at some point someone desides to change a tag, it can be easy on the editor side but if you have couple of tag comparison on the code then can be a little mess.
By using "Tag" class you will be able to add, remove and rename tags easly.
When defining a tag you need to define at editor and the "Tag" class. That could be count as the downside.

Renaming is much easier
EasyRename

While you are typing it is very easy to make mistake and hard to find the issue but this way you have to just check if the tag defined correctly or not and you dont even have to open the editor to see defined tags,
EasyToCheck

IDE will warn you if tag is deleted.
WhenTagDeleted


schodemeiss:
My $$ on this. I don't really understand the methods in Tag.cs. If the goal is to statically type the tags via the public const strings, then why would you ever need "TagExists" and "GetTags"?

Whilst these are perfectly fine bits of C# it feels a little redundant to include those methods. If anything you're encouraging other developers to continue using magic strings in their code to check if a tag exists, when they should just be using the defined constants in the first instance. You could mark those Methods as [Obsolete("Please use the consts")], but I feel if you know about those methods, then you know about the consts already.

My only other issue is if another developer adds a new tag, they have to be aware of this new class and update it with additions or changes. If you wanted a really robust solution, I'd advise writing an Editor method that automatically generates the Tag.cs file and places it in the asset folder. Thankfully, C# includes some handy code generators and Unity can give you an array of the tags as strings:

`string[] tags = InternalEditorUtility.tags.OrderBy(tag => tag).ToArray();`

With that array, check out the System.CodeDom.CodeCompileUnit namespace. It contains loads of methods for generating C# code for you.

schodemeiss:
> You can delete the methods if you dont need them of course.
> 
> Again, thanks for your effort @schodemeiss !

Given I've created an entirely separate pull request for my additions, it's probably cleaner to close this PR to save any confusion should the PR's be accepted. Otherwise, they'll just have to reverse this one out should mine be accepted. In hindsight I should have forked your repo, made the changes and submitted PR's to your repo for you to incorporate - but I didn't think about that at the time, my apologies.

PR #379, opened 2021-02-20
Add Toon Wind Sail shader

TJHeuvel:
Ticket: https://open.codecks.io/unity-open-project-1/decks/16-3d-shading-animation/card/1ax-sail-shader

Attempt at making something similar to: https://youtu.be/iykoYKq42uU?t=262

New shadergraph that uses the wind module, and then adds more bulging depending on where the object is facing. First iteration, works best with the new sail model being made by [Smurjo here](https://github.com/UnityTechnologies/open-project-1/pull/376), with some tweaks.

[This is the result](https://dl.dropboxusercontent.com/s/cuw3cni3p2yjw9u/Bs6lC8dGNP.mp4), with both the current boat model as well as the new sail that Smurjo is making. I like the initial effect, but would like some feedback on how to improve it, and for what it will be used. 
When comparing it with the effect from Zelda it seems like they mostly have the sail either extremely to one side or the other, there is hardly ever in between. This is difficult to emulate in only shaders, but perhaps modifying the dot product can give better results. E.g. right now the sail is linearly pushed left or right depending on the rotation of the boat, if we make this quadratic it will already behave much more erratic.

I've made a new material for the sail specifically, not sure what all those changes in the Boat material are. The changes in the Boat prefab are assigning the new sail material, and my rough vertex painting. Please feel free to overwrite those with something better!

**[for Generic PRs]**  
https://forum.unity.com/threads/wind-shader-add-on.1004935/#post-6858731
I hope this brings a fairly easy to understand effect, namely bulging of the sail according to the object rotation.


TJHeuvel:
Please note this is also my first time using ShaderGraph, if i'm doing something obviously wrong please point it out!

TJHeuvel:
Maybe we should close this one, and just use https://github.com/UnityTechnologies/open-project-1/pull/376 for both the sail and the model

ciro-unity:
Ok, I'll close it!

PR #381, opened 2021-02-22
Respawn when falling OOB

ontrigger:
Card: https://open.codecks.io/unity-open-project-1/decks/15-code/card/1a7-respawn-when-falling-out-of-bounds

Forum: https://forum.unity.com/threads/respawn-when-falling-out-of-bounds.1057676/

Made it so that the game will check for ground below the player if they are falling for more than the set amount of time.
If there is no ground, then the player will be teleported to the last valid grounded position (updated in the walking state).


ciro-unity:
Sorry @ontrigger, I'm ultimately closing this since we used a different technique which integrates with the existing entrance/exit prefabs.
But thanks anyway!!

PR #384, opened 2021-02-27
Town Whiteboxing

itsLevi0sa:
You can find the scene in the "Whiteboxing>3. Town>Proposal_3" Folder, along with a Materials folder, the lightmap data folder and a script for the AreaLoad (I decided not to discard it, even though we are not going to use it, it is still being used in the scene).
Also added a SceneDataSO named Town in the "SceneData>Locations" folder that created a change on the "Assets\AddressableAssetsData\AssetGroups\Default Local Group.asset" that I included in the commit since discarding it didn't let me play the scene properly. Mentioning that because it was the only change that was orange. Enjoy!(hopefully)


ciro-unity:
Thanks @itsLevi0sa!

Two very innocuous comments:
- Try not to modify the configuration of Addressables (I assume you marked the scene?), we can do it while merging the PR or later. That might create conflicts that are hard to solve.
- No spaces in filenames and folder names, though this is less important for whiteboxing scenes which are more throwaway... but still :)

itsLevi0sa:
I noticed the change on the Addressables was created as soon as I'd drag my SceneDataSO on the EditorInitializer. (I didn't do any marking, not even sure what that means) That's why when I discarded that change I couldn't play the scene (it would give me an error). So that's why I mentioned it here also and decided to put it in the commit, even though it was one of the scariest things to do! :P All comments noted, thanks!

PR #287, opened 2020-12-30
Editor Window to efficiently find SOs

HarshNarayanJha:
Do you have a forum thread linked to this PR?  
https://forum.unity.com/threads/editor-window-to-efficiently-productively-find-scriptable-objects-sos.1030345/

What will this PR bring to the project for everyone?  
Efficiency and more productivity.

Why are these changes necessary?  
Because I have not contributed to this project yet and I want to.

How did you implement them?  
Details in Forum Link.

PR #387, opened 2021-03-06
Improve SOsQuickAccessToolWindow and rename to ScriptableObjectBrowser.

IsaiahKelly:
Improved version of the ScriptableObjects quick access editor window from #287. 

This is really just a highly modified version of SOsQuickAccessToolWindow.cs, which I renamed to ScriptableObjectBrowser.cs. In hindsight I guess I should have renamed the file before or after making such massive modifications to ensure git didn't think it was a brand new file, but I hope this is still okay.

[Related forum thread](https://forum.unity.com/threads/editor-window-to-efficiently-productively-find-scriptable-objects-sos.1030345/)

This PR improves upon the great contribution by @HarshNarayanJha to add a dedicated editor window for locating ScriptableObject assets in the project. Changes include:

- Improved navigation (additional scroll view instead of massive popup menu)
- Improved text formatting (including the removal of type namespaces).
- Improved efficiency (clicking an item selects it instead of just pinging it).


HarshNarayanJha (approved):
Nice.

PR #388, opened 2021-03-09
Combat System - Attack input consume based on animation events

treivize:
Forum: [Combat System](https://forum.unity.com/threads/combat-system.1032685/)
Codeck card: [Implement second cane hit animation](https://open.codecks.io/unity-open-project-1/decks/16-3d-shading-animation/card/19v-implement-second-canehit-animation)

This new version relies on Animation events to consume the attack input. This way the player can register for a second hit right after the first one has effectively started. And the timing depends on the cane animation.

Playing with it, it seems quite smooth and natural. Let me know how it fells for you @amel-unity.


amel-unity:
This is great! Thanks a lot @treivize. The attack feels much better now that it relies on the animation events to consume the input. I will merge this now 👍

PR #391, opened 2021-03-25
Refactored RuntimeAnchorBase to be more generic

schodemeiss:
Refactored RuntimeAnchorBase to be more generic, along with using a WeakReference (https://docs.microsoft.com/en-us/dotnet/api/system.weakreference-1?view=netstandard-2.0)  so we don't stop garbage collection unnecessarily.

Updated TransformAnchor and PathAnchor to use the more generic RuntimeAnchorBase. I marked them as "Sealed" as nothing is inheriting them for now.

I've made the required changes throughout the code to support those updated anchors.

I made these changes as I was working on another feature and I happened to require some anchors, so in the spirit of the Boy Scouting code rule (https://www.matheus.ro/2017/12/11/clean-code-boy-scout-rule), I felt I should make these changes before moving on.

### DIFF
--- a/UOP1_Project/Assets/Scripts/Camera/CameraManager.cs
+++ b/UOP1_Project/Assets/Scripts/Camera/CameraManager.cs
@@ -11,7 +11,7 @@ public class CameraManager : MonoBehaviour
 	private bool _isRMBPressed;
 
 	[SerializeField, Range(.5f, 3f)]
-	private float _speedMultiplier = 1f; //TODO: make this modifiable in the game settings											
+	private float _speedMultiplier = 1f; //TODO: make this modifiable in the game settings
 	[SerializeField] private TransformAnchor _cameraTransformAnchor = default;
 
 	[Header("Listening on channels")]
@@ -37,7 +37,7 @@ public class CameraManager : MonoBehaviour
 		if (_frameObjectChannel != null)
 			_frameObjectChannel.OnEventRaised += OnFrameObjectEvent;
 
-		_cameraTransformAnchor.Transform = mainCamera.transform;
+		_cameraTransformAnchor.Anchor = mainCamera.transform;
 	}
 
 	private void OnDisable()
--- a/UOP1_Project/Assets/Scripts/Characters/Protagonist.cs
+++ b/UOP1_Project/Assets/Scripts/Characters/Protagonist.cs
@@ -71,12 +71,12 @@ public class Protagonist : MonoBehaviour
 		float targetSpeed = 0f;
 		Vector3 adjustedMovement;
 
-		if (gameplayCameraTransform.isSet)
+		if (gameplayCameraTransform.IsSet)
 		{
 			//Get the two axes from the camera and flatten them on the XZ plane
-			Vector3 cameraForward = gameplayCameraTransform.Transform.forward;
+			Vector3 cameraForward = gameplayCameraTransform.Anchor.forward;
 			cameraForward.y = 0f;
-			Vector3 cameraRight = gameplayCameraTransform.Transform.right;
+			Vector3 cameraRight = gameplayCameraTransform.Anchor.right;
 			cameraRight.y = 0f;
 
 			//Use the two axes, modulated by the corresponding inputs, and construct the final vector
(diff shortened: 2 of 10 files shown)


schodemeiss:
I'd also like to do a similar bit of refactoring around the EventChannelBaseSO and EventListeners. There's a lot there that could be refactored into simpler generic classes that would result in much less duplicated code. However, I'd rather do that if someone over at Unity (@ciro-unity, @amel-unity ?) officially gives me the go-ahead, as I think that will touch quite a few more areas than this much simpler Anchor example above.

schodemeiss:
Apologies for the force "pushes", I messed up a rebase whilst resolving a new conflict. It was easier for me to pull from main and cherry-pick the commits and push it all again. This should be mergeable again now.

ciro-unity:
Sorry just getting to this. Not sure about the whole refactor. I think people misunderstood the concept of "anchors". The `PathSO` doesn't have much reason to be as an anchor, it was pulled in as part of a PR but never should have been.

Anchors are intended as a way to couple two GameObjects via a ScriptableObject. In the original use, GO A would reference the Transform of GO B. For a ScriptableObject (like `PathSO`) there's no need, it can just be referenced directly. So I will change that when I can.

Another question: what's the benefit of using WeakReference? I think I don't fully get the intention 🤔
What happens if the Camera (for instance) accesses the player's Transform, and this has been collected by the garbage collection? Like, why would we want that to happen?

schodemeiss:
> Anchors are intended as a way to couple two GameObjects via a ScriptableObject. In the original use, GO A would reference the Transform of GO B. For a ScriptableObject (like `PathSO`) there's no need, it can just be referenced directly. So I will change that when I can.

I understood the original use case for them was to effectively bridge two objects together; this PR obviously doesn't change that behaviour at all. I totally agree that some of the usages in the project could just be direct references, like the PathSO you mention above. Even the word 'Anchor' doesn't make me entirely feel warm and fuzzy ;-).

> What happens if the Camera (for instance) accesses the player's Transform, and this has been collected by the garbage collection? Like, why would we want that to happen?

In my testing, this hasn't ever occurred, primarily because the objects currently being anchored (Camera, GameObject or whatever thing being anchored) is typically remains in scope during gameplay. Using your example, the player's Transform wouldn't be GC'd in the first place as it's always being referenced by something else in the scene. With the Anchors as they are right now, because the Anchor is **also** holding a reference to that player Transform, the Transform wouldn't be GC'd, even if that Transform were to be destroyed (as the Anchor SO is still referencing it and the SO appears to remain in scope).

In short, whatever is being anchored right now will hold a reference to that object (thus stopping GC reclaiming the memory) whilst the Anchor SO is in scope (which appears to be always).

The intention of using the WeakReference was similar to how you need to unregister methods from listeners / events to stop memory leaks. Without the WR, it'll forever hang onto the anchored object, regardless of whether the anchored object has been destroyed or not, thus the GC can't reclaim it. The WR will (eventually) allow the GC to reclaim the thing being Anchored once it has been destroyed (much like unsubscribing from events).

None of what I've described above causes large memory overheads given the small number of anchors currently in use, and unlike events, it's not an array of delegates which typically grow in size with every subscriber; but the WR would assist in not unintentionally clinging onto object references that can be reclaimed by the GC. If the Anchor SO goes out of scope, the OnDisable callback setting the Anchor to 'null' also frees up the anchored object for GC, but the objects being anchored aren't guaranteed to be tied to the lifetime of the Anchor SO; especially true if a developer creates further anchors.

All that aside, as I mention above, I'm not entirely sure that Anchors are quite the right answer for some of the current use cases and I'm **certainly not precious** about this PR. I wont be offended one bit if it's closed :-).

ciro-unity:
> Using your example; the player's Transform wouldn't be GC'd in the first place as it's always being referenced by something else in the scene. With the Anchors as they are right now, because the Anchor is **also** holding a reference to that player Transform, the Transform wouldn't be GC'd, even if that Transform were to be destroyed (as the Anchor SO is still referencing it and the SO appears to remain in scope).

But... in the current scenario, when the character faints and a new one is spawned (or when changing scenes), as the anchor is provided a new reference to the new Transform, wouldn't the old Transform be collected by GC anyway?

In general it's a valid concern, I'm just a bit wary of adding unneeded complexity to this project since it's being visited by SO many people and each decision and extra layer of complexity needs to be explained.

PR #393, opened 2021-03-26
Editor tool: Quest Editor Window

BrettHuang2019:
**Forum:**
[Forum link](https://forum.unity.com/threads/quests-and-scenario.1041601/page-2#post-6975674)

**Purpose:**
- Add a Editor Window to easily review and config different questlines / quests / steps ...
- Preview the whole quest in the Editor Window with all steps. 

**Preview**
QuestEditorWindow_Preview

**Files Changes**
(10 Files Added)
2*C#
> QuestEditorWindow.cs (Main Editor Window Script)
> LocalizationUtils.cs (Plugin for loading LoacalizedString in Editor, thanks to Localization team from forum)

3*uxml
> QuestEditorWindow.uxml (Main panel uxml)
> StepDetail.uxml (Quest part)
> DialogueLine.uxml (Dialogue part)

1*uss
> QuestEditorWindow.uss (Only 1 stylesheet in used)

4*png
(All NPC portraits with the same name for the quest preview)
> BardHare.png
> Earl.png
> TownsFolk.png
> FryKing.png

Hope it helps!
Please review my code. It's not that clean in my opinion.. 

Thank you!
Brett Huang


UOP1_Project/Assets/Scripts/Quests/Editor/LocalizationUtils.cs
```diff
@@ -0,0 +1,67 @@
...
+    /// Gets the localized string from an already loaded table, taking into account whether we are in edit mode, play mode, or a build.
+    /// </summary>
+    /// <param name="localizedStringReference">The <see cref="LocalizedString"/>.</param>
+    /// <returns>The localized string.</returns>
+
+    public static string GetLocalizedStringImmediateSafe(this LocalizedString localizedStringReference)
+    {
+        // If we are in the editor in edit mode, we need to find a valid locale and get the localized string from it:
+        #if UNITY_EDITOR
+        {
```
schodemeiss, line 43: You can lose the {}'s here. The compilation defines don't require them at all.

UOP1_Project/Assets/Scripts/Quests/Editor/LocalizationUtils.cs
```diff
@@ -0,0 +1,67 @@
...
+            if (!EditorApplication.isPlaying)
```
schodemeiss, line 44: You could also invert these if's to reduce nesting. For example:

`if (EditorApplication.isPlaying) return;
// Rest of Code`

UOP1_Project/Assets/Scripts/Quests/Editor/QuestEditorWindow.cs
```diff
@@ -0,0 +1,289 @@
...
+	    foreach (var item in enumberable)
+	    {
+		    data = item as T;
+	    }
+	    return data;
+    }
+
+    private void SetUpQuestPreview(QuestSO quest)
+    {
+	    if (quest != null)
```
schodemeiss, line 94: Again, you can reduce nesting here by doing something like:

`if (quest == null) return;
// rest of code`

AlexandreGheraibia:
Hi I'm not a C # expert but using where T: Object seems weird.

https://stackoverflow.com/questions/10644703/c-sharp-generics-constraining-t-where-t-object-doesnt-compile-error-constr

schodemeiss:
> Hi, I'm not a C # expert but using where T: Object seems weird.
> 
> https://stackoverflow.com/questions/10644703/c-sharp-generics-constraining-t-where-t-object-doesnt-compile-error-constr

Which line, exactly? It's entirely likely either way that due to a "using UnityEngine" , it probably refers to UnityEngine.Object rather than the C# Object.

AlexandreGheraibia:
> 
> 
> > Hi, I'm not a C # expert but using where T: Object seems weird.
> 
> If you're referring to this line
> 
> ```cs
> private void FindAllSOByType<T>(out T[] foundSOs) where T : Object //bug: if I wrote "ScriptableObject", it doesn't work.
> ```
> 
> I'm quite sure because of the import of " using UnityEngine", the Object is actually a "UnityEngine.Object", which makes sense. However, I am a little curious as to why "ScriptableObject" didn't work for @BrettHuang2019.

You must be right, like I said I 'm not C# expert.

BrettHuang2019:
> > Hi, I'm not a C # expert but using where T: Object seems weird.
> 
> If you're referring to this line
> 
> ```cs
> private void FindAllSOByType<T>(out T[] foundSOs) where T : Object //bug: if I wrote "ScriptableObject", it doesn't work.
> ```
> 
> I'm quite sure because of the import of " using UnityEngine", the Object is actually a "UnityEngine.Object", which makes sense. However, I am a little curious as to why "ScriptableObject" didn't work for @BrettHuang2019.

Yes. I don't understand either.. It doesn't compile when I use "ScriptableObject",  which force me to use Object to make it work.

schodemeiss:
> Yes. I don't understand either.. It doesn't compile when I use "ScriptableObject", which force me to use Object to make it work.

I cloned your fork and tried changing the "Object" to "ScriptableObject". In both instances it compiled fine on my side and the tool seemed to work exactly the same. Both the Object's on lines 205 and 217 can be changed. Give it another go? Perhaps the UnityEngine important statement was missing at the time or something?

BrettHuang2019:
> > Yes. I don't understand either.. It doesn't compile when I use "ScriptableObject", which force me to use Object to make it work.
> 
> I cloned your fork and tried changing the "Object" to "ScriptableObject". In both instances it compiled fine on my side and the tool seemed to work exactly the same. Both the Object's on lines 205 and 217 can be changed. Give it another go? Perhaps the UnityEngine important statement was missing at the time or something?

Yes you are right. Maybe I have namespace problem before..

uChema:
Hey Here! 
Awesome tool ! I loved using it! And like I said in the forums it's the perfect way to have an overview on the Quest system ! 

I added an "Add" buttons to the Questline, Quest and Steps to the tool. 

I also added some controls so that we don't have "out of bounds" errors, or "Null reference" exceptions. 
`	if (currentSeletedQuest != null && currentSeletedQuest.Steps != null )`
Control over the selected quest and selected questline 

`if (quest.Steps == null)
			return;
		if (quest.Steps.Count>0&& quest.Steps[0].Actor!=null)
	    LoadActorImage(quest.Steps[0].Actor.name);
`
Control over the actor in the steps 

`           if(itemsSource.Length>0)
	    listview.selectedIndex = 0;`
Control over the selected index if the selection is empty 
Along with other controls scattered everywhere. 

I found out that I made an error in creating the QuestSO.cs and QuestlineSO.cs, switching the classes in the files. So I fixed those as well. 

Finally, I hope you don't mind but I changed a bit in the colors used in the tool. 

https://user-images.githubusercontent.com/59874519/115260459-0db06480-a133-11eb-9705-5ceb9e3dda5f.mov

There are still some issues with the refresh when informations are changed in the SO, but I'm still working on that.

BrettHuang2019:
> those
Hi Chema, 
The add button is something I wanted to do at first. But it seems like a lot of work and technical concerns. For example, should it be also a 'delete' button too? I thought the project is going to end soon so I didn't go too far on this..
I am happy you did this! Good for me to learn from your code... 
I also find the Quest/Questline file problem. I tried to fix it but it seems like it breaks everything... So I didn't touch that. 
Color scheme looks great. I just pick random color before..
Thank you!

uChema:
Hey Brett ! Completely understandable about the technical concerns ! It was difficult to implement and I still think there's room to improve it. 
I also thought about a delete button but this feature implies a lot of dependencies (file naming/ steps-quests-questlines dependencies ... ). 
I'll continue working on the refresh of the tool, then merge this into main so we can use it to create quests and questlines using it.

PR #394, opened 2021-03-26
Added L0D Models for Townsfolk

MortyJHin:
Adding the L0D models for the Male and Female Townsfolk. 

Located in Arts/Townsfolk(M/F) and contain an FBX file and a Prefab that includes an Animator/Textures.  

https://youtu.be/pwvpk4ETTMI


amel-unity:
Hi @MortyJHin, thanks a lot for making the LOD Models for the townsfolk, I just merged the PR and made a quick commit to rename to LOD instead of L0D. Thanks :)

PR #396, opened 2021-03-28
Create save file on editor cold startup if it is not present (allow users to use play mode more than once)

fqureshi:
I think it makes sense to create the save file on editor cold startup, since we have developers who may play the game without going through the initialization scene.

I first noticed this when trying to enter play mode on the beach scene for the second time. I had no save file present, which caused the beach scene to not fully load and display a "no cameras rendering" message. 

As for how play mode works the first time.... I believe this is due to the async scene loading that we are doing with `EditorColdStartup`. 

The first time you hit play mode, you will observe that the `EditorColdStartup` will call `loadEventChannelSO.RaiseEvent` before the save system has had a chance to subscribe to `_loadLocation.OnLoadingRequested`. Because the save system is not subscribed, we haven't done any saving, so there is no error.

The second time you hit play mode, the save system subscribes to  `_loadLocation.OnLoadingRequested`, and _then_ `EditorColdStartup` will call `loadEventChannelSO.RaiseEvent`.

Async loading shenanigans aside, it makes sense to me to create the save file for devs who use play mode on random scenes. That way, they don't run into save issues.

To Reproduce: 
1. Ensure you have no save file present.
2. Load up any scene besides the initialized scene (I used the beach scene)
3. Press play mode, and then exit
4. Press play mode again
5. Scene will not be fully loaded, and you will see a "no cameras rendering" message.

### DIFF
--- a/UOP1_Project/Assets/Prefabs/Gameplay/EditorInitializer.prefab
+++ b/UOP1_Project/Assets/Prefabs/Gameplay/EditorInitializer.prefab
@@ -51,3 +51,4 @@ MonoBehaviour:
     m_SubObjectName: 
     m_SubObjectType: 
     m_EditorAssetChanged: 0
+  _saveSystem: {fileID: 11400000, guid: edc355c4a7d5028408d322b90814e19e, type: 2}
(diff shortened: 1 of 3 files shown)


uChema:
Great idea for the Save load on editor cold startup! 
I'll merge it and add a change to it. Instead of writing an empty save file by calling` _saveSystem.WriteEmptySaveFile(); `we're going to set the save as if it was a first time Game `_saveSystem.SetNewGameData();`

Issue #350, opened 2021-02-03
The character gets stuck between rocks


Eddrico:
Short description
The character gets stuck between rocks, the directional input is recognized but the character does not move, the character can get out of the place after several tries, This is reproducible with keyboard and Joysticks, but keyboards get stuck longer than joysticks.

Expected behavior
The Character should be able to move.

To Reproduce

    1.- Open a scene with the TestingGround. Press play.
    2.- Jump over the rocks and place the character between them.
    3.- observe that you can not move, you can jump, but can not move in the air either.

Notes

    Here's a video of the issue, you can see the inputs while i tried to get out of the place:
[https://www.youtube.com/watch?v=rdt4VixZmJI&feature=emb_logo](url)

phx1999:
Though directional input is useless, you can easily jump out the the stuck. 
Do you mean that you want use WSAD to leave the stuck place? (It seems unreasonable)
Or you want to see the movement ainimation?
Or change the rock architecture so that the character can leave the stuck?

JenniTheDev:
I opened up the TestingGround scene and reproduced the bug that the PigChef gets stuck in the same place. I then went to try and reproduce the same bug in other scenes and places. 

Looking at the State Machine Script, whenever the Pig Chef is stuck, the Current State is Sliding. In that state, the only thing it can do is transition to JumpAscending if it's not actually moving. I think this might be causing the bug.  

[Here is a video showing PigChef getting stuck in  a few different places, along with the inspector showing the state changes](https://youtu.be/3tqUvufPbPM)

JoaoSant0s:
I'm studing the SlideActionSO.cs and I think that we can use a similar aprouch that was apploued in AerialMovementActionSO.cs

With that, we can aplly a Slide_Resistance to control the movement velocity of the protagonit in this situation.

Trying to figurirout a mathematical alternative to use same logic of the others actions

JoaoSant0s:
I tried other aprouch for this problems. I made a comment in the PR description

amel-unity:
Hey everyone,
I think it is intended to not be able to leave by simply walking. You just need to jump. So I am not sure if we can still condiser this an issue since we are not stuck but we just need to jump to get out of the rocks. Any thoughts on this or shall we close the issue? 
Thanks :)

JenniTheDev:
I think that it can be closed. Maybe a new bug should be made for the player being able to use this to jump up rocks into places where they really shouldn't be?

[...]

PR #397, opened 2021-03-28
fix: making the chef walk between rocks

JoaoSant0s:
Issue: https://github.com/UnityTechnologies/open-project-1/issues/350
Forum: https://forum.unity.com/threads/the-character-gets-stuck-between-rocks.1050683/

Solution:
- I tried make a minimum modification in the game
- To that, studie all state machine: conditions, actions and transictions
- Checking the IsSlidingCondition I noticed than current slopeLimit its lower for some problematics situations
- So I increased: 45 -> 65 to solve the stucked problem

Verification:

1. Open a scene with the TestingGround. Press play.
2. Jump over the rocks and place the character between them.
3. Try walk, attack and jump to get out of the rocks

### DIFF
--- a/UOP1_Project/Assets/Prefabs/Characters/PigChef.prefab
+++ b/UOP1_Project/Assets/Prefabs/Characters/PigChef.prefab
@@ -171,7 +171,7 @@ CharacterController:
   serializedVersion: 2
   m_Height: 1.3
   m_Radius: 0.3
-  m_SlopeLimit: 45
+  m_SlopeLimit: 65
   m_StepOffset: 0.05
   m_SkinWidth: 0.02
   m_MinMoveDistance: 0.001


amel-unity:
Hi @JoaoSant0s , thanks for opening the PR. I just added a comment on [the issue ](https://github.com/UnityTechnologies/open-project-1/issues/350)because I am not sure if this is really considered a bug. Also, I think just tweaking the value would not be the solution in case we decide that the issue is valid. This will make it possible to walk to get out of the rocks but may also impact the character's behavior on other locations which might produce an unexpected/ unintended behavior. 
Thanks!

JoaoSant0s:
Great @amel-unity I saw this bug report in the Codecks report

- https://open.codecks.io/unity-open-project-1/decks/13-bugs/card/1ar-the-character-gets-stuck-between-rocks

Can you filter the reports and remove olds tasks?

amel-unity:
Hi @JoaoSant0s , we updated the bugs in Codecks so all should be good now. Thank you :)

PR #398, opened 2021-03-28
Editor Tool: Tag Class Generator - Tackles issues outlined in PR #377

schodemeiss:
**Overview**

Whilst I absolutely agree with the idea defined in #377 and commend @alimertcetin for their efforts in tackling the issue, I felt having a class that couldn't be generated by the Unity Editor was potentially not a robust solution longer term. To alleviate some of those concerns, I have created a simple [Menu Item](https://docs.unity3d.com/ScriptReference/MenuItem.html) that automatically creates the Tags class on your behalf, along with some customization and safety measures to protect against any overwriting accidents.

**Thoughts / Feedback:** 

1.  **Update:** I discovered the magic of [EditorApplication.projectChanged](https://docs.unity3d.com/ScriptReference/EditorApplication-projectChanged.html) and updated the generator to listen for project changes, then as efficiently as possible update the Tags.cs if required. ~~If anyone knows of a hook or callback in the Unity Editor for when a tag is added or removed that I could use to automatically run this generator, that'd make this even more robust than it currently is. Lemme know!~~

2. I haven't scoured the codebase itself and actually implement the usage of the Tags class in the rest of the project as I didn't want to create a nightmare merge. So this PR is intentionally left simply as 4 additions. If you'd like me to change all the usages of CompareTag() to use the new class,  I certainly can do, just ping me an issue (assuming this PR is accepted, of course) and I'll make another PR with those changes.

3. I potentially could have earnt an extra gold star ⭐ had I also written a [SettingsProvider](https://docs.unity3d.com/ScriptReference/SettingsProvider.html) rather than putting the configuration in the [TagClassGenerator.cs](UOP1_Project/Assets/Scripts/Editor/TagsClassGenerator.cs) itself (see [Configuration Region](https://github.com/Alkimee/Unity-Technologies-Chop-Chop/blob/c9c84d4ac77a92805c5c31de6a60b2729b6b79e3/UOP1_Project/Assets/Scripts/Editor/TagsClassGenerator.cs#L18)), but given I don't expect these values to change often (if at all), I didn't feel the extra complication was warranted in this instance. But, again, I'd be happy to do that should it become apparent that's required.

4. **Update**: See #401 ~~Layers! Yes, the same can also be done for generating a Layers file. Again, I'm happy to throw that together in a similar way, along with the [LayerMasks](https://docs.unity3d.com/ScriptReference/LayerMask.html) and some helper methods should you guys think that's useful.~~

**The Code**
- **Update**: The Tags.cs file is now automatically re-created when tags are updated, but you can force a regeneration with the menu item. ~~Adding a menu item that automatically generates a [Tags class](UOP1_Project/Assets/Scripts/Tags.cs) so we can use constant defined strings for tag comparison in the scripts.~~

- The generator is pre-configured to generate a file called "[Tags.cs](UOP1_Project/Assets/Scripts/Tags.cs)". It contains a sealed class called "Tags" which resides in the "UOP1" namespace. This class contains "public const strings" of each tag in the project.

- **Update:** I removed the FileExists check given the file is generated automatically now. ~~There are protections in the generator to stop accidental overwriting of already existing files in the project should the configuration values change.~~

- The generated file Tags file should be committed to the repository (and is included in this commit) to save others from regenerating it.

- ~~Finally, it contains instructions on how to regenerate the class when a tag is added or removed from the project.~~


alimertcetin:
Im very glad to see that. I was trying to give a starting point in #377 . Lately i was very busy so that i couldn't look what is goin' on. As i see your approach is going very well. Good luck @schodemeiss !

schodemeiss:
> Im very glad to see that. I was trying to give a starting point in #377 . Lately i was very busy so that i couldn't look what is goin' on. As i see your approach is going very well. Good luck @schodemeiss !

No worries at all, I appreciate you kicking things off and it reminded me I needed to put this together for my own project anyway. It's worth getting this low-level stuff into project as soon as possible; so I'm hopeful it'll get accepted in, but we'll see. If it doesn't, you're welcome to nab it for you're own project either way :-).

I think my approach is **there** now. I can't think of any other fringe cases where this would blow-up in ways that you **don't** want it to.

The code **should** now:

- Generate the Tags.cs with public const string's for each tag.
- Detect when the project has added new, renamed old, or removed tags entirely - regenerating the file in the process.
- Warn you if the Tags.cs has been deleted or has had its class or namespace changed (even though the header in the file says not to, it doesn't stop a developer doing so).
- Once the consts are in use, you should get errors prior to compilation if any of those tags have changed or if the generator itself has changed where it's storing it's Tag class (IE, you change what the tag class is called, or it's namespace). Both of which are the good kind of compile errors you want as it'll mean you can't get runtime bugs by using out of date tag strings (which was the whole point of your PR in the first place).

I'd really welcome a review if you've got time, though? Anything you can think of that I've missed? I also did the same for Layers, but using Enums (https://github.com/UnityTechnologies/open-project-1/pull/401).

alimertcetin:
> If it doesn't, you're welcome to nab it for you're own project either way :-).

Thanks! Probably i'll need that :)

> I'd really welcome a review if you've got time, though?

Checked out your code and it seems like you covered all the things... I didnt see in the editor though but i think that "this is it!" :)

Let's see now what others will say about it.

schodemeiss:
After some deliberation, I decided to have a go at writing this using SettingsProviders and some other improvements I thought I could make along the way. With that, I created a new PR #402 which is essentially this PR, but taken to the next level.

PR #401, opened 2021-03-29
Editor Tool: Layer & LayerMasks Enum Generator - Tackles issues related to PR #377 & PR #398

schodemeiss:
**Overview**

Similar to my other Pull Request #398, but this generates a pair of Enums for the layers in the project. These enums can then be used anywhere that you might wish to check a layer or use a layer mask.

The "Layer" enum is simply the layer ID's and be used to set a layer on a GameObject or compare a single Layer on a GameObject. See [GameObject.layer](https://docs.unity3d.com/ScriptReference/GameObject-layer.html).

The "LayerMasks" enum has the [FlagsAttribute](https://docs.microsoft.com/en-us/dotnet/api/system.flagsattribute?view=netstandard-2.0) set so you can combine them or perform any kind of bitmasking against the Enum. You can then use this mask anywhere you can use [LayerMask](https://docs.unity3d.com/ScriptReference/LayerMask.html), such as in Physics.Raycast. You can also use either the Layer or LayerMasks enum as members of MonoBehaviours or ScriptableObjects as Unity now supports Enum with Flags natively in the Editor.

I also added in an extension method for [LayerMask](https://docs.unity3d.com/ScriptReference/LayerMask.html) to make it easier to check if a single Layer ID or a GameObject's layer is in a given [LayerMask](https://docs.unity3d.com/ScriptReference/LayerMask.html).

**Thoughts / Feedback:**

1. **Update:** I discovered the magic of [EditorApplication.projectChanged](https://docs.unity3d.com/ScriptReference/EditorApplication-projectChanged.html) and updated the generator to listen for project changes, then as efficiently as possible update the Layers.cs if required.
~~If anyone knows of a hook or callback in the Unity Editor for when a layer is added or removed that I could use to automatically run this generator, that'd make this even more robust than it currently is. Lemme know!~~ 

2. I haven't scoured the codebase itself and actually implement the usage of the Layer enums in the rest of the project as I didn't want to create a nightmare merge. I'd be more than happy to do that work should this PR be accepted.

**The Code**

- **Update**: The Layers.cs file is now automatically re-created when layers are updated, but you can force a regeneration with the menu item. ~~Adding a menu item that automatically generates a Layer and LayerMasks enums so we can use their defined values for layer comparison in the scripts.~~

- The generator is pre-configured to generate a file called "Layers.cs". It contains a pair of enums; Layer and LayerMasks. These reside in the "UOP1" namespace. These enums contains the layer ID's in the project and their respective layer masks. The LayerMasks enum has the Flags attribute applied to allow you to correctly use these LayerMasks to check against multiple layers at once, such as in Physics.Raycast.

- **Update:** I removed the FileExists check given the file is generated automatically now. ~~There are protections in the generator to stop accidental overwriting of already existing files in the project should the configuration values change.~~

- The generated Layers.cs file should be committed to the repository (and is included in this commit) to save others from regenerating it. The generated file also contains instructions on how to use it ~~and how to regenerate the file should a new layer be added to the project~~.

- Finally, I included a handful of extension methods for UnityEngine.LayerMask to make it easy to check if a layer is contained within a UnityEngine.LayerMask.


schodemeiss:
After some deliberation, I decided to have a go at writing this using SettingsProviders and some other improvements I thought I could make along the way. With that, I created a new PR #402 which is essentially this PR, but taken to the next level.

PR #406, opened 2021-04-09
Add beer keg model, material, and prefab

redcurry:
[Forum entry](https://forum.unity.com/threads/3d-art-assets.980892/page-5#post-7005836)

Added the beer keg model with prefab variant.


redcurry:
Wrong author e-mail. Closing and trying again.

PR #407, opened 2021-04-09
Add beer keg and cutting board models

redcurry:
Beer keg: [Forum thread entry](https://forum.unity.com/threads/3d-art-assets.980892/page-5#post-7005836)
Cutting board: [Forum thread entry](https://forum.unity.com/threads/3d-art-assets.980892/page-5#post-7005965)

Added beer keg and cutting board models and corresponding prefab variants.


amel-unity:
Hi @redcurry , thanks a lot for opening the PR, the models look great! I just updated the materials to Unlit URP and assigned the textures you made to the Base Map so we can see the models textured as soon as we drag the prefabs you made. I will merge the PR now :)

PR #409, opened 2021-04-20
Buttress root tree - clean version

Smurjo:
Extra big tree that
- allows the camera to move around underneath
- makes the player feel small
- provides big roots to contribute to a maze
- creates the typical forest atmosphere with light and dark patches

Triangles:
Tree crown 554
Tree trunk 576
Tree trunk collider mesh 193

Now only containing the files related to the buttress root tree.


ciro-unity:
Hey @Smurjo, are you still refining the tree? As of now we can't take it in as it doesn't match the rest of the assets in many stylistic aspects, including the texture (it's not very defined, image 1, plus the trunk is very dark), the topology (it's very angular, and I think some normals are broken, cfr image 2).

1:
image

2:
image

Smurjo:
I guess the shader still isn't working properly on my computer- I remember making the texture darker several times... 
I will fix this place that you showed on the top. It is hardly ever visible. 

Else I might remind that the geometry would look much less angular if had as many polygons as the "big tree" from the example assets. I will gladly add polygons if you say so...

ciro-unity:
Sorry @Smurjo, the tree is still not on the level we're hoping for :/
It has improved somehow, but the same issues remain.

Also another ask if you end up working on it more: can you make trunk and canopy into one mesh, and have the green vertex channel colour to modulate the wind? This way we save on one draw call (per tree).
Thanks!

PR #414, opened 2021-04-29
I'm curious why the AO handle result is not used in this context?

rocktavious:
Question:

I recently learned about open projects from dev blog 4 because of the addressables system (as i'm looking to implement a similar loading strategy) and i'm digging through the Initialization code and i'm wondering why the Async Operation Handle Result is not used?  Here is a PR that switches to using the Async Operation Handle Result - is this the more correct paradigm?


ciro-unity:
Hmm, no, not really. That is kind of a leftover from some refactoring. I don't think there's any big difference in either, if anything the present code declares a new `LoadEventChannelSO` just for the purpose of casting, so your patch is good. I'll approve.

👍🏻

PR #417, opened 2021-05-02
Fixed a simple typo in LoadingScreenManager

zcavaness:
In LoadingScreenManager.cs we subscribed to a listener (+=) in OnDisable instead of unsubscribing (-=).

Issue N/A
Card N/A


ciro-unity:
Thanks, good eye!

Issue #418, opened 2021-05-02
Enemy attacks protagonist during a dialogue


soloschenko-grigoriy:
**Short description**
When protagonists engaged in the conversation it gets attacked by a nearby enemy. This behavior distracts from conversation and quest, the player is completely defenseless during dialogue

**Expected behaviour**
An enemy should not be able attack the player while a conversation is on

**To Reproduce**
1. Launch `TestingGround` scene
2. There is a group of 2 townsfolks and Bard Here by the rock, get close to them.
3. Wait a few seconds and then start a conversation. Alternatively, you can start a conversation right away and wait...
4. ...until Slime Critter approaches and starts attacking you
5. Notice that you are being attacked in the middle of the conversation

![chop-chop-attack-during-dialogue](https://user-images.githubusercontent.com/5962670/116825169-0c802e00-ab5c-11eb-943d-0b42c1480c3a.gif)

PR #420, opened 2021-05-02
[Bugfix] New condition prevents Critter from attacking when dialogue on

soloschenko-grigoriy:
Issue:https://github.com/UnityTechnologies/open-project-1/issues/418
Forum: https://forum.unity.com/threads/bug-enemy-attacks-protagonist-during-a-dialogue.1103305/

This PR has two main changes:
1. Introduce `IsDialogueActive` Condition
2. Updates `Slime Critter` state machine to include aforementioned Condition in transition from `Idle` or `Roaming` state to `Alert` state

`IsDialogueActive`Condtion simply listens to Start and End dialog events emitted by Dialogue Manager. It preserves internal boolean value that becomes truthful when the new dialogue starts, falsy when dialogue ends. Assuming we may have only one active dialogue at a time, Condition should correctly indicate if there is any dialogue active currently. Finally,` Condition` uses that internal value as a result of `Statement`

Notice that Critter now keeps roaming without disturbing Hamlet who's busy in the convo with townsfolk:

chop-chop-critter-keeps-roaming

### DIFF
--- a/UOP1_Project/Assets/Scenes/WIP/TestingGround.unity
+++ b/UOP1_Project/Assets/Scenes/WIP/TestingGround.unity
@@ -1460,7 +1460,7 @@ PrefabInstance:
     - target: {fileID: 8341946629798114147, guid: 0b4e286cfc4c6e943841c36cc5529b22,
         type: 3}
       propertyPath: m_RootOrder
-      value: 38
+      value: 37
       objectReference: {fileID: 0}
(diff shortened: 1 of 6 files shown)


amel-unity:
Hello @soloschenko-grigoriy, thanks a lot for opening this PR! It solves the issue to make sure the player cannot be attacked while engaged in a dialogue. I will merge the PR now :) This should solve the linked issues. 
I will post on the forum about the next step to make sure the player cannot start a dialogue when engaged in combat. Thanks!

PR #424, opened 2021-05-05
Event Aggregator proposal

zaradai:
A limited change to demonstrate events published from Input.

An EventAggregator was created and a SO to hold the instance so that it can be shared within the project.

The publishers were located in Inputs and for limited changes only those events that were applicable to Protagonist was applied.

Also there is a UnityEventBus that was created to avoid reflection and link to UnityActions that could be exposed to the inspector or an existing function can be used.  The later was demonstrated in the camera manager.

### DIFF
--- a/UOP1_Project/Assets/Scripts/Characters/Protagonist.cs
+++ b/UOP1_Project/Assets/Scripts/Characters/Protagonist.cs
@@ -1,12 +1,21 @@
 ﻿using System;
 using UnityEngine;
+using static EventAggregator;
 
 /// <summary>
 /// <para>This component consumes input on the InputReader and stores its values. The input is then read, and manipulated, by the StateMachines's Actions.</para>
 /// </summary>
-public class Protagonist : MonoBehaviour
+public class Protagonist : MonoBehaviour,
+	IHandle<AttackEvent>,
+	IHandle<JumpEvent>,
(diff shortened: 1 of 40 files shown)


UOP1_Project/Assets/Scripts/Characters/Protagonist.cs
```diff
@@ -1,12 +1,21 @@
...
 using UnityEngine;
+using static EventAggregator;
 
 /// <summary>
 /// <para>This component consumes input on the InputReader and stores its values. The input is then read, and manipulated, by the StateMachines's Actions.</para>
 /// </summary>
-public class Protagonist : MonoBehaviour
+public class Protagonist : MonoBehaviour,
+	IHandle<AttackEvent>,
+	IHandle<JumpEvent>,
```
soloschenko-grigoriy, line 10: The naming here is gorgeous!

TH3UNKN0WN-1337:
Assets\ScriptableObjects\Input\InputReader.asset is missing the assignment of EventAgregator and EventBus. 
I like the approach you take here. We use something similar, inspired by Caliburn.Micro as well. Good job! :)

ciro-unity:
Cool ideas. I've written some feedback on the forums!
https://forum.unity.com/threads/event-system-refactoring.1113253/#post-7162234

PR #425, opened 2021-05-07
Fix fade-in volume. When fade in, volume is restored to 1 not  the vo…

LHSG:
…lume of the setting.

Set the end value of DOFade to the value of the setting

**[for Generic PRs]**  
How did you implement them?  

**[for Bugfix PRs]**  


ciro-unity:
Thanks, nice catch!

PR #433, opened 2021-05-19
[Feature] Block protagonist from starting dialogue if not in Gameplay state

soloschenko-grigoriy:
Forum link: https://forum.unity.com/threads/enemy-attacks-protagonist-during-a-dialogue.1103305/

This PR blocks protagonist from starting dialogue if game is currently in any state but Gameplay. 

1. Block UI Interaction Panel from popping up if Game State is not Gameplay
2. Similarly, preventing dialogue from starting if Game State is not Gameplay
3. Prevent Game from updating its state if new one is the same as current. This is helpful since otherwise previous state becomes the same as current and `ResetToPreviousGameState` looses its purpose 

Note: I do apologies for whitespace noise... My editor is kinda picky...


uChema:
Hey ! Thanks for the PR! 
I'm looking into it (at the same time as the other PR you opened #434 ). 
I've added a control on the interaction button (in the InputReader.cs). Since the UI has nothing to do with the interaction in itself, even it the interaction panel is not shown on screen, you can still interact with the character/pot/item.

uChema:
Hey @soloschenko-grigoriy , 
Thanks for your contribution ! We merged the #434 PR you opened later on. 
We're now using the StateMachine system to control the GameState and we don't have a problem with starting a dialogue if not in Gameplay State.

PR #434, opened 2021-05-19
[Feature] Set Game to Combat when Critter detects player

soloschenko-grigoriy:
Forum link: https://forum.unity.com/threads/enemy-attacks-protagonist-during-a-dialogue.1103305/

This PR sets Game State to combat when SlimeCritter detects protagonist and reverts to prev state when it loses the protagonist


uChema:
Hey ! 
I'm looking into this PR and the other one you opened. I checked and apparently the function that reverts the Game state to the previous one is called when the critter is in Fighting state

`public override void OnStateExit()	{		_gameState.ResetToPreviousGameState();	}`

Also when there are two enemies attacking, and one of them changes state, it will reset the GameState for everyone to the previous, even though one of them is still attacking. 

We also need to check when the scene changes, when the enemy dies or when the character dies.

PR #435, opened 2021-05-19
ForestMaze proposal

itsLevi0sa:
Scenes>Whiteboxing>Forest>Proposal_3 contains the following sub-folders:
-Materials
-MazeTiles
-Scene_FiniteForest 
-Scene_InfiniteForest 
-Scene_TransitionArea 
-Scripts

Each Scene has been pushed as a separate commit to not exceed the 100mb threshold and perhaps to make it easier for you to choose what to merge or not. 
*SPECIAL NOTE*
While pushing, I was prompted to fetch an automatic(?) update named: [Bot] Automated dotnet-format update, not sure what that is so perhaps you might want to skip that.

**General notes**
Each scene has the location exit-green box in it, but is not linked to the proper exit as I thought this should probably be done by you. The transition of one scene to another in this proposal would be: 
Town -> Exit to forest TransitionArea scene -> Exit to FiniteForest scene 
and from there exit briefly to the InfiniteForest scene and back to FiniteForest once the entrance to the KeyCharacter is found.
No scene is marked as Addressable, I included no LocationSOs.

**Scene notes**
For the TransitionArea scene:
Scene note: There is a mesh that is really big and stretched, I named it StitchMesh and it's used to make the border between the 3rd world and the skybox indistinguishable. When in play mode, the scale of it is not really (that) noticeable. 
Skybox note: The skybox used is with a scene not currently existing in the PR, it shows the forest with the previous trees size which was smaller but seems nicer to me in the overview.

For the InfiniteForest scene:
Camera note: The camera is meant to be top-down. If it starts changing perspective to 3rd person view, the player might be able to see the tiles rearranging. 
Tiles note: The tiles have a scale of 2.5 which I decided was a good size after some trial and error for the camera to not capture the rearrangement of the tiles while in top-down view. 
Code: Not much of a programmer here, but as long as it works...Currently the tiles are rearranging forever in any direction, so I guess everything might need to reset around 0,0,0 after some point.

For the FiniteForest scene:
It has some really low settings for the lightmap in order to not exceed the 100mb threshold.


ciro-unity:
Thanks, merging!

PR #437, opened 2021-05-21
Added a faint animation for the rock critter

mitchtheriault:

**[for Generic PRs]**  
Hi there! In the Animations card I saw that you need a "faint" animation for the rock critter, so I went ahead and made one.

**[for Bugfix PRs]**  


amel-unity:
Hi @mitchtheriault, thank you for opening the PR for the faint animation of the rock critter. 
Which model did you use to animate? After testing the PR locally, I noticed that the animation was not played from the Animator Controller and after checking in Blender you can notice that the rig is not the same as the fbxs we already have in the project (on the left the Idle fbx from the project, on the right the faint fbx from the PR). 

image (4)

You need to take one of the .fbxs already in the project, so the rig matches with the one we already have for the different animations.

amel-unity:
@mitchtheriault we have some new suggestions on [the rock critter thread](https://forum.unity.com/threads/rock-critter.1109786/) which [I just replied to](https://forum.unity.com/threads/rock-critter.1109786/#post-7199614) including ideas related to the faint animation and I wanted to know what are your thoughts on it :) Thanks!

Issue #442, opened 2021-06-04
PlantCritter y axis rotation (attack direction) resets during attack animation.


hinesca:
![2021-06-04 11_45_13-UOP1_Project - TestingGround - PC, Mac   Linux Standalone - Unity 2020 3 7f1 Per](https://user-images.githubusercontent.com/38190540/120849408-ecf96e00-c52a-11eb-811c-1203fbeb58ea.png)
![2021-06-04 11_46_19-UOP1_Project - TestingGround - PC, Mac   Linux Standalone - Unity 2020 3 7f1 Per](https://user-images.githubusercontent.com/38190540/120849429-f387e580-c52a-11eb-8c40-d484fe97c838.png)

**Sort description**

When the player approaches a PlantCritter, it rotates to face the character during the alert animation. When the character moves within attack range, the direction that the plant critter faces resets to the instantiated position during the attack animation.

**Expected behavior**

The PlantCritter should maintain attack direction during the attack animation.

**To Reproduce**

1. Open the game play scene and press play.
2. Move the character towards a PlantCritter and allow it to attack.
3. Move the character to the opposite side of the PlantCritter and allow it to attack.
4. Notice how the direction the PlantCritter resets during the attack animation.

**Notes**
- See uploaded files

hinesca:
This bug can be assigned to me. It seems that this will be an easy issue to fix. I don't quite have the experience with Unity and this project yet to know exactly where the issue lives within the project, but I intent on learning.

shuttle127:
Bugs don't get "assigned" per se, they're just noted and people work on them. If this is something you'd like to figure out, fork the main repo, code up the fix, and make a PR. Hopefully it's a simple enough fix, good luck!

amel-unity:
Hello!
Thank you @hinesca  for noticing the issue. This is happening because the Face Protagonist action called on the Alert, Fighting and Reloading states on the state machine of the plant critter is the one meant to be used for the NPC and not the Critter.

When I started working on the state machines of the NPCs, I used the FaceProtagonist action we had in the project, which was used for the critters. However when the protagonist starts talking to a townsfolk, we want them to reset to their initial rotation when they finish the conversation. This means if we have 2 townsfolks talking to each other, and the protagonist starts talking to one of them, the townsfolk will face the protagonist and when they finish the conversation, the townsfolk will reset to his initial position so he looks to the other townsfolk he resumes talking to.

[...]

PR #443, opened 2021-06-07
Patch for issue #442 (Plant Critter Attack Direction bug)

hinesca:
Fix for issue [#442](https://github.com/UnityTechnologies/open-project-1/issues/442)

This was fixed by adding the existing NPCFaceProtagonist action to the Attack State in the Plant Critter State Machine.

This can be verified resolved by: 
1. Open the game play scene and press play.
2. Move the character towards a PlantCritter and allow it to attack.
3. Move the character to the opposite side of the PlantCritter and allow it to attack.
4. Notice how the direction the PlantCritter now maintains the attack animation.


amel-unity:
Thanks again for opening the PR, I will close it and close the issue since it is already solved as discussed [on the Issue. ](https://github.com/UnityTechnologies/open-project-1/issues/442)

PR #444, opened 2021-06-07
Add multi_compile keywords to CustomLighting.hlsl, remove ShaderGraph keywords

hybridherbst:
This PR simplifies the ShaderGraph shaders: the multi_compile pragma can be in an include file, the keywords don't need to be declared separately on each Shader Graph. This is error prone, since someone could accidentally remove a keyword or use it in the graph incorrectly. Also, changing keywords (e.g. finding another one needs to be added to support lighting mode XY) is way harder if you have to hunt the keywords down in 10+ shaders.

The list of multi_compiles comes from URP's Lit.shader and contains a couple more items than were in ShaderGraph previously. I'm not sure if all lighting modes are used in Chop Chop but I think a more "complete" toon lighting can't hurt to have.

image


ciro-unity:
It's a good idea, I'll merge it as soon as possible (heavy work going on in the shaders on my side, so I don't want to mix up changes). Although I pulled your changes locally and it seems like it's merging fine.

Little preview :)
image

PR #445, opened 2021-06-07
Add additional light shadow support to Toon lighting model

hybridherbst:
The current custom lighting implementation didn't support additional light shadows. They are turned off on the UniversalRenderPipelineAsset, but since the existing ShaderGraph implementation did contain the relevant keywords I assume they are supposed to work.

This PR adds support for additional light shadows by calling the right GetAdditionalLight overload that takes shadowCoords. There's _probably_ an edge case missing for lightmapped objects that are supposed to also receive mixed lighting with shadows from additional lights, but I think for now it's nice to have shadows when wanted :)

Before / After:
20210608-014912_Unity
20210608-014803_Unity

### DIFF
--- a/UOP1_Project/Assets/Shaders/CustomHLSL/CustomLighting.hlsl
+++ b/UOP1_Project/Assets/Shaders/CustomHLSL/CustomLighting.hlsl
@@ -41,6 +41,21 @@ void DirectSpecular_float(float Smoothness, float3 Direction, float3 WorldNormal
 #endif
 }
 
+half4 GetShadowMask()
+{
+    // TODO: not sure if/how we can access inputData.shadowMask here.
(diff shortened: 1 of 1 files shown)


ciro-unity:
Not sure if needed. We generally don't have spotlights, I don't see an use for them in this game, and the point lights (torches, the phoenix chick, etc.) don't cast a shadow. As you know, point light shadows only came in more recent URP versions.

hybridherbst:
I saw that the ShaderGraph shaders defined the ADDITIONAL_LIGHT_SHADOWS keyword so it seems it was used at some point.

Not sure either if it's needed here, but it doesn't hurt either – people are gonna look at Open Project as a reference (many people wait for a proper toon shader for URP) so I think a bit more complete feature support is nice to have.

Issue #426, opened 2021-05-09
Player automatically hops on Slime Critter when walking into it


asb9599:
**Short description**
Very minor issue, but if the player is walking into the Slime Critter (which can happen if you are attacking while moving) or any other small objects with similar colliders (i.e. bushes), they will automatically hop up. This can make fighting the Slime Critters pretty awkward if you're facing them head on.

**Expected behaviour**
Player will stay on the ground as it walks around the Slime, just like how they walk around the Plant Critter.

**To Reproduce**
0. Set player health to an absurdly high number to avoid death while testing (optional) 
1. Open TestingGround or any gameplay scene.
2. Approach the Slime Critter (or one of the bushes) and walk/run into it
3. Player will sometimes forcibly hop as if trying to get over it

**Notes**
- The following video shows the typical occurrence of the bug as you try to walk into a Slime Critter:

https://user-images.githubusercontent.com/32971129/117580974-5e7b0380-b0c8-11eb-9e58-32811ce50b8f.mp4

- The bug is easier to reproduce if you disable the AI of the Slime Critter and just keep walking into it, like this:

https://user-images.githubusercontent.com/32971129/117581035-bb76b980-b0c8-11eb-8a05-48b865ab4d7d.mp4

- As mentioned before, this isn't just limited to the Slime Critter. You can reproduce this bug with any small physical object as you can see from the bushes here:

https://user-images.githubusercontent.com/32971129/117581184-99ca0200-b0c9-11eb-9bf1-2bf1c7551dd0.mp4

- I did notice that this does not happen when walking into the Plant Critters. The player just walks around them as shown here:

https://user-images.githubusercontent.com/32971129/117581300-1c52c180-b0ca-11eb-9695-024a48958e38.mp4

amel-unity:
Hi @asb9599 , thank you for opening the issue. There is a PR open that might fix it but I just wanted to mention it here also that we are working on fixing the character controller first since we have some unpleasant situations like what you mentioned on the issue or when the character gets stuck so we will try to fix that first and then we can work on the minor issues with the colliders on the objects. 
Thanks!

PR #447, opened 2021-06-17
Bugfix slime hop

pointfivetee:
Issue: https://github.com/UnityTechnologies/open-project-1/issues/426
Forum thread: https://forum.unity.com/threads/player-automatically-hops-on-slime-critter-when-walking-into-it.1107092/

Improved collision detection for a handful of game objects.

- Raised the slime critter's collider slightly, which eliminates the glitchy hopping behavior when the chef tries to walk near one.
- Decreased the height of the plant critters capsule collider, allowing the chef to hop over it without crashing into an invisible barrier.
- Gave the bush a tall capsule collider that "turns off" (becomes a trigger) when the player enters a trigger zone above the bush. This eliminates the hopping problem while still allowing the chef to jump over/onto the bush normally.

To test, enter a scene and try navigating near the objects listed above. When walking into them, the chef shouldn't hop or slide around. The chef should also be able to jump over/onto them without hitting an invisible barrier or hovering in midair. (You may want to disable the critters' state machines first so they don't keep trying to bite you.)

### DIFF
--- a/UOP1_Project/Assets/Prefabs/Characters/PlantCritter.prefab
+++ b/UOP1_Project/Assets/Prefabs/Characters/PlantCritter.prefab
@@ -69,9 +69,9 @@ CapsuleCollider:
   m_IsTrigger: 0
   m_Enabled: 1
   m_Radius: 0.5180242
-  m_Height: 5
+  m_Height: 2
   m_Direction: 1
-  m_Center: {x: -0.01654327, y: 2, z: 0.03308654}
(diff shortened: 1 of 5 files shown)


amel-unity:
Hi @pointfivetee , thank you for opening the PR to fix the issue :) I just wanted to mention that we are working on fixing the character controller first since we have some unpleasant situations like mentioned on the issue or when the character gets stuck so we will try to fix that first and then we can work on the minor issues with the colliders on the objects. 
Thanks!

Issue #450, opened 2021-06-25
Deleting transition from table causes errors


pointfivetee:
In the transition table editor, removing the last transition in a list causes a bunch of NullReference exceptions while the table inspector disappears out of focus.

**Expected behaviour**
The transition should simply disappear from the list without error messages or changing the focus of the table inspector.

**To Reproduce**
1. Open the transition table editor.
2. Pick a "from" state in the table and click on the "-" button near the last transition in the sub-list.
3. A bunch of error messages appear in the log and the transition table becomes unselected in the left-hand side of the editor popup.

**Notes**
Forum thread: https://forum.unity.com/threads/how-to-delete-a-transition-from-the-transition-table.1104163/

amel-unity:
Thanks a lot @pointfivetee for opening the issue and proposing a solution! The PR fixes it, I will merge it now :)

PR #451, opened 2021-06-25
Bugfix delete transition

pointfivetee:
Issue: https://github.com/UnityTechnologies/open-project-1/issues/450
Forum thread: https://forum.unity.com/threads/how-to-delete-a-transition-from-the-transition-table.1104163/

Resolved the issue by looking up an object's name *before* the object gets deleted.

To verify, open the transition table editor popup and delete the last transition in the list for a given "from" state.


ciro-unity:
Thanks, will check and merge it next week when @amel-unity is back at work.

amel-unity:
As mentioned on the issue, the PR fixes the issue. I will merge it now, thanks @pointfivetee  :)

PR #452, opened 2021-06-25
Add lightmap scale setter tool

pointfivetee:
Forum thread: https://forum.unity.com/threads/suggestions-on-where-to-start.1127687/#post-7270441
Codecks card: https://open.codecks.io/unity-open-project-1/decks/15-code/card/1gi-lightmap-scale-setter-tool

This editor tool makes it easy to change the lightmap scale for all components within a prefab chunk of terrain.

For best performance, the lightmap scale value should be smaller for components that are further away from the playable area of a scene.

Added a custom component with a field that triggers a mass update when its value is changed.


ciro-unity:
Thanks, I'm merging it. I just removed the cap at 1, I already have a spot in the game where I'll be using 1.3 or something, so it doesn't make sense to cap - it would disrupt work. Capping at zero does, however.

PR #454, opened 2021-06-27
Add camera shake on protagonist hit.

luvjungle:
Feature described in this card:
https://open.codecks.io/unity-open-project-1/decks/15-code/card/1gk-camera-shake-on-hit

Added Action to GettingHit state, created Void Event Channel and raising it in OnStateEnter() of Action.
Created GO under CameraSystem with Impulse Source. Added reference to Impulse Source and Event SO to Camera Manager script and subscribed.

### DIFF
--- a/UOP1_Project/Assets/Prefabs/Gameplay/CameraSystem.prefab
+++ b/UOP1_Project/Assets/Prefabs/Gameplay/CameraSystem.prefab
@@ -144,6 +144,79 @@ MonoBehaviour:
   m_AmplitudeGain: 0.3
   m_FrequencyGain: 0.2
   mNoiseOffsets: {x: 0, y: 0, z: 0}
+--- !u!1 &953227511577075091
+GameObject:
+  m_ObjectHideFlags: 0
+  m_CorrespondingSourceObject: {fileID: 0}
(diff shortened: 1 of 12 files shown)


amel-unity:
Thank you @luvjungle for adding the camera shake on hit. The PR works very well, I will merge it now :)

ciro-unity:
Just FYI, when you add a listener in OnEnable it's always good practice to remove it in OnDisable. As it is now, the camera yields an error if you change scene and then fight, since the old camera is still hooked to the event channel, but the ImpulseSource component that should respond has been destroyed.

I'll fix it on `main`, don't worry.

ciro-unity:
Had to remove the lambda expression to make it work.
The code is now:
`if (_camShakeEvent != null)
			_camShakeEvent.OnEventRaised += impulseSource.GenerateImpulse;`

and not
`if (_camShakeEvent != null)
			_camShakeEvent.OnEventRaised += () => impulseSource.GenerateImpulse();`

Then the listener is correctly removed in `OnDisable`
Was there a reason to use it?

luvjungle:
Sorry, my bad. I forgot about unsubscribing and used lambda because looked at couple of lines above, at another subscription. No real need for lambda here.

Issue #453, opened 2021-06-25
Looping SFX are still playing when changing scene


ciro-unity:
**Short description**
Looping SFX like the fire of the cooking pot are not stopped when moving between scenes.

**Expected behaviour**
All gameplay SFX should stop when a Location scene gets unloaded.

**To Reproduce**
1. Open the Beach scene.
2. Move to the Field_Hill scene.
3. Observe when the scene is loading, for an instant, you can hear the fire SFX very loud.
4. Once in Field_Hill, expand the list of sound emitters under the AudioManager in the scene PersistentManagers.
5. Observe that one of them is still playing the fire SFX from the previous scene.

PR #455, opened 2021-06-28
Stop audio cue when associated Component is destroyed

pointfivetee:
Issue: https://github.com/UnityTechnologies/open-project-1/issues/453
Codecks: https://open.codecks.io/unity-open-project-1/decks/13-bugs/card/1go-looping-sfx-are-still-playing-when-changing-scene
Forum: https://forum.unity.com/threads/suggestions-on-where-to-start.1127687/#post-7279576

Added a call to StopAudioCue() to the AudioCue script's OnDisable method.

To verify:
1. Start the Beach scene.
2. In the hierarchy explorer, go to PersistentManagers > AudioManager > SoundEmitterPool and note that one of its children is continuously playing the Env_Campfire audio clip.
3. Exit to Field_Hill. The campfire is no longer audible. The associated SoundEmitter may still be in the pool but is now disabled.
4. Return to the Beach. There is once again a SoundEmitter looping the campfire audio. (Possibly a different emitter in the pool than before.)

### DIFF
--- a/UOP1_Project/Assets/Scripts/Audio/AudioCue.cs
+++ b/UOP1_Project/Assets/Scripts/Audio/AudioCue.cs
@@ -27,6 +27,7 @@ public class AudioCue : MonoBehaviour
 	private void OnDisable()
 	{
 		_playOnStart = false;
+		StopAudioCue();
 	}
 
 	private IEnumerator PlayDelayed()


amel-unity:
Thanks @pointfivetee for this quick fix! I will merge it now :)

PR #457, opened 2021-06-29
Add script to orient audio listener with the camera

pointfivetee:
Forum: https://forum.unity.com/threads/audio-listener.1049138/#post-7282822
Codecks: https://open.codecks.io/unity-open-project-1/decks/15-code/card/1gw-orient-audio-listener-to-camera

Sound effects now go to stereo channels based on the camera's position, not the pig chef's. The result is less jarring to the player when the pig changes direction but not the camera (or vice versa).

Added a script to the listener that checks the main camera's `transform.forward` and applies it to its own transform.

To test:
1. Go to the Beach scene and walk to the campfire, so that the pig is facing forward and the fire is by his right ear. The crackling fire sound comes through the right speaker.
2. Walk around in circles without changing the camera angle. The fire sound remains in the right speaker.
3. Rotate the camera without turning the pig chef. The fire sound pans left and right based on the camera's direction.


ciro-unity:
Works perfectly, thanks! Merged!

PR #461, opened 2021-07-01
Water Mill 3d model

RTBRuhan:
Water Mill 3d model


ciro-unity:
Is the PR incorporating all the latest texture changes discussed on the forums?

RTBRuhan:
> Is the PR incorporating all the latest texture changes discussed on the forums?

Yes

ciro-unity:
I was asking mostly about the aligned UVs we discussed on the forums, [the user MortyJHyn made a good suggestion](https://forum.unity.com/threads/3d-art-assets.980892/page-7#post-7281259).

But even if you don't want to fix the wheel, I would still like if the poles could be fixed. It's very obvious that the UVs are taking some random wood grooves in random directions:
image

The other important thing is that the two supports should be separate objects (and in fact, you can recycle the UVs or even provide just one and we duplicate it in Unity). This would allow us to 1) reposition the wheel precisely and 2) make it spin!

As a third thing, but this depends on your 3D skills, it would be super nice if the wheel (only the wheel, not the supports) could have some baked AO. Only AO! Not directional lighting, because it will be rotated. This is just an extra plus to make it integrate better in the environment.

That's how it looks now:
image

RTBRuhan:
Oh no I am not sure why the old model is here 🤦‍♂️🤦‍♂️🤦‍♂️
Can you try using this 👇 model.... I need to learn how to use Github properly till then
https://drive.google.com/drive/folders/1kNQ2AX805D3hZpP7hDVm4uVqe1ypLw1_?usp=sharing

ciro-unity:
> Can you try using this 👇 model.... I need to learn how to use Github properly till then
Thanks!

Still a few issues with the texturing though :/

image

Again, mostly direction and sizing in the UVs, there's a bunch of random disconnected lines, and the supports have very thin veins compared to the rest of the object.

Another thing is that the wheel has a pivot on the base so it can't be rotated, but that can be fixed by reparenting the object to another GameObject to "fake" a different pivot.

But maybe at this point I could try to fix the UVs myself?

RTBRuhan:
https://drive.google.com/file/d/1Lnir3Vz88GvJ2mn-WQehtHqRmbW37nkx/view?usp=sharing

👆 Here is the update . I tried fixed the texturing issues and changed the wheel pivot in middle. And not sure where can I work on Uv so I have also included the project file so that you can modify it as you want. 😅😅😅

ciro-unity:
Thanks @RTBRuhan. I have taken the model from inside the Google Drive, so I'll close this PR. (it doesn't matter too much for you since you already figure as a contributor on the repo)

You'll see the watermill in the scenes next time I push to `main`.

PR #463, opened 2021-07-02
Implement pause menu music

luvjungle:
Feature described in this card:
https://open.codecks.io/unity-open-project-1/decks/15-code/card/1gx-implement-pause-menu-music

Implemented using events, added required SOs to Music Player prefab.
To test simply open and close pause menu (ESC key)


ciro-unity:
Looks good, nice clean implementation!

ciro-unity:
Lol, already found a small bug: the event gets fired not only when closing the menu, but also when moving to settings or even to the confirmation to go back to the main menu. But don't worry, it's not your fault, it's something we have to fix in the UI on when this event is fired.

Issue #459, opened 2021-06-30
Landing particles and SFX play when attacking in the air


ciro-unity:
**Short description**
When the attack button is pressed while the character is in the air, the landing particles and SFX play.

**Expected behaviour**
These effects should play only when landing.

**To Reproduce**
1. Open any scene. Press Play.
2. Jump, and press attack in the air.
3. Notice that the puff of dust particles are emitted, and the landing SFX plays.

**Notes**
- See video (contains regular jump + jump and attack)

https://user-images.githubusercontent.com/20049224/123949207-986ad680-d9a2-11eb-9063-79e93bf743f8.mp4

shuttle127:
So I saw this and thought, "hey, it's gotta be a simple state machine/transition issue," and I *think* I was right. Unfortunately, not quite sure how to solve it, so here's what I've got so far (apologies if someone else has already looked into it, I didn't see a thread on the forums yet).

==============================================

When you do a normal jump without attacking, the states are simply JumpAscending -> JumpDescending, and I tracked down a bit of the stack trace as follows:

1. PlayLiftoffJumpParticlesActionSO.OnStateEnter
2. PlayerEffectController.PlayJumpParticles (because I changed the call in step 1 to playjump instead of playland because it didn't make sense as playland when player hadn't even started jumping yet)
3. PlayJumpParticlesActionSO.OnStateEnter
4. PlayerEffectController.PlayJumpParticles
5. PlayLandParticlesActionSO.OnStateExit
6. PlayerEffectController.PlayLandParticles(intensity)

JumpAscending happens before step 1, and JumpDescending happens before step 5.

==============================================

When you do a jump followed by a mid-air attack, the states go JumpAscending -> JumpDescending -> JumpDescendingAttacking, and the stack trace I got for that is as follows:

1. PlayLiftoffJumpParticlesActionSO.OnStateEnter
2. PlayerEffectController.PlayJumpParticles
3. PlayJumpParticlesActionSO.OnStateEnter
4. PlayerEffectController.PlayJumpParticles
5. PlayLandParticlesActionSO.OnStateExit
6. PlayerEffectController.PlayLandParticles(intensity)
7. PlayLandParticlesActionSO.OnStateExit
8. PlayerEffectController.PlayLandParticles(intensity)

JumpAscending happens before step 1, JumpDescending happens before step 5, and JumpDescendingAttacking happens before step 7. The puff of dust particles and sound during the attack happen because of steps 5&6 I believe.

[...]

PR #464, opened 2021-07-02
Fixed landing particles playing in air when attacking bug

arkii1:
Issue:
https://open.codecks.io/unity-open-project-1/decks/13-bugs/card/1h2-landing-particles-and-sfx-play-when-attacking-in-the-air
Thread:
https://github.com/UnityTechnologies/open-project-1/issues/459

Like mentioned in the thread, the bug is caused because the state machine goes Ascending->Descending->DescendingAttacking where Descending calls the particles when exiting. I fixed this simply by adding a ground check on the PlayLandParticles action, so it cannot play in the air. The issue can be verified by playing the game, jumping and then attacking.


ciro-unity:
Thanks, seems like a sensible fix!

ciro-unity:
Congrats on your first contribution 🚀

PR #466, opened 2021-07-05
update: remove extra color param documentation for FadeIn

shukerullah:


ciro-unity:
Hey @shukerullah, thanks for the contributions but please, don't do these very very tiny fixes. They add work for us as we have to evaluate the PRs one by one. Thanks

PR #468, opened 2021-07-05
Rock Critter Attack SFX

LiamPollard:
(My first time using Git, I am unsure/think I have made an error in the merging of the main branches and then my attempt at correcting it just made it more messy) hopefully you can just accept the SFX commit and ignore the others but I don't know if thats possible from your end, I shall post in the forum for further guidance.

as referenced in the forum (Calstock8) https://forum.unity.com/threads/sound-design-sfx.1051298/#post-7276432
An Attack SFX for the Rock Critter.
Hopefully attached to the animation via Amel's script.


ciro-unity:
https://forum.unity.com/threads/sound-design-sfx.1051298/#post-7301347

PR #469, opened 2021-07-05
update: Refactor SpinningUI Script

shukerullah:


shukerullah:
Script name should be UISpinner and it should be inside UI folder instead SceneManagement.

ciro-unity:
Same as for the other mini-PRs. I did implement the changes you suggested though, including the relocation of the file (which is not in the PR).

PR #470, opened 2021-07-06
Fade manager

shukerullah:
Remove unused namespaces and rename color param to match its documentation


ciro-unity:
Same comment as for the other mini-PRs.

PR #473, opened 2021-07-08
Golden rock critter egg

Smurjo:
Golden_and_normal_rock_critter_egg
Rock critter egg prefab with new material with glowing golden veins


ciro-unity:
Thanks!
A few things that need to be tweaked before accepting the PR:
- The veins look stretched in some parts and generally very noisy. Personally I would paint them by hand rather than use a filter (?)
image
- The object is really small! I'd scale it 3-4 times, to make it visible. Current at 1,1,1: 
image
(this is actually true also for the regular egg)
- The emission map is very dark where it should glow, meaning that to get some decent emission going we need to crank up the emission value to crazy numbers (i.e. > 10) which also "burns" some areas. At the same time, it's not black where it shouldn't glow, meaning that if we crank up the value it will glow everywhere.
- Rotation is a random one, needs to be 8 degrees on the X to make the egg stand up (?)

Would be nice if you could fix these things, thank you!

Smurjo:
I don't like it to be that extremly emissive as you showed Ciro - remember normally eggs don't glow at all. This is what I did:
golden_Rock_critter_Egg_animation_007

ciro-unity:
Thanks. Ok, we won't make the egg so glowy. It just has a slight gleam on the edges, as if gold that is reflecting light.

I still found the texture very stretched and maybe too low-res, so I took the occasion to redo the UVs, and repainted the base rock critter egg texture too to make it more stony. I used Blender's texture paint mode, so I could get seamless veins.

This is how they look in-game now:

image

So I will close the PR now since I've reorganised materials and renamed, but I'll note the contribution down anyway, of course!

Smurjo:
"repainted the base rock critter egg texture too to make it more stony" - Uhm, I fear when I first saw it I didn't think of stone - but something more smelly. Maybe you get some second opinion on that texture. With veins it is ok.

PR #474, opened 2021-07-09
Modifying Railed Bridge Meshes(Changing Mesh and Colliders) 

MortyJHin:
This PR was based on a discussion on the Forums, where I modified the bridges for specific reasons, such as having support beams to reach the ground or no support beams to make the bridge look like it's floating. The mesh, collider, and some of the textures have been replaced and put in specific folders to stay organized. 

I haven't updated the scenes with the new long beamed bridges, but since I replaced the old beamed bridge, the one that leads to the kind chef's home has been replaced. 

Longer Beams: https://forum.unity.com/threads/3d-art-assets.980892/page-7

No Support Beams: https://forum.unity.com/threads/3d-art-assets.980892/page-8


ciro-unity:
Thanks, great work! Also for providing the simplified collision meshes! 👍🏻 

Only one small suggestion: you can recycle textures and materials with elements like this. In fact, I noticed the Material for the long bridge with beams work for the long bridge without. I'll do the cleaning.

Thanks, you'll see this merged soon when I merge my local changes with `main`, later this week. 🙏🏻

MortyJHin:
Oh cool! I made a new texture because the longer beams looked weird for some reason. That's awesome that we can use recycle the texture. Thank you!

PR #475, opened 2021-07-09
Cave whiteboxing

itsLevi0sa:
Rough prototype (still WIP?). Everything is in the Mountain_Cave folder on the Whiteboxing>Community folder. Not very experienced with NavMesh, you might find irrelevant stuff in there :) I also included the PressurePlate scripts made from @pointfivetee who posted his work on the Pressure Plates Thread on the forum, since I still use them in the scene. They are located in a sub-folder called PressurePlate, just in case! Cheers.


ciro-unity:
I really like your whiteboxings! :D I can see what you're doing with the level design, clever stuff.

However, due to a bunch of things I can't use the cave as it is, I will need to remake it (but maybe not from scratch).
It goes on top of the Path, and the entrance doesn't line up (it's higher). Plus you used a lot of rocks that are the prototype prefab, with materials as override (that can be changed with the Prefab tool). Finally, the rocks need to be broken into chunks, so we can use some as the "outer shell" and hide the ones inside the cave when you're not in the cave. Finally as I said, we don't have the stalagmites so unless somebody makes them, they will have to go too.

So I'll take some time to cleanup the scene before you see it in game, but for now, good work!

itsLevi0sa:
Hello again! Thanks a lot for your comments :) I am back, and I was currently working on that too, just wanted to share what I got back on Friday since I know we are on *fire*. I am about to send another commit actually with those "outer shell" chunks included and also the Path (that is based on your path! But with some minor tweaks). I know my cave is higher/bigger (*sigh*) than yours, I only decided to share my proposal as it was cause it would take more time to re-scale everything based on your scene (I just wanted you to have the general feel of it before moving on to the actual clean-up).  
Soo, this is what I was planning to commit now: 
2 scenes, the one I've sent above but improved (Scene A), with the path included and the outter shell organized etc. and another one _reduced (Scene B), that it fits the boundaries you've set (which means that you land straight into the main cave, I've discarded the point A and B on that one). 
For the Scene A I was thinking of grouping stuff together as shown below:
- For entering the Path you would use the main town square and only a first row of rocks to show a fake outline of the mountain you are about to enter:
Untitled-1
- For the Path itself, I would put that in a separate scene cause I think we could skip several steps and begin our escalation somewhere in the middle so it won't take much time to actually climb all the way up.
Untitled-2
- The main cave remains as is without all the outter stuff. For the stalagmites, sure, go ahead and hide them/delete them, same for the Pressure Plate, I thought you should be the one to decide :)
Untitled-3

However I do realize this Scene A goes waay beyond the prefab walls you planned. Even though it does make a biit more sense to me if the mountain does form some sort of tip (after all what is the point of climbing if we don't reach a peak..hmm :P) : 
Untitled-4
 
That's why I am also working on this Scene B now too, to fit your scene base (no photos yet). 
What do you think should I go ahead and commit those changes? You can expect to have these by tomorrow (hopefully).

ciro-unity:
Sounds good, show me your masterplan and then I will adapt it to what we have.

> * we could skip several steps and begin our escalation somewhere in the middle so it won't take much time to actually climb all the way up.

That is fine, the only thing is that I believe it would be good if you turn around and you see the exit and you recognise it takes you back to town - even if it's in the distance. This way you are not lost. Otherwise, you get dropped in the middle of a lot of rocks and it's like... "How am I here now?!"

And yes, anytime you can send them. I'll be waiting then :)
One thing: be careful, you have a lot of unrelated files in your commit, how comes? That is creating conflicts.

itsLevi0sa:
Hello, I've just commited the two scenes (with no Lightmaps yet, sorry). I did the changes mentioned above: hid the stalactites, replaced the "Rock_Proto" assets with the BigRocks2 prefabs with the Replace tool (fancy tool!), tried to create some better(?) rock chunks and tried to align the cave with the path both in my initial proposal (SceneA: PathAndCave scene) and your base scene (SceneB: PathAndCave_reduced scene). My playthrough of SceneA lasted 2.5 minutes. I agree with your note on entering the path and seeing the exit, we could keep a little bit of both?
Needless to say, keep whatever suits your needs, combine/keep/discard whatever chunk you might like or not like, the reason I am commiting two scenes is because we can see two approaches and keep the best of both/come up with something new :) 
Below you can see the chunks I created on SceneA:
Untitled-8
Untitled-5
Untitled-6
Untitled-7

As for the conflicting files, they were 100% not there earlier yesterday. I noticed they appeared after a new push in main from Chema that I haven't yet merged in my branch so that's why these conflicts emerged? Still, not sure. Should I try to merge main back in my branch and see if they insist? (I am still cautious and not very confident with GitHub, but I def haven't made nor included any changes on these unrelated files worry not).

itsLevi0sa:
Merged main into my branch again and made sure I have kept the modified files from main seemed to have solved the problem (I think). It's weird though, since I didn't touch those files. I guess this has something to do with scene dependencies?

ciro-unity:
Git can be hard to deal with sometimes, and we're working recklessly :D (since the communication is deferred and delayed).

> I guess this has something to do with scene dependencies?

No, that's not possible. The commit shows changed lines in scripts, etc. It's the result of some wild merge.

> seemed to have solved the problem (I think).

Unfortunately no. If you check the changed files at the top, you'll see there's 86 files and among them many scripts.
Maybe just send what you got, and I'll do some patching.

itsLevi0sa:
Yes I am done, it's the commit above where I created the PathAndCave and PathAndCave_reduced scene. Would it be better if I just make a new branch, up to date with main and do a new PR with those 2 scenes to avoid any confusion?

ciro-unity:
> Yes I am done, it's the commit above where I created the PathAndCave and PathAndCave_reduced scene. Would it be better if I just make a new branch, up to date with main and do a new PR with those 2 scenes to avoid any confusion?

No, it's ok, I can take care of it. 👍🏻

Issue #478, opened 2021-07-16
Bug:: UIMenuManager:: Throws NullReferenceException on game exit


paciFIST-Studios:
**Short description**
When exiting play mode via the in-game menu, UIMenuManager throws a NullReferenceException

**Expected behaviour**
No exceptions are thrown as the game shuts down

**To Reproduce**

1. Load project and enter play mode
2. Open pause menu -> Exit to Main Menu -> confirm exit

3. From Main Menu:
4. Choose Exit -> confirm exit

Result
5. Unity throws NullReferenceException

![image](https://user-images.githubusercontent.com/27395700/126015903-1f3242a5-05f4-49ed-b1f4-2f233db805dd.png)

NOTES:

This happens with or without an attached visual studio debugger.  I reproduced the bug with the debugger on, leading to the following picture.  I swapped the ordering, and I found no difference, whether    "_onGameExitEvent.OnEventRaised"    was called first, or if    "Application.Quit"    was called first.

"_onGameExitEvent" is a VoidEventChannelSO object;

I looked into the scene with the object inspector, and found the UIMenuManger to have a link to a VoidEventChannelSO.  I plan to tinker with this some today, and see if I can find a fix.

![image](https://user-images.githubusercontent.com/27395700/126016444-7027967c-bcca-45f8-9542-82b46c70fe6a.png)

uChema:
Hey @paciFIST-Studios  ! 
Thanks for bringing this to our attention !

I took a look at this issue and found out two things : 
First the NullRef error is thrown because we're calling _onGameExitEvent.OnEventRaised() and not _onGameExitEvent.RaiseEvent() 
Second : I don't think we're using this event anywhere else in the project. So no one is subscribed to this channel nor listening to the event. Also it's called after the Application.quit() . 
I'll continue investigation that part of the code but my guess is : this line will most certainly be removed !

PR #480, opened 2021-07-17
Update UIMenuManager.cs

paciFIST-Studios:
Summary:
Found a small bug, which was causing a NullReferenceException.

Bug ticket:
https://github.com/UnityTechnologies/open-project-1/issues/478

Forum Thread:
https://forum.unity.com/threads/bug-uimenumanager-throws-nullreferenceexception-on-game-exit.1142410/#post-7337182

Resolution:
Simple code modification, so we call RaiseEvent(), rather than OnEventRaised()

Verification Steps:
1. Launch game using editor play mode.
2. Using controller or keyboard and mouse:
3. Navigate back to the main menu
4. Attempt to exit the game
5. Observe the debug log

Result:
6. There are no NullReferenceException notifications in the log.

### DIFF
--- a/UOP1_Project/Assets/Scripts/UI/UIMenuManager.cs
+++ b/UOP1_Project/Assets/Scripts/UI/UIMenuManager.cs
@@ -152,7 +152,7 @@ public class UIMenuManager : MonoBehaviour
 		if (quitConfirmed)
 		{
 			Application.Quit();
-			_onGameExitEvent.OnEventRaised();
+			_onGameExitEvent.RaiseEvent();
 		}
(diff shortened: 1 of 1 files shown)


uChema:
Hi @paciFIST-Studios ! 
You're right ! That's the way to do it (I answered the issue thread and didn't see this one at that time) 
Although, like I said there, I don't think we need this line anymore. 
I pushed a Hotfix for it : Removing the line, the reference in the script and even the event Asset. 
If the event is needed in the future, I feel it's best if it's created at that time.

PR #482, opened 2021-07-20
Waterfall sfx

rokaso:
Adding Watefall sfx with adjusted spacializing to feel more natural walking around it
  
No thread was created for this.
This will add sfx of a waterfall to it's foam prefab.
To be able to hear what is happening on screen, thus making it more alive.
Added the sound to sfx/environment and implementing it on the waterfall foam prefab as a audio listener,


ciro-unity:
Hey @rokaso, thanks for the SFX. Can I ask you for 2 things?
- First, it would be great if you just send the audio file, as right now the changes in the prefab and ProBuilder data are creating a conflict. If it's easier, you might close this PR and create a new one.
- Did you make the SFX? Can you provide the running stream part and the falling water part as separate? Since we have a scene where the character moves close to the water and multiple waterfalls, it would be nice to be able to separate the 2 SFX and position them farther away.

Thank you!

Issue #486, opened 2021-08-06
Loading the currently loaded scene gives error.


HarshNarayanJha:
Hey there. Whenever we try to load a scene, it gets loaded from the AssetBundle and the the scene is loaded. But if we try to load it again while it is loaded, Addressables gives error that the **Asset** is already loaded. And that is expected.

**Expected behaviour**
Expected behaviour is the scene assets should be properly unloaded, then again loaded. I tried this, but was unsuccessful. So what I just did if currently loaded is equal to the requested scene, just return.

**To Reproduce**
I exactly don't know how to reproduce it, but maybe user **stroibot** on forums may know this. He reported the issue to me.
But I tested it in the Debugging System I made, and that was reproduced.

**Notes**
Forum Thread: https://forum.unity.com/threads/reloading-scene-using-addressables.1149521/#post-7377545

Thanks.

ciro-unity:
Sorry but this is not really an issue - unless I misunderstood? I'll be closing Issue and PR.

PR #487, opened 2021-08-06
Fixed Scene Reloading issue in SceneLoader.cs

HarshNarayanJha:
Yes, it is: https://github.com/UnityTechnologies/open-project-1/issues/486

Yes, it is too: https://forum.unity.com/threads/reloading-scene-using-addressables.1149521/#post-7377545

Just by returning from the LoadLocation function if currentScene == LocationToLoad;

Maybe using this: https://forum.unity.com/threads/debug-system.1151783/
I also verified that this issue is reproduced and also that it is fixed using this.
Or the user [stroibot](https://forum.unity.com/members/stroibot.1297325/) on unity forums.
He made me aware of that bug.

Thanks.


ciro-unity:
I misunderstood the PR in the first place. This is trying to "fix" an issue we don't have, and doing so with a simple silent failure (I know it gives the Debug.Log, but still, useless in the final build). So I'll be closing it.

Issue #488, opened 2021-08-16
NullReferenceException when entering dark red block


normyp:
**Short description**
This sometimes occurs:
```
NullReferenceException: Object reference not set to an instance of an object 
LocationExit.OnTriggerEnter (UnityEngine.Collider other) (at Assets/Scripts/SceneManagement/LocationExit.cs:20)
```
**Expected behaviour**
Should not raise an exception.

**To Reproduce**
Walk into the dark red block near the starting point.

ciro-unity:
As mentioned in the PR #489, we generally don't want to prevent NullReference Exceptions. They are a great warning that something is missing, and "silencing" them might lead to invisible bugs in the build.

I'll be closing this soon, just leaving it a bit for yours and others' awareness. Thanks anyway!

thedeepakchaturvedi:
The NullReferenceException occurs because we haven't specified the Location To Path & Location Storage to any initial value in the inspector panel, although its value is default in the script. So it is not able to load any scene when we enter the trigger. Is this issue closed ?? Can I work on this issue to remove the error that is preventing scene switching.

ciro-unity:
As mentioned in PR #489, the PR and the Bug are looking at the wrong issue and the wrong way to fix it. The issue is in the scene (i.e. the referenced SOs), not in the script. In fact, it doesn't mention the scene where it happens.
So don't worry about fixing it, I'm working on all scenes to wrap up the project so that might cause a conflict. I'll review all entrances/exits.
Thanks though!

normyp:
Okay, thanks for your feedback. :-)

PR #489, opened 2021-08-16
Prevent NullReferenceException

normyp:
Fixes https://github.com/UnityTechnologies/open-project-1/issues/488


ciro-unity:
Thanks @normyp, but we don't want to check for null references like that. The correct way to fix a missing reference would be just to... fix it, by referencing the correct object.

Did you see the issue happen in a specific scene?

PR #490, opened 2021-08-20
Debug system

HarshNarayanJha:
Yes, sure: https://forum.unity.com/threads/debug-system.1151783/

A Debugging System which will certainly help everyone developing the project.

If we want to go to another location while in play mode, we have to either restart or walk all the way. With this Debug System, we can warp to any location quickly. More info in thread.

How did you implement them? 
See in thread for full implementation.

Also @ciro-unity, pull request can't be automatically merged, but don't worry, you can accept both changes when it conflicts. Only one line that I have deleted is this:
Screenshot_20210820-081059_Chrome

This is the localization entry I did, so accept my change in that place.
Hope this helps.

Thanks.


HarshNarayanJha:
Hey Ciro, no need to fix conflicts. I fixed all of them when I realized that I am able to do so here.

Issue #499, opened 2021-09-25
Horizontal movement is stuck on rock


kablouser:
**Short description**
Moving into the rock sometimes stops all horizontal movement. You must jump to move again.

**Expected behaviour**
Moving into the rock shouldn't stop horizontal movement. (But obviously stops you from moving into the rock)

**To Reproduce**
1. Open to Scenes/Locations/Beach
2. Go to the rock south (-z direction) to the Hare.
3. Walk into rock.
4. All horizontal movement stops.

**Notes**
This bug is related to the Sliding state.
In the following video, I show the bug. A blue ray is drawn to indicate the slide direction.
https://user-images.githubusercontent.com/45899999/134778523-e1ecda95-ca5c-4c59-9711-37d7c80173d7.mp4

PR #532, opened 2023-04-22
Fixed rock sliding bug

dkphill:
This PR addresses issue #499.

I noticed that, while the player is stuck on the rock, adjusting the minimum slope on the characterController to be around 90 allowed the player to move again. Therefore, I decided to implement a failsafe in Protagonist.cs that detects when the player is colliding with a rock, and if so, sets the minimum slope value to 90. When they leave the rock collider, the minimum slope is set back to its original value. Although this does not necessarily address the root of this bug, I found that it made for a good temporary fix.

After implementing the changes, this is how the character and rocks behaved on my laptop: https://imgur.com/a/l1vArnX


avdheshb6 (approved):

avdheshb6:
Complete the review section

Ress-sama (approved):

